PR #2, opened 2020-09-04
Set up workspaces and initial API setup

tpluscode:


.gitignore
```diff
@@ -21,3 +21,5 @@ pnpm-debug.log*
 *.njsproj
 *.sln
 *.sw?
+
+.env
```
martinmaillard, line 25: I would move that with the other `.env` files. 

And I think `/dist` will still be nested in `ui`? Not really sure how it works with yarn workspaces

ui/package.json
```diff
@@ -32,13 +32,7 @@
     "@vue/eslint-config-standard": "^5.1.2",
     "@vue/eslint-config-typescript": "^5.0.2",
     "@vue/test-utils": "^1.0.3",
-    "eslint": "^6.7.2",
-    "eslint-plugin-import": "^2.20.2",
-    "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-promise": "^4.2.1",
-    "eslint-plugin-standard": "^4.0.0",
-    "eslint-plugin-vue": "^6.2.2",
-    "typescript": "~3.9.3",
```
martinmaillard, line 41: I'm guessing you removed these dependencies because linting will be done from the project root. I still have a eslint configuration in `ui/package.json`, should I remove it? Otherwise I'll extract it in `ui/.eslintrc.json` to stay consistent with the rest of the project.

.eslintrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "root": true,
+  "extends": [ "@tpluscode" ],
```
martinmaillard, line 3: What is `@tpluscode`? Is this a Zazuko-safe preset? :D

ui/package.json
tpluscode: Those moved to the core package.json to prevent conflicting versions (eslint fails to find them if the are in nested `node_modules`.

I thought to unify the lint configs but the one generated for UI is specific enough that I kept it largely intact for now.

martinmaillard: Ok perfect, I'll just extract it in a separate file then

.eslintrc.json
tpluscode: Pretty much [standard+typescript](https://github.com/tpluscode/eslint-config/blob/master/index.js). Used to be `zazuko/esling-config` but politics... ;)

PR #4, opened 2020-09-08
build: set up docker images for app and apis

tpluscode:
I created docker files for the API and UI.

Separate because I want to avoid the complication of having UI conditionally hosted inside express and build together. In this approach the UI is simply a static nginx container. The API is a node app like before but also made simpler

This time my proposal is to have them deployed like

| What | Where |
| -- | -- |
| API | https://creator.cube |
| UI | https://app.creator.cube |

### DIFF
--- a/api.Dockerfile
+++ b/api.Dockerfile
@@ -1,0 +1,42 @@
+# First step: build the assets
+FROM node:lts-alpine AS builder
+
+WORKDIR /
+ADD package.json yarn.lock ./
+ADD ./apis/core/package.json ./apis/core/
+ADD ./ui/package.json ./ui/
+
+# for every new package foo add
+#ADD ./packages/foo/package.json ./packages/foo/
+
+# install and build backend
+ENV NODE_ENV=
+RUN yarn install --ci
+
+COPY . .
+
+RUN yarn tsc --outDir dist
+
+FROM node:14-alpine
+
+WORKDIR /app
+
+ADD package.json yarn.lock ./
+ADD ./apis/core/package.json ./apis/core/
+ADD ./ui/package.json ./ui/
+
+# for every new package foo add
+#ADD ./packages/foo/package.json ./packages/foo/
+
+RUN yarn install --production
+COPY --from=builder dist/apis ./apis/
+#COPY --from=builder dist/packages/ ./packages/
+
+RUN apk add --no-cache tini
+ENTRYPOINT ["tini", "--", "node"]
+
+EXPOSE 45670
+USER nobody
+
+WORKDIR /app/apis
+CMD ["core"]
--- a/api.Dockerfile
+++ b/api.Dockerfile
@@ -10,7 +10,6 @@ ADD ./ui/package.json ./ui/
 #ADD ./packages/foo/package.json ./packages/foo/
 
 # install and build backend
-ENV NODE_ENV=
 RUN yarn install --ci
 
 COPY . .
--- a/apis/core/tsconfig.json
+++ b/apis/core/tsconfig.json
@@ -1,13 +1,7 @@
 {
+  "extends": "../../tsconfig.json",
   "compilerOptions": {
     "module": "CommonJS",
-    "target": "ESNext",
-    "esModuleInterop": true,
-    "strict": true,
-    "noImplicitAny": true,
-    "noImplicitThis": true,
-    "noImplicitReturns": true,
-    "strictNullChecks": true,
-    "allowSyntheticDefaultImports": true
+    "target": "ESNext"
   }
 }
--- a/app.Dockerfile
+++ b/app.Dockerfile
@@ -5,7 +5,6 @@ WORKDIR /
 ADD package.json yarn.lock ./
 ADD ./ui/package.json ./ui/
 # install and build backend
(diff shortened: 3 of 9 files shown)


apis/core/tsconfig.json
```diff
@@ -1,13 +1,7 @@
...
   "compilerOptions": {
     "module": "CommonJS",
-    "target": "ESNext",
-    "esModuleInterop": true,
-    "strict": true,
-    "noImplicitAny": true,
-    "noImplicitThis": true,
-    "noImplicitReturns": true,
-    "strictNullChecks": true,
-    "allowSyntheticDefaultImports": true
```
tpluscode, line 11: Moved up to have `tsc` happy when called from root

api.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+#ADD ./packages/foo/package.json ./packages/foo/
+
+RUN yarn install --production
+COPY --from=builder dist/apis ./apis/
+#COPY --from=builder dist/packages/ ./packages/
+
+RUN apk add --no-cache tini
+ENTRYPOINT ["tini", "--", "node"]
+
+EXPOSE 45670
```
tpluscode, line 38: @sandhose do we need a configurable PORT?

nginx/default.conf
```diff
@@ -0,0 +1,38 @@
+server {
```
tpluscode, line 1: Please be hard on this, as I reused what I found

Is this good? Is nginx fine?

api.Dockerfile
```diff
@@ -0,0 +1,42 @@
+# First step: build the assets
+FROM node:lts-alpine AS builder
+
+WORKDIR /
+ADD package.json yarn.lock ./
+ADD ./apis/core/package.json ./apis/core/
+ADD ./ui/package.json ./ui/
```
martinmaillard, line 7: Do we need the UI package.json here?

api.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# for every new package foo add
+#ADD ./packages/foo/package.json ./packages/foo/
+
+# install and build backend
+ENV NODE_ENV=
```
martinmaillard, line 13: Not sure what that means

api.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+RUN yarn tsc --outDir dist
+
+FROM node:14-alpine
+
+WORKDIR /app
+
+ADD package.json yarn.lock ./
+ADD ./apis/core/package.json ./apis/core/
+ADD ./ui/package.json ./ui/
```
martinmaillard, line 26: Same as above

nginx/default.conf
martinmaillard: I don't have enough expertise to review this part. But @sandhose or @vhf probably do.

api.Dockerfile
tpluscode: Strictly it would not be but that we cannot simply `COPY . .` and selectively choose the files needed for build. It does not hurt and the Dockerfiles is a little simpler

api.Dockerfile
tpluscode: I take this from the previous Dockerfile in curation. @sandhose is that needed?

vhf:
Nginx looks ok

api.Dockerfile
sandhose: This isn't set by the parent image, so no need to do that

api.Dockerfile
sandhose: Not needed, but in an ideal world I would not have to worry because every container that expose a HTTP service should listen on http-alt/8080 by default ; the port mapping should be done by the container runtime :)

Not a big deal though

nginx/default.conf
sandhose: there could be a way to make nginx work as non-root, but it's not a priority for now. Otherwise LGTM

app.Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+ENV NODE_ENV=production
+RUN yarn build
+
+FROM nginx:1.16.0-alpine
+HEALTHCHECK --timeout=1s --retries=99 \
+        CMD wget -q --spider http://127.0.0.1:80/ \
+         || exit 1
+
+RUN apk add --update --upgrade --no-cache wget
```
sandhose, line 23: There is a `wget`-lite in busybox (which is included in alpine). `wget -q --spider` seems to work with it, so probably no need to install the full `wget`

app.Dockerfile
```diff
@@ -0,0 +1,26 @@
+# First step: build the assets
+FROM node:lts-alpine AS builder
+
+WORKDIR /
+ADD package.json yarn.lock ./
+ADD ./ui/package.json ./ui/
+# install and build backend
+ENV NODE_ENV=
```
sandhose, line 8: Same as in API, probably not needed

sandhose (changes requested):
Left a few comments ; really good dockerfiles nevertheless 👍

lucafurrer (approved):

lucafurrer (approved):

PR #5, opened 2020-09-08
build: set up local OIDC endpoint with dex

tpluscode:
Adds dex OIDC server to the docker-compose (via lando) to have a real authentication flow locally

### DIFF
--- a/.local.env
+++ b/.local.env
@@ -1,9 +1,17 @@
 NODE_ENV=development
 
+# OIDC config
+AUTH_ISSUER=https://auth.cube-creator.lndo.site
+AUTH_AUDIENCE=cube-creator
+AUTH_JWKS_URI=http://dex:5556/keys
+
+# Core API
 API_CORE_BASE=https://cube-creator.lndo.site/
 
+# Vue UI app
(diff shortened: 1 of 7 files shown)


.local.env
```diff
@@ -1,9 +1,17 @@
 NODE_ENV=development
 
+# OIDC config
+AUTH_ISSUER=https://auth.cube-creator.lndo.site
+AUTH_AUDIENCE=cube-creator
+AUTH_JWKS_URI=http://dex:5556/keys
```
tpluscode, line 6: The `lndo.site` URLs cannot be used inside the containers right (which I think is a bug, reported as https://github.com/lando/lando/issues/2579)

Instead the service name has to be used to call dex from API to verify token

apis/core/index.ts
```diff
@@ -17,6 +17,10 @@ async function main() {
 
   app.use(authentication())
 
+  app.get('/authenticated', (req, res) => {
+    return res.status(200).end()
+  })
+
   app.get('/protected', guard().check('scope'), (req, res) => {
```
tpluscode, line 24: @sandhose how would we configure dex to have custom scopes?

PR #6, opened 2020-09-08
feat(ui): authentication using OIDC

martinmaillard:
We still have an issue causing the UI to continuously call the OIDC provider for a while for some reason. This issue is also present in the original data-cube-curation tool.

### DIFF
--- a/.local.env
+++ b/.local.env
@@ -1,9 +1,14 @@
 NODE_ENV=development
 
 API_CORE_BASE=https://cube-creator.lndo.site/
+VUE_APP_API_URL=https://cube-creator.lndo.site/
 
 UI_BASE=https://app.cube-creator.lndo.site/
 
 STORE_QUERY_ENDPOINT=http://store:3030/cube-creator/query
 STORE_UPDATE_ENDPOINT=http://store:3030/cube-creator/update
 STORE_GRAPH_ENDPOINT=http://store:3030/cube-creator/data
+
+AUTH_AUDIENCE=https://datacube.zazukoians.org/
+AUTH_ISSUER=https://keycloak.zazukoians.org/realms/zazuko-dev
+AUTH_CLIENT_ID=cube-creator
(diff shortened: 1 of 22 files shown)


.local.env
```diff
@@ -1,9 +1,14 @@
...
 
 UI_BASE=https://app.cube-creator.lndo.site/
 
 STORE_QUERY_ENDPOINT=http://store:3030/cube-creator/query
 STORE_UPDATE_ENDPOINT=http://store:3030/cube-creator/update
 STORE_GRAPH_ENDPOINT=http://store:3030/cube-creator/data
+
+AUTH_AUDIENCE=https://datacube.zazukoians.org/
+AUTH_ISSUER=https://keycloak.zazukoians.org/realms/zazuko-dev
+AUTH_CLIENT_ID=cube-creator
```
martinmaillard, line 14: Not sure we should have this in the repo...

martinmaillard: I just saw #5. We're just gonna have to merge some stuff ;)

tpluscode: Maybe you'd merge this into #5 and we handle OIDC in one PR?

.local.env
```diff
@@ -1,9 +1,14 @@
 NODE_ENV=development
 
 API_CORE_BASE=https://cube-creator.lndo.site/
+VUE_APP_API_URL=https://cube-creator.lndo.site/
```
tpluscode, line 4: Why add the same value as another variable?

martinmaillard: It needs to start with `VUE_APP_` to be visible in vue. Maybe we can hack something in the webpack config to avoid the duplicated variable. I'll check.

tpluscode: I tried some simple tricks but failed. 

Nothing super critical. We might never ever touch that again and take care if this ever bites us. I propose to keep it like that for now

martinmaillard: 👍

tpluscode (approved):

PR #7, opened 2020-09-10
chore: complete setting up auth middleware

tpluscode:


.local.env
```diff
@@ -1,14 +1,18 @@
...
 UI_BASE=https://app.cube-creator.lndo.site/
 
+# SPARQL endpoints
 STORE_QUERY_ENDPOINT=http://store:3030/cube-creator/query
 STORE_UPDATE_ENDPOINT=http://store:3030/cube-creator/update
 STORE_GRAPH_ENDPOINT=http://store:3030/cube-creator/data
 
-AUTH_AUDIENCE=https://datacube.zazukoians.org/
+# OIDC config
+AUTH_AUDIENCE=cube-creator
```
tpluscode, line 16: Do we need a new client in Keycloak to have a proper audience here? @sandhose

apis/core/lib/auth.ts
```diff
@@ -12,15 +14,15 @@ declare module 'express-serve-static-core' {
...
-const createJwtHandler = (credentialsRequired: boolean) => jwt({
+const createJwtHandler = (credentialsRequired: boolean, jwksUri: string) => jwt({
   // Dynamically provide a signing key
   // based on the kid in the header and
   // the signing keys provided by the JWKS endpoint.
   secret: jwksRsa.expressJwtSecret({
     cache: true,
     rateLimit: true,
     jwksRequestsPerMinute: 5,
-    jwksUri: `${env.AUTH_ISSUER}/.well-known/jwks.json`,
```
tpluscode, line 23: Reading this from `.well-known/openid-configuration` seeing how the JWKS endpoint has varying URL in oidc providers. So much or `.well-known` 🙄

martinmaillard:
I get a 502 on `/env-config.js` when testing this PR

.local.env
sandhose: Yup; I think the instructions are up to date here: https://hackmd.io/rW_4v3BsRQS04DKDLKYpDQ#Full-stack-app-check-list

tpluscode:
Can you try again @martinmaillard? 502 could mean that the API was not up yet (that is mostly coming from Lando's proxy)

martinmaillard:
It may have been caused by a failure during build. I re-installed and rebuilt everything and it seems to work fine.

PR #8, opened 2020-09-14
Build docker images with GitHub Actions

sandhose:
This is heavily inspired by the [example in the docker/build-push-action](https://github.com/docker/build-push-action#complete-workflow) repo.

It builds two images: `zazuko/cube-creator-{api,app}`. Tags as follow: 
 - `sha-{commit}` on every commit
 - `edge` for the latest commit on master
 - `[branch]` for other branches
 - on git tags like `vX.Y.Z`, tags `vX`, `vX.Y`, `vX.Y.Z` and latest
 - `pr-X` on pull requests (builds but does not push)

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,141 @@
+name: docker
+
+on:
+  push:
+    branches:
+      - "**"
+    tags:
+      - "v*.*.*"
+  pull_request:
+
+jobs:
+  build-api:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Prepare
+        id: prep
+        run: |
+          DOCKER_IMAGE=zazuko/cube-creator-api
+          VERSION=noop
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+          elif [[ $GITHUB_REF == refs/heads/* ]]; then
+            VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
+            if [ "${{ github.event.repository.default_branch }}" = "$VERSION" ]; then
+              VERSION=edge
+            fi
+          elif [[ $GITHUB_REF == refs/pull/* ]]; then
+            VERSION=pr-${{ github.event.number }}
+          fi
+          TAGS="${DOCKER_IMAGE}:${VERSION}"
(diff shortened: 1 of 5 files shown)


tpluscode:
Would be nice to also have a docker build on Pull Requests?

sandhose:
Its the case, just that it is not pushed. Secrets are not available when building PRs, so we can't push them to Docker Hub. Also, it runs and push for each branches anyways

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,147 @@
+name: docker
+
+on:
+  push:
+    branches:
+      - "**"
+    tags:
+      - "v*.*.*"
```
tpluscode, line 8: Aha, this makes me realise. My default choice would be to have separate version history for the API and UI. They would produce their own tags like `@cube-creator/core-api@v0.1.0`.

Unless you would like a single version for the entire repository @martinmaillard ?

martinmaillard: I would personally have opted for a single version for the entire repository, as it makes maintenance easier. Any argument for having separate versions?

tpluscode: Building and deploying separately. Many changes in the UI will be unrelated to the API - no version bump. And vice versa.

I'm not sure that a single version is going to make reduce maintenance

martinmaillard: Building and deploying will be automated so I don't see a problem deploying everything for every version, even if one of the parts didn't change.

vhf: @tpluscode ok for you if we go with a single version?

tpluscode: Yes, tentatively. I would like to see a proposal for managing this version. As far as tooling is concerned there is nothing that I know of which will help us.

To explain, we have a monorepo which is at least the UI and API which we can evolve somewhat independently by keeping their change history and versions. Tooling exists to manage this (`lerna version` or `atlassian changesets`)

The alternative is what? To ignore the `version` in packages and only keep the one in root `package.json`?

tpluscode (approved):

.github/workflows/docker.yaml
vhf: Unless we publish the API and the UI as packages I'm not sure it really matters. What's important at this point is having docker images to deploy

tpluscode: We package them as two docker images which is close enough, isn't it?

vhf: Right, I missed that part. Let's go with this and we'll revisit when we need to.

PR #9, opened 2020-09-15
feat: first hydra-box resources

tpluscode:
This is PR brings in a basic read-only capability via hydra-box (actually it's a wrapper on top of hydra-box I build at [hypermedia-app/labyrinth](https://github.com/hypermedia-app/labyrinth); more details in comments).

The core principles are simple: 

1. Every resource `Foo` has its entire contents inside `GRAPH <Foo>` in the triplestore.
1. "Static" or "virtual" resources such as collections or the API entrypoint are also added to the triple store
   * they are inserted to the store on app start
2. As much of API behavior as possible is will be controlled by metadata directly in the API Documentation

The API Documentation is right now deployed from filesystem from turtle files in `hydra/` directory

### DIFF
--- a/apis/core/bootstrap/cubes.ts
+++ b/apis/core/bootstrap/cubes.ts
@@ -1,0 +1,13 @@
+import { turtle } from '@tpluscode/rdf-string'
+import { hydra, rdf } from '@tpluscode/rdf-ns-builders'
+import { cube } from '@cube-creator/core/namespace'
+
+export const cubes = turtle`
+<cubes> {
+  <cubes> a ${hydra.Collection} ;
+    ${hydra.manages} [
+      ${hydra.property} ${rdf.type} ;
+      ${hydra.object} ${cube.Cube}
+    ] .
+}
+`
--- a/apis/core/bootstrap/entrypoint.ts
+++ b/apis/core/bootstrap/entrypoint.ts
@@ -1,0 +1,9 @@
+import { turtle } from '@tpluscode/rdf-string'
+import { hydra } from '@tpluscode/rdf-ns-builders'
+
+export const entrypoint = turtle`
+<> {
+  <> a ${hydra.Resource} ;
+    ${hydra.title} "Cube Creator"
+}
+`
--- a/apis/core/bootstrap/index.ts
+++ b/apis/core/bootstrap/index.ts
@@ -1,0 +1,25 @@
+import $rdf from 'rdf-ext'
+import Parser from '@rdfjs/parser-n3'
+import toStream from 'string-to-stream'
+import StreamClient from 'sparql-http-client/StreamClient'
+import { cubes } from './cubes'
+import { log } from '../lib/log'
+import { entrypoint } from './entrypoint'
+
+const parser = new Parser()
+const initialResources = [cubes, entrypoint]
+
+export async function bootstrap(storeUrl: string, base: string) {
+  const client = new StreamClient({
+    storeUrl,
+  })
+
+  const dataset = await initialResources.reduce((previous, turtle) => {
+    return previous.then(next => {
+      return next.import(parser.import(toStream(turtle.toString({ base }))))
+    })
+  }, Promise.resolve($rdf.dataset()))
+
+  log('Bootstrapping API resources')
+  await client.store.put(dataset.toStream())
+}
--- a/apis/core/hydra/index.ttl
+++ b/apis/core/hydra/index.ttl
@@ -1,15 +1,39 @@
 @prefix hydra: <http://www.w3.org/ns/hydra/core#> .
 @prefix code:  <https://code.described.at/> .
 
-<api> a                    hydra:ApiDocumentation ;
-      hydra:entrypoint     <> ;
-      hydra:supportedClass hydra:Resource,
-                           hydra:Collection .
+<api>
+  a
+    hydra:ApiDocumentation ;
+  hydra:entrypoint
+    <> ;
+  hydra:supportedClass
+    hydra:Resource,
+    hydra:Collection .
 
-hydra:Resource hydra:supportedOperation [ hydra:method       "GET" ;
-                                          code:implementedBy [ a         code:EcmaScript ;
-                                                               code:link <node:@hydrofoil/labyrinth/resource#get> ] ] .
+hydra:Resource
+  hydra:supportedOperation
+    [
+      hydra:method
+        "GET" ;
+      code:implementedBy
+        [
+          a
+            code:EcmaScript ;
+          code:link
+            <node:@hydrofoil/labyrinth/resource#get>
+        ]
+    ] .
 
-hydra:Collection hydra:supportedOperation [ hydra:method       "GET" ;
-                                          code:implementedBy [ a         code:EcmaScript ;
-                                                               code:link <node:@hydrofoil/labyrinth/collection#get> ] ] .
+hydra:Collection
+  hydra:supportedOperation
+    [
+      hydra:method
+        "GET" ;
+      code:implementedBy
+        [
+          a
+            code:EcmaScript ;
+          code:link
+            <node:@hydrofoil/labyrinth/collection#get>
+        ]
+    ] .
--- a/apis/core/hydra/index.ttl
+++ b/apis/core/hydra/index.ttl
@@ -1,0 +1,15 @@
+@prefix hydra: <http://www.w3.org/ns/hydra/core#> .
+@prefix code:  <https://code.described.at/> .
+
+<api> a                    hydra:ApiDocumentation ;
+      hydra:entrypoint     <> ;
+      hydra:supportedClass hydra:Resource,
+                           hydra:Collection .
+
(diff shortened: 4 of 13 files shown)


apis/core/bootstrap/cubes.ts
```diff
@@ -0,0 +1,13 @@
...
+import { hydra, rdf } from '@tpluscode/rdf-ns-builders'
+import { cube } from '@cube-creator/core/namespace'
+
+export const cubes = turtle`
+<cubes> {
+  <cubes> a ${hydra.Collection} ;
+    ${hydra.manages} [
+      ${hydra.property} ${rdf.type} ;
+      ${hydra.object} ${cube.Cube}
+    ] .
```
tpluscode, line 11: Hydra crash course 💥 

`hydra:manages` declares indirect statements about the collection contents. There are always two properties from `subject/predicate/object` describing a triple pattern.

The example above would translate to something like 

```SPARQL
SELECT ?member 
{ 
  ?member 
    rdf:type  # hydra:property
    cube:Cube # hydra:object
}
```

And accurate enough, `@hydrofoil/labyrinth` does something similar as a basis for a query to retrieve the `/cubes` resource

apis/core/bootstrap/entrypoint.ts
```diff
@@ -0,0 +1,9 @@
+import { turtle } from '@tpluscode/rdf-string'
+import { hydra } from '@tpluscode/rdf-ns-builders'
+
+export const entrypoint = turtle`
+<> {
+  <> a ${hydra.Resource} ;
```
tpluscode, line 6: Hydra crash course 💥

`hydra:Resource` is defined as 

> a class of dereferencable resources

Anything which has this class and exists in its own named graph will be immediately available as resource on the API surface (via `labyrinth`)

apis/core/hydra/index.ttl
```diff
@@ -0,0 +1,15 @@
...
+@prefix code:  <https://code.described.at/> .
+
+<api> a                    hydra:ApiDocumentation ;
+      hydra:entrypoint     <> ;
+      hydra:supportedClass hydra:Resource,
+                           hydra:Collection .
+
+hydra:Resource hydra:supportedOperation [ hydra:method       "GET" ;
+                                          code:implementedBy [ a         code:EcmaScript ;
+                                                               code:link <node:@hydrofoil/labyrinth/resource#get> ] ] .
```
tpluscode, line 11: Hydra crash course 💥

This is how `hydra-box` attaches actual `express` handlers to RDF classes, using the `code:implementedBy` property.

In the simple case reading resources, generic handlers are loaded from `labyrinth` to provide the `GET` "routing" for most resources. This should cater for 90% of the read requests.

apis/core/index.ts
```diff
@@ -13,21 +18,21 @@ async function main() {
...
+      endpointUrl: env.STORE_QUERY_ENDPOINT,
+      updateUrl: env.STORE_UPDATE_ENDPOINT,
+      storeUrl: env.STORE_GRAPH_ENDPOINT,
+    },
   })
 
-  app.get('/protected', guard().check('pipelines:read'), (req, res) => {
-    return res.status(200).end()
-  })
+  await bootstrap(env.STORE_GRAPH_ENDPOINT, baseUri)
```
tpluscode, line 35: This is where the "static" resources from `bootstrap` directory are inserted into the store on startup

apis/core/bootstrap/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const parser = new Parser()
+const initialResources = [cubes, entrypoint]
+
+export async function bootstrap(storeUrl: string, base: string) {
+  const client = new StreamClient({
+    storeUrl,
+  })
+
+  const dataset = await initialResources.reduce((previous, turtle) => {
```
martinmaillard, line 17: I don't understand this part. `cubes` and `entrypoint` are a dataset right? You are transforming the `dataset` into a `string`, then a `stream`, and then parsing it again into a `stream of quads`, and finally importing it into another `dataset`?

martinmaillard:
Great!

Two questions:
- What does `bootstrap` mean in `apis/core/bootstrap`? Is it temporary?
- How confident are you that we should use labyrinth? I'm open to try, but it looks like it adds one more piece that's hard to understand

vhf (approved):
Thanks for the helpful comments, keep them coming :)

tpluscode:
> What does bootstrap mean in apis/core/bootstrap? Is it temporary?

Temporary; yes and no. This is for resources which do not change, or change only when code changes. So it makes sense to update them when deploying/running the application. 

What I view as temporary is the way of simply replacing everything but that should be enough for a while.

> How confident are you that we should use labyrinth? I'm open to try, but it looks like it adds one more piece that's hard to understand

Quite confident. It is after all also code we control, so it can become anything we need. Right now it wraps hydra-box to reduce much repeatable patterns for creating resource handlers and comes with a bunch of conventions for resource authorizations/access control, collection filtering/ordering, error handling. Much that is missing from bare hydra-box. Mostly inspired by the experience from curation project.

I like to think that it is to hydra-box what rdf-ext is to the core RDF/JS stack. Features needed to be productive which might at some point become part of the core library

apis/core/bootstrap/index.ts
tpluscode: `cubes` and `entrypoint` are just strings. They are all reduced-parsed into the `dataset` and then pushed to the store

martinmaillard: Do I understand correctly: every time we boot up the app, it will actually insert data in the triplestore?

How does it handle upgrades? What if the data shape changes? Does it replace the data every time?

tpluscode: Yes, simply replace those resources every time.

martinmaillard: Ok thanks!

lucafurrer:
Thanks a lot for your comments, they are helpful. 
Is there any documentation for the turtle` ... ` notation?

lucafurrer (approved):

tpluscode:
@lucafurrer it is a EcmaScript [tagged template string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates). Basically a language feature where you can customise how string interpolation is done inside the `${}` parts
 
The `turtle` function obviously serialises RDF/JS objects into turtle strings and also generated the `@prefix` declarations for common vocabularies (via [@zazuko/rdf-vocabularies](https://npm.im/@zazuko/rdf-vocabularies)) so that you don't have to

Here are some details and example: https://t-code.pl/rdf-string/#/format/turtle

apis/core/hydra/index.ttl
```diff
@@ -0,0 +1,15 @@
...
+
+hydra:Collection hydra:supportedOperation [ hydra:method       "GET" ;
+                                          code:implementedBy [ a         code:EcmaScript ;
+                                                               code:link <node:@hydrofoil/labyrinth/collection#get> ] ] .
```
martinmaillard, line 15: I don't really like this turtle formatting because:
- It's really hard to do by hand
- Changing something in one line may require changing all the other lines of the resource (= terrible diffs)
- It doesn't help reading

Do we have linting/formatting rules for turtle?

tpluscode: What do you propose?

I get this formatting by a WebStorm plugin. It's not great but at least I get consistent structure. Problem is that not everyone uses webstorm.

I wish there was an eslint turtle parser+formatter. Otherwise we're back to manually formatting which can only be worse.

martinmaillard: > at least I get consistent structure

The line above is not even consistent, `code:implementedBy` is not aligned correctly ;)

My proposition would be something like:
```
hydra:Collection hydra:supportedOperation [ 
  hydra:method  "GET" ;
  code:implementedBy [ 
    a  code:EcmaScript ;
    code:link <node:@hydrofoil/labyrinth/collection#get> ;
  ] ;
] ;
.
```

tpluscode: Heh yea, not consistent because it does not format itself unless reformatted in the IDE 😬.

Thus my question, I'm fine with tweaking the desired formatting but how do we keep it that way? I don't want to spend time doing that manually all the time

martinmaillard: I don't have an automated solution, but I think we should at least chose a format that is doable by hand

apis/core/hydra/index.ttl
```diff
@@ -1,15 +1,39 @@
...
+      hydra:method
+        "GET" ;
+      code:implementedBy
+        [
+          a
+            code:EcmaScript ;
+          code:link
+            <node:@hydrofoil/labyrinth/collection#get>
+        ]
+    ] .
```
martinmaillard, line 39: What the hell, so many new lines 😆

tpluscode: 🤷

martinmaillard (approved):

PR #16, opened 2020-09-16
style(ui): enforce slightly stricter rules

vhf:
Not a big change but it should make upcoming code a bit more consistent


martinmaillard (approved):
👍

PR #18, opened 2020-09-17
UI->API authentication

### LINKED ISSUE #15: Setup authentication between UI and API
The authentication process is as follows:
1. The UI authenticates with the OIDC provider
2. The OIDC provider provides an authentication token to the UI
3. The UI sends this token with each request to the API
4. The API verifies the token with the OIDC provider

Steps 1, 2 and 4(?) are already working. This issue is about step 3.

We are going to use [Alcaeus](https://github.com/wikibus/Alcaeus) (Hydra client) as API client. An example of how to send the token with each API request is available on https://github.com/zazuko/data-cube-curation/blob/master/ui/src/api/index.ts#L50

martinmaillard:
Setup the API client (Alcaeus) with authentication token (if available).

The `Authenticated` route will automatically load the API entrypoint when entered. All the children routes will have access to it and be able to load other resources.

Will close #15


martinmaillard:
I just noticed a typescript error. Will try to find a solution.

ui/vue.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  transpileDependencies: ['alcaeus'],
```
tpluscode, line 2: Is this necessary? I tried and Alcaeus "just worked"

ui/src/api/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (token) {
+    headers.set('Authorization', `Bearer ${token}`)
+  }
+
+  return headers
+}
+
+export const api = {
+  async fetchResource (url: string): Promise<HydraResource> {
+    const response = await Hydra.loadResource(url)
```
tpluscode, line 29: Maybe not possible right here, I usually take advantage of destructuring like 

```
const { response, representation } = await Hydra.loadResource(url)
```

ui/src/api/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    if (response?.response?.xhr.status !== 200) {
```
tpluscode, line 31: `loadResource` always returns truthy

```suggestion
    if (response.response?.xhr.status !== 200) {
```

tpluscode (changes requested):

ui/vue.config.js
martinmaillard: I didn't try without it. Just ported that from the old project. Will try

tpluscode (approved):

PR #19, opened 2020-09-17
feat: rename cubes to projects and add to entrypoint

tpluscode:
This PR aligns the API root resource with current model as described by #10 

* `<cubes>` changes to `<cub-projects>`
* linked form the API entrypoint
* added setup for e2e api tests

Issue #12, opened 2020-09-15
List cube projects


martinmaillard:
Users need to be able to list existing cube projects

tpluscode:
Right now every user will see all projects?

martinmaillard:
At first, yes 👍

lucafurrer:
This is done, no?

vhf:
Kinda, it still needs quite some work. For instance @tpluscode mentioned we will probably only show your own projects (or the projects you have access to).

PR #20, opened 2020-09-17
Barebones projects list

martinmaillard:
Part of #12 

Depends on #19 

I figured it would make easier for other people to continue if I wrote an example of how to use the API.

This PR loads the projects collection and displays a list of projects. The main difference with the data-cube-curation project is that I tried to "embrace" Hydra/Alcaeus more. The API wrapper is way lighter and I do a bit more work in the Vuex store. We'll see how this works out.

I didn't setup any [Alcaeus mixin](https://alcaeus.hydra.how/latest/representations/mixins.html) yet, but that's probably going to make things easier in the next steps.

### DIFF
--- a/api.Dockerfile
+++ b/api.Dockerfile
@@ -6,8 +6,8 @@ ADD package.json yarn.lock ./
 ADD ./apis/core/package.json ./apis/core/
 ADD ./packages/core/package.json ./packages/core/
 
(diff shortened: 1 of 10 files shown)


vhf (approved):
👍

tpluscode (approved):

PR #23, opened 2020-09-18
add workflow to read-me

lucafurrer:
I added a section workflow into the read-me.
A question still around from my side was if there is a need for a standard naming conventions for branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,6 @@ To trust the CA, follow the steps on <https://docs.lando.dev/config/security.htm
 
 ## Workflow
 
-This repository is using the [GitHub-Flow](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/github-flow). Hence all changes should integrated using pull requests.
+This repository is using the [GitHub-Flow](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/github-flow). Hence all changes should be integrated using pull requests.
 
 To submit a bug or a feature request please create an issue in this repository.
--- a/README.md
+++ b/README.md
@@ -25,9 +25,15 @@ Docker containers will start and the services will be available under the these 
 
 | Service | URL                                 |
 | --      | --                                  |
-| API     | https://cube-creator.lndo.site/     |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,9 +25,15 @@ Docker containers will start and the services will be available under the these
...
+| UI      | <https://app.cube-creator.lndo.site/> |
+| Fuseki  | <https://db.cube-creator.lndo.site/>  |
 
 Lando uses its own Certificate Authority and it won't be trusted by your system.
-To trust the CA, follow the steps on https://docs.lando.dev/config/security.html#trusting-the-ca
+To trust the CA, follow the steps on <https://docs.lando.dev/config/security.html#trusting-the-ca>
+
+## Workflow
+
+This repository is using the [GitHub-Flow](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/github-flow). Hence all changes should integrated using pull requests.
```
vhf, line 37:
```suggestion
This repository is using the [GitHub-Flow](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/github-flow). Hence all changes should be integrated using pull requests.
```

vhf (approved):
👍 , could you please also add a reference to https://www.conventionalcommits.org/en/v1.0.0/#summary ?

tpluscode:
@vhf actually no decision regarding conventional commits has been made so far.

vhf:
IMO documenting what we have been doing so far would make sense

lucafurrer:
> @vhf actually no decision regarding conventional commits has been made so far.

@tpluscode @vhf I am open to include conventional commits. I can add it and we change it, if needed or we can have a discussion first. as you prefer.

vhf:
Please add it then, glancing at the commit log it's a convention we've been (more or less) applying so far, adding a reference to its source makes sense. No need to 100% commit to it so let's stay flexible and only have a note saying something along the lines of "Commit messages usually follow …". That way we won't have issues with the occasional automatically-generated commit messages that won't fit.

PR #26, opened 2020-09-22
feat: creating a cube project

tpluscode:
This is a first end-to-end workflow of a writing operation using Hydra+SHACL

There is quite a bit happening here. Please refer to comments


apis/core/bootstrap/shapes/cube-project.ts
```diff
@@ -0,0 +1,30 @@
...
+      ${sh.maxCount} 1 ;
+      ${sh.in} (
+        "CSV"
+        "Existing cube"
+      ) ;
+      ${sh.order} 1 ;
+    ] ;
+  .
+}
+`
```
tpluscode, line 30: `sh:targetClass` is required to get validation running. It should match the `hydra:expects` annotation on the operation

apis/core/hydra/projects.ttl
```diff
@@ -8,9 +12,19 @@ cc:ProjectsCollection
...
-        cc:CreateProjectAction ; ;
+        cc:CreateProjectAction ;
       hydra:title
         "New project" ;
       hydra:method
-        "POST"
+        "POST" ;
+      hydra:expects
+        cc:CubeProject,
+        <shape/cube-project/create> ;
```
tpluscode, line 22: Both `hydra:Class` and `sh:NodeShape` are listed here. This way even if the request does not explicitly assert a `<> rdf:type cc:CubeProject` triple, validation will occur

apis/core/lib/domain/indentifiers.ts
```diff
@@ -0,0 +1,15 @@
+import UrlSlugify from 'url-slugify'
+import $rdf from 'rdf-ext'
+import { NamedNode } from 'rdf-js'
+import { GraphPointer } from 'clownface'
+import env from '../env'
+
+const url = new UrlSlugify()
+
+export function cubeProject(label: string): NamedNode {
+  return $rdf.namedNode(`${env.API_CORE_BASE}cube-project/${url.slugify(label)}`)
```
tpluscode, line 10: `url-slugify` ensures unique and URL friendly idenifiers

apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -0,0 +1,33 @@
...
+    const csvMapping = clownface({ dataset: $rdf.dataset() })
+      .namedNode(id.csvMapping(project))
+      .addOut(rdf.type, [cc.CsvMapping, hydra.Resource])
+
+    project.addOut(cc.csvMapping, csvMapping)
+
+    await store.put(csvMapping)
+  }
+
+  await store.put(project)
```
tpluscode, line 31: I pondered a `store.put` method which would accept a dataset and split it into named graphs but that would require much care when handling graphs using clownface.

This way it is safer because any resource pointer is assumed to be a default graph with only its triples

apis/core/lib/handlers/cube-projects.ts
```diff
@@ -0,0 +1,15 @@
...
+import { createProject } from '../domain/cube-projects/create'
+
+export const post = protectedResource(shaclValidate, asyncMiddleware(async (req, res) => {
+  const project = await createProject({
+    resource: await req.resource(),
+    store: req.app.resources(),
+  })
+
+  res.status(201)
+  res.header('Location', project.value)
```
tpluscode, line 13: Location header is necessary when doing POST so that the client knows what resource got created

apis/core/lib/middleware/resource.ts
```diff
@@ -0,0 +1,35 @@
...
+  req.resource = async () => {
+    const dataset = await req.dataset()
+
+    const expectedTypes = req.hydra.operation
+      .out(hydra.expects)
+      .has(rdf.type, hydra.Class)
+
+    return clownface({ dataset })
+      .namedNode('')
+      .addOut(rdf.type, expectedTypes)
```
tpluscode, line 31: Here is where the hydra Classes are added to the request payload to ensure SHACL validation correctly finds focus nodes to process

apis/core/lib/middleware/shacl.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (resource.dataset.size === 0) {
+    return next(new error.BadRequest('Resource cannot be empty'))
+  }
+
+  const validationResult = new SHACLValidator(shapes).validate(resource.dataset)
+  if (validationResult.conforms) {
+    return next()
+  }
+
+  return next(new error.BadRequest())
```
tpluscode, line 38: In the future the Bad Request should be extended with actual SHACL validations

apis/core/lib/middleware/shacl.ts
```diff
@@ -0,0 +1,39 @@
...
+  const resources = req.app.resources()
+
+  const shapes = $rdf.dataset()
+  await Promise.all(req.hydra.operation.out(hydra.expects).map(async (expects) => {
+    if (expects.term.termType !== 'NamedNode') return
+
+    const pointer = await resources.get(expects.term, false)
+    if (pointer.has(rdf.type, [sh.NodeShape]).values.length) {
+      await shapes.addAll([...pointer.dataset])
+    }
```
tpluscode, line 17: SHACL Shapes are resources so they have to be retrieved from the triple store

babel.config.json
```diff
@@ -0,0 +1,13 @@
...
+      "@babel/preset-env",
+      {
+        "targets": {
+          "node": "current"
+        }
+      }
+    ],
+    "@babel/preset-typescript"
+  ]
+}
```
tpluscode, line 13: Used by mocha

package.json
```diff
@@ -24,5 +26,16 @@
     "standard": "^14.3.4",
     "ts-node": "^9.0.0",
     "typescript": "^4.0.2"
+  },
+  "resolutions": {
+    "@rdfjs/express-handler": "tpluscode/express-handler#dataset-multiple-times"
+  },
```
tpluscode, line 32: Temporary until an update to `@rdfjs/express-handler` is available

packages/core/namespace.ts
```diff
@@ -1,4 +1,18 @@
...
+  'ProjectsCollection' |
+  'CsvMapping' |
+  'EntryPoint'
+
+type CubeCreatorProperty =
+  'projectSourceKind' |
+  'projects' |
+  'csvMapping'
+
+type CubeCreatorNamespace = Record<CubeCreatorClass | CubeCreatorProperty, NamedNode>
```
tpluscode, line 15: By making the namespace builder a fixed interface, any usage of the `cc` namespace will require only the use of terms listed above to prevent accidental typos and having to look up the actual terms every time.

WiIl also make refactoring easier

e2e-tests/cube-projects/create.hydra
```diff
@@ -0,0 +1,34 @@
...
+
+                Expect Type <https://cube-creator.zazuko.com/vocab#CubeProject>
+                Expect Property rdfs:label "Test Project"
+                Expect Link cc:csvMapping {
+                    Expect Type <https://cube-creator.zazuko.com/vocab#CsvMapping>
+                }
+            }
+        }
+    }
+}
```
tpluscode, line 34: A not on the `.hydra` filed. This is a DSL for end-to-end testing of Hydra API. It's documentation is at https://testing.hypermedia.app/dsl/

It is similar to `supertest` used in mocha but obviously executes on an actual application and has built-in understanding of Hydra semantics of `Operation`, `Link` etc.

apis/core/lib/domain/indentifiers.ts
```diff
@@ -0,0 +1,15 @@
+import UrlSlugify from 'url-slugify'
```
vhf, line 1: `identifiers.ts`

apis/core/types/hydra-box/Api.d.ts
```diff
@@ -0,0 +1,32 @@
+declare module 'hydra-box/Api' {
```
vhf, line 1: `.d.ts` don't seem to be linted?

vhf (approved):
👍

apis/core/types/hydra-box/Api.d.ts
tpluscode: Probably not. I would also not care too much. Quite soon I would like these types to go to `DefinitelyTyped`

vhf: Works for me, I was simply curious about it being the only file with 4 spaces indentation 👍

tpluscode: Ah, fixed that actually 🎉

PR #27, opened 2020-09-22
Docker/API: use a numeric UID instead of a user name

sandhose:
The Kubernetes manifests have a `runAsNonRoot` policy. But because the user set in the Dockerfile is a username and not a user ID, Kubernetes can't check if that username is not actually `root` without running the image, so it refuses to run the image.

*/me goes fixing his "Docker best practices" doc*


vhf (approved):
Thanks!

PR #28, opened 2020-09-22
Docker: Serve the app under /app/

sandhose:
That was not really the case, now it is. Please check that the changes in `vue.config.js` do not break the dev environment, I haven't set it up locally.


vhf (approved):
Checked locally, nice 👍

PR #29, opened 2020-09-23
refactor(ui): do not show entire API URLs in browser

tpluscode:
The goal here is to shorten the browser addresses. Right now the full URL is repeated in the router path which causes ugly (url escaping) and long UI addresses.

The idea is to simply remove the base URL from resource identifiers which go into the routing and replace slashes with something else to avoid `%2F` escape sequence

```diff
-/cube-projects/https:%2F%2Fcube-creator.lndo.site%2Fcube-project%2Ftest-project-5vvgzxadd3x
+/cube-projects/cube-project!!test-project-5vvgzxadd3x
```

---

It's also an example of how one can extend the model of Hydra's resource objects.

By augmenting the `HydraResource` interface and registering a mixin function with Alcaeus it is possible to add properties and methods to API resource objects


ui/src/api/mixins/ApiResource.ts
```diff
@@ -0,0 +1,24 @@
...
+      }
+    }
+
+    return BaseClass
+  }
+
+  ApiResourceMixin.shouldApply = true
+
+  return ApiResourceMixin
+}
```
tpluscode, line 24: Hydra resource objects are extended using partial mixin classes which are added to the prototype chain.

The `shouldApply` can also be a function and the preferred way is to associate a mixin function with an RDF type like `ApiResourceMixin.appliesTo = _NamedNode_`

ui/src/api/mixins/ApiResource.ts
```diff
@@ -0,0 +1,24 @@
+import { Constructor } from '@tpluscode/rdfine'
+import { Mixin } from '@tpluscode/rdfine/lib/ResourceFactory'
+
+declare module 'alcaeus/Resources' {
+  export interface HydraResource {
+    clientPath: string
+  }
+}
```
tpluscode, line 8: This extends the base interface so that the new property is recognised on all instances of `HydraResource`

ui/src/api/index.ts
```diff
@@ -2,8 +2,10 @@ import { Hydra } from 'alcaeus/web'
 import { HydraResource } from 'alcaeus'
 import store from '@/store'
 import { APIError } from './errors'
+import { apiResourceMixin } from './mixins/ApiResource'
 
 const rootURL = process.env.VUE_APP_API_URL
+const segmentSeparator = '!!' // used to replace slash in URI to prevent escaping
```
tpluscode, line 8: I chose the double exclamation mark arbitrarily as it has a close to nil risk of colliding with actual API resource identifiers. And the stand out well enough in the address bar...

ui/src/api/index.ts
```diff
@@ -12,6 +14,8 @@ if (!rootURL) {
 // Tells Hydra to use the API root URI as base URI for relative URIs
 Hydra.baseUri = rootURL
 
+Hydra.resources.factory.addMixin(apiResourceMixin(rootURL, segmentSeparator))
```
tpluscode, line 17: Make Alcaeus aware of a mixin

vhf (approved):
LGTM

Interesting how it works, this kind of setup seems to require quite some knowledge of hydra and alcaeus but the end result looks clean. 👍

PR #30, opened 2020-09-24
refactor: improve the ResourceStore to simplify consumer code

tpluscode:
@vhf inspired by our call earlier today I improved the `ResourceStore` class so that it actually is used to manage the resource graph pointers over the course of a request

* [x] use it to `create` new in-memory resources
* [x] persist all resources at once to their correct named graphs using `save`
* [x] added unit tests


apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -13,21 +12,17 @@ interface CreateProjectCommand {
 export async function createProject({ resource, store }: CreateProjectCommand): Promise<GraphPointer> {
   const label = resource.out(rdfs.label).term!
 
-  const project = clownface({ dataset: $rdf.dataset() })
-    .namedNode(id.cubeProject(label.value))
+  const project = store.create(cubeProject(label.value))
```
tpluscode, line 15: Small change but I think this produces a much more readable effect

apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -13,21 +12,17 @@ interface CreateProjectCommand {
...
       .addOut(rdf.type, [cc.CsvMapping, hydra.Resource])
 
-    project.addOut(cc.csvMapping, csvMapping)
-
-    await store.put(csvMapping)
+    project.addOut(cc.csvMapping, mapping)
   }
 
-  await store.put(project)
+  await store.save()
```
tpluscode, line 26: And everything will be saved in one go, without doing multiple roundtrips to the SPARQL graph endpoint

vhf (approved):
Easier to understand indeed 👍

apis/core/lib/domain/cube-projects/create.ts
lucafurrer: Nice!

lucafurrer (approved):
I like the improved readability

PR #31, opened 2020-09-25
refactor(api): authorize explicitly

tpluscode:
Related to a discussion with @sandhose, I propose not to secure the entire API by default and instead apply scope-based restrictions on specific handlers/classes

### DIFF
--- a/apis/core/hydra/index.ttl
+++ b/apis/core/hydra/index.ttl
@@ -1,5 +1,7 @@
-@prefix hydra: <http://www.w3.org/ns/hydra/core#> .
-@prefix code:  <https://code.described.at/> .
+prefix hydra:  <http://www.w3.org/ns/hydra/core#>
+prefix code:   <https://code.described.at/>
+prefix auth:   <http://hypermedia.app/auth#>
+PREFIX scopes: <https://cube-creator.zazuko.com/scopes#>
 
 <api>
   a
@@ -8,9 +10,12 @@
     <> ;
   hydra:supportedClass
     hydra:Resource,
-    hydra:Collection .
+    hydra:Collection
+.
 
 hydra:Resource
+  auth:scopes
+    scopes:pipelines-read ;
   hydra:supportedOperation
     [
       hydra:method
@@ -25,6 +30,8 @@ hydra:Resource
     ] .
 
 hydra:Collection
+  auth:scopes
+    scopes:pipelines-read ;
   hydra:supportedOperation
     [
       hydra:method
--- a/apis/core/hydra/projects.ttl
+++ b/apis/core/hydra/projects.ttl
@@ -1,9 +1,13 @@
-BASE          <urn:hydra-box:api>
-PREFIX cc:    <https://cube-creator.zazuko.com/vocab#>
-prefix hydra: <http://www.w3.org/ns/hydra/core#>
-prefix code:  <https://code.described.at/>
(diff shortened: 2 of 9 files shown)


apis/core/hydra/index.ttl
```diff
@@ -25,6 +31,10 @@ hydra:Resource
     ] .
 
 hydra:Collection
+  auth:required
+    true ;
+  auth:scopes
+    scopes:pipelinesRead ;
```
tpluscode: Reading resources and collections would require the `pipelines:read` scope

apis/core/hydra/projects.ttl
```diff
@@ -27,4 +31,6 @@ cc:ProjectsCollection
           code:link
             <file:handlers/cube-projects#post> ;
         ] ;
+      auth:scopes
+        scopes:pipelinesWrite
```
tpluscode: While and request to write resources might require `pipelines:write`

apis/core/lib/auth.ts
```diff
@@ -40,13 +40,13 @@ function devAuthHandler(req: Request, res: Response, next: NextFunction) {
...
-    const permissionHeader = req.headers['x-permission']
-    const permissions = typeof permissionHeader === 'string' ? permissionHeader.split(',').map(s => s.trim()) : permissionHeader || []
+    const permissionHeader = req.headers['x-scope']
+    const scope = typeof permissionHeader === 'string' ? permissionHeader.split(',').map(s => s.trim()) : permissionHeader || []
 
     req.user = {
       id: $rdf.namedNode(sub),
       sub,
-      permissions,
+      scope,
```
tpluscode, line 49: `scope` not `permissions` claim used to authorise

apis/core/hydra/index.ttl
```diff
@@ -8,9 +10,12 @@
     <> ;
   hydra:supportedClass
     hydra:Resource,
-    hydra:Collection .
+    hydra:Collection
+.
 
 hydra:Resource
+  auth:scopes
+    scopes:pipelines-read ;
```
tpluscode, line 18: Adding this to a class restricts access to tokens with specific scope(s).

apis/core/hydra/projects.ttl
```diff
@@ -27,4 +31,6 @@ cc:ProjectsCollection
           code:link
             <file:handlers/cube-projects#post> ;
         ] ;
+      auth:scopes
+        scopes:pipelines-write
```
lucafurrer, line 35: Does the creation of a project require the ability to write to the pipelines or could this be seperated?

tpluscode: Certainly. The entire permission is up for discussion

lucafurrer: I was not sure, if you had already some discussions.

martinmaillard:
Does that mean anyone could query some endpoints (e.g. api docs), or do you still need a valid token?

Does it make anything more complex?

tpluscode:
Yes, this will open api docs.

Complex? Maybe. The question is how fine-grained we want our access control...

martinmaillard:
That's why I'm asking. We don't need any access control for the MVP. I'm fine with having it already if it doesn't make everything else more complex, as we're probably going to need it later in the project. But if it slows down development significantly, I would drop it for now.

tpluscode:
Fair enough. Closing for now

PR #32, opened 2020-09-25
APIs readme

tpluscode:
Some notes on the back end project structure etc

### DIFF
--- a/apis/readme.md
+++ b/apis/readme.md
@@ -1,0 +1,154 @@
+# Cube-creator Hydra APIs
+
+The projects inside the `apis/` directory are express applications which use [Hydra Core][hydra] vocabulary to provide the necessary features for the Cube Creator application(s).
+
+[hydra]: http://www.hydra-cg.com/spec/latest/core/
+
+This document presents some important design and implementation choices to observe when working with the API as well as during implementation.
+
+## Project structure
+
+The directory tree below presents a preferred structure of an API project.
+
+```
+.
+└── project/
+    ├── bootstrap/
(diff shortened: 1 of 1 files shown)


apis/readme.md
```diff
@@ -0,0 +1,154 @@
...
+    │   ├── domain/
+    │   └── handlers/
+    └── test/
+```
+
+### `bootstrap/`
+
+Contains static resources which are inserted into the triple store at application start to ensure it is functional. This is because every resource exposed by the API requires its representation stored in the database. This includes the API root (entrypoint) as well as "virtual" resources such as [collections][collection] and write-only resources.
+
+[collection]: http://www.hydra-cg.com/spec/latest/core/#collections
```
lucafurrer, line 31: Hi
A markdown linter hinted once to me, that bare urls are not that well supported in Markdown
https://github.com/updownpress/markdown-lint/blob/master/rules/034-no-bare-urls.md. 

```suggestion
[collection]: <http://www.hydra-cg.com/spec/latest/core/#collections>
```

vhf: It's not a bare url, it's a link reference, see it that way: `[title](something)`, something is the URL of the link, not a bare url in a text block. ;)

vhf (approved):

apis/readme.md
lucafurrer: Missed that, sorry

vhf (approved):

martinmaillard:
I like this PR :)

Issue #13, opened 2020-09-15
Mapper: upload CSV file


martinmaillard:
In the mapper, users need to be able to upload a CSV file.

- The file will be stored in a pre-defined S3 bucket
- It must handle large files
- A sample of the file must be extracted and stored somewhere

tpluscode:
I don't think we need to extract the sample. 

If we intend to have the files always accessible from S3 it would be simpler to simply open a stream directly from storage and read a portion. This would even allow running the preview on any arbitrary subset of rows from anywhere in the csv

lucafurrer:
One mapping can have more than one file, correct?

lucafurrer:
Open questions to be discussed/approved

**Workflow**: 
1. When creating a project starting from a CSV, a new cube with a node csv-mapping is created. 
2. The user uploads csv files. Is it possible to have several files
a) We have to use a stream to be able to store big files
b) The location of the file has to be attached to the node node csv-mapping
3. The user starts the mapping -> other task

**Buckets**
Are all files stored in the same bucket? As far as I understood yes.
If yes, with a folder per project? Would be useful when looking at the bucket

**API**
Since we have to use a stream.
Do we post a stream and the backend uploads the file and creates the needed nodes. How do we communicate the node
Or should there be a turnarond to the UI
1. Upload file and get location
2. Save location in Graph

**Graph**
Is there more than one file per csv-mapping?
Does every upload have node with link to the file location, and later on the mapping?

lucafurrer:
@tpluscode @martinmaillard @vhf These are the questions open for me with respect to the csv upload. 
Is there a need for a meeting or should we first discuss this here?

vhf:
Open questions to be discussed/approved

**Workflow**: 
1. *When creating a project starting from a CSV, a new cube with a node csv-mapping is created.*
  -> creating a project should not automatically create a CSV mapping but we might later on give the user the option to create a project with an initial CSV mapping
2. *The user uploads csv files. Is it possible to have several files* Yes
  a. We have to use a stream to be able to store big files
  b. The location of the file has to be attached to the node node csv-mapping
3. The user starts the mapping -> other task

**Buckets**
Are all files stored in the same bucket? As far as I understood yes.
If yes, with a folder per project? Would be useful when looking at the bucket

**API**
Since we have to use a stream.
Do we post a stream and the backend uploads the file and creates the needed nodes. How do we communicate the node
Or should there be a turnarond to the UI
1. Upload file and get location
[...]

PR #34, opened 2020-09-28
feat(api): CSV file upload

vhf:


apis/core/lib/domain/identifiers.ts
```diff
@@ -10,6 +9,6 @@ export function cubeProject(label: string): NamedNode {
   return $rdf.namedNode(`${env.API_CORE_BASE}cube-project/${url.slugify(label)}`)
 }
 
-export function csvMapping(project: GraphPointer<NamedNode>): NamedNode {
-  return $rdf.namedNode(`${project.value}/csv-mapping`)
+export function csvMapping(filename: string): NamedNode {
+  return $rdf.namedNode(`${env.API_CORE_BASE}csv-mapping/${url.slugify(filename)}`)
```
tpluscode, line 13: This change breaks `createProject` line 23.

The idea was that a project is created first and then the CSV Mapping resource to "dependent" on it. Hence the `project/FOO/csv-mapping` hierarchical URI. You would then upload the file to the mapping resource to create `project/FOO/csv-mapping/${filename}`.

apis/core/lib/domain/csv-mapping/create.ts
```diff
@@ -0,0 +1,26 @@
...
+  const filename = resource.out(dbo.filename).term!
+
+  await saveFile(filename.value, 'hello')
+
+  const csvMapping = clownface({ dataset: $rdf.dataset() })
+    .namedNode(id.csvMapping(filename.value))
+    .addOut(dbo.filename, filename)
+    .addOut(rdf.type, [cc.CsvMapping, hydra.Resource])
+
+  await store.put(csvMapping)
```
tpluscode, line 24: When you merge/rebase master you will be able to write

```suggestion
  const csvMapping = store.create(id.csvMapping(filename.value))
    .addOut(dbo.filename, filename)
    .addOut(rdf.type, [cc.CsvMapping, hydra.Resource])

  await store.save()
```

apis/core/lib/domain/identifiers.ts
tpluscode: Maybe we could handle project creation by CSV upload and create the project based off the CSV name?

| What | URI |
| -- | -- |
| File | `cube data.csv` |
| Project | `/project/cube-data-slug` |
| Mapping | `/project/cube-data-slug/csv-mapping` |
| Source | `/project/cube-data-slug/csv-mapping/cube-data.csv` |

All those would have to be created when a project is created from a CSV directly. @martinmaillard ?

apis/core/lib/domain/csv-mapping/create.ts
vhf: 👍

apis/core/lib/domain/identifiers.ts
vhf: It's definitely something we need to discuss to move forward with CSV uploads.

lucafurrer: I think it would be good to have a discussion about the data model and workflow as well how we store the the data on a S3 bucket. I will amend the corresponding issue #13

apis/core/lib/storage/s3.ts
```diff
@@ -0,0 +1,130 @@
...
+  }).promise()
+
+  if (file.Body) {
+    return file.Body.toString()
+  }
+  logError('Could not read file "%s" from S3.', path)
+  return null
+}
+
+export async function loadFileHeadString(path: string, uniqueLines = 20): Promise<string> {
```
lucafurrer, line 48: @vhf Do you think it is possible to extract the business logic and separate it from the S3 code

apis/core/lib/storage/s3.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (lines.length >= uniqueLines) {
+      lines.splice(uniqueLines)
+      break
+    }
+
+    nextByte += bytesToRead
+  }
+
+  return lines.join('\n')
+}
```
lucafurrer, line 130: @vhf Do you think it is possible to extract the business logic and separate it from the S3 code

apis/core/test/domain/csv/parse.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    const input1 = await loadFileString('test/CH_yearly_air_immission_basetable.csv')
+    const input2 = await loadFileHeadString('test/CH_yearly_air_immission_basetable.csv')
+
+    expect(input1).not.to.eq(input2)
+
+    const lines1 = input1?.split('\n') || []
+    const [firstLine2] = input2?.split('\n')
+    expect(lines1[0]).to.eq(firstLine2)
+  })
```
lucafurrer, line 46: These tests write to the filesystem and change the state of the system. We should either mock S3 or make the code testable without needing to write to S3 @vhf

martinmaillard:
One issue I noticed: if a file with the same name is uploaded twice, it creates two sources but the file is actually overwritten in S3. What should be the behavior here?
1. Store 2 files (use a generated name in S3)
2. Overwrite the source

tpluscode:
I think it should overwrite the source when users might want to update their input (changed format, new columns, etc)

martinmaillard:
Yeah, updating the source is a bit dangerous anyway (if it's already used in a mapping).

Error works for me. We may also overwrite the file and tell the user that it should have the same structure, otherwise it will break stuff.

lucafurrer:
> Yeah, updating the source is a bit dangerous anyway (if it's already used in a mapping).
> 
> Error works for me. We may also overwrite the file and tell the user that it should have the same structure, otherwise it will break stuff.

so error for the moment, or do you want to introduce an error?

martinmaillard:
If the API returns a validation error, I'll show it in the UI

martinmaillard:
Can we already merge this and work on the remaining tasks separately?

PR #36, opened 2020-09-28
Docker: Have some logs by default

sandhose:
This adds some logs by default. Having `*` is too verbose (especially because of the healthchecks), but at least having some logs when the app starts is nice.

@tpluscode If you have other log scopes to suggest, please do!

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -15,6 +15,7 @@ services:
       environment:
         TS_NODE_TRANSPILE_ONLY: "true"
         TS_NODE_PROJECT: apis/core/tsconfig.json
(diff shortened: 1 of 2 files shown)


tpluscode:
in local we have `creator*,hydra*,hydra-box*,labyrinth*`

sandhose:
> in local we have `creator*,hydra*,hydra-box*,labyrinth*`

Thanks, I set it to the same value, with a note saying it should be kept in sync between the Dockerfile and the lando config.

@vhf it's ready to merge if you want

vhf (approved):
👍

PR #37, opened 2020-09-28
Base URL workaround for OIDC URLs

sandhose:
Related to this conversation: https://zulip.zazuko.com/#narrow/stream/49-bafu-cube-creator/topic/frontend.20domains/near/113440

I'm using this branch to test the workaround live. This does not have to be a long-lasting solution, I can work on what I suggested on Zulip, I just needed this pushed somewhere to test that live.

PR #38, opened 2020-09-29
Proper runtime config in app Docker image

sandhose:
Supersede #37

This needs some modifications in the deployment at vshn so do not merge right away, but it can be reviewed

Now the app image is flexible enough to be run on separate domain or on the same domain in a subpath.


vhf (approved):
👍

martinmaillard:
If I understand correctly, the only variable that has to be set at build time is "PUBLIC_PATH". That means we won't have to build a separate image for each environment. If that's the case, this change looks good to me 👍

martinmaillard (approved):

Issue #39, opened 2020-10-05
Read and parse CSV files from S3


vhf:
- [ ] Read file
- [ ] Read only parts of the file (the backend shouldn't have to fetch the whole file to display a preview: CSV header + a few unique lines)
- [ ] Parse / options

PR #40, opened 2020-10-06
Read and parse CSV

vhf:
#39 

- [x] Sniff + Parse, options
- [x] Read file from S3
- [x] Read only parts of the file (the backend shouldn't have to fetch the whole file to display a preview: CSV header + a few unique lines)

I might move the fixtures to Minio's bucket later on.


apis/core/test/domain/csv/parse.test.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, it } from 'mocha'
+import { expect } from 'chai'
+import { promises as fs } from 'fs'
+import { resolve } from 'path'
+
+import { sniffParse } from '../../../lib/domain/csv'
+
+describe('domain/cube-projects/create', () => {
```
tpluscode, line 8: Looks like a mismatch between the test name and import/file. Copy/paste?

apis/core/lib/domain/csv/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  const records = await parse(csv, parserOptions)
+
+  const [columns, ...fileSample] = records
+
+  return {
+    header: columns,
+    rows: fileSample,
+  }
```
tpluscode, line 35: I think this should also return the detected dialect so that we can store it with the `Source` in the database

apis/core/lib/domain/csv/index.ts
```diff
@@ -0,0 +1,36 @@
...
+  return new Promise((resolve, reject) => CSVparse(csv, options, (err, records) => {
+    if (err) {
+      reject(err)
+      return
+    }
+    resolve(records)
+  }))
+}
+
+export async function sniffParse(csv: string) {
```
tpluscode, line 17: > I might move the fixtures to Minio's bucket later on.

I don't think we need that for unit tests of the `sniffParse` function. This would only be necessary if we have a request handler to re-run the sniffing. Such as in the scenario we discussed where the user would change the dialect or maybe upload a newer version of the file.

Instead I would suggest accepting a stream as input so that in the final code we will pass in an S3 stream and in tests a file stream.

```suggestion
export async function sniffParse(csv: Stream) {
```

tpluscode (commented):
Good going. I assume WIP but here're some early suggestions

apis/core/lib/domain/csv/index.ts
vhf: So `sniffParse` would take a stream, read the first chunk only, sniff and parse it, push the dialect and the chunk to the stream, then push the other chunks as they come?

tpluscode: Not exactly. My idea was to save the entire file first to S3 and then open in from the bucket to do the sniffing.

vhf: Right, that's what I was going for:

- Something to write the file to S3
- Something to read a whole file
- Something to read the beginning of a file <- this part will sniff and parse, the parsed part will be used for preview, the sniffed dialect will be stored and reused when reading the whole file

Issue #41, opened 2020-10-06
Store user information when creating project


tpluscode:
As per issue [158](https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/158) we need to keep information about the user who creates a project

The auth middleware will provide the user name in the `req.user` object in express handler. As discussed, I propose to prepend it with some base namespace URL like `${baseURI}/user/${username}`. Maybe use `dcterms:creator` for the relation.

DOD:

* [ ] Added creator object to the project resource

PR #46, opened 2020-10-09
feat: store user info when creating project

lucafurrer:
Add the creation user to a newly created cube project

Closing #41

### DIFF
--- a/apis/core/lib/domain/cube-projects/create.ts
+++ b/apis/core/lib/domain/cube-projects/create.ts
@@ -1,20 +1,22 @@
 import { GraphPointer } from 'clownface'
-import { rdfs, rdf, hydra } from '@tpluscode/rdf-ns-builders'
+import { rdfs, rdf, hydra, dcterms } from '@tpluscode/rdf-ns-builders'
 import { cc } from '@cube-creator/core/namespace'
 import { ResourceStore } from '../../ResourceStore'
-import { cubeProject, csvMapping } from '../identifiers'
+import { cubeProject, csvMapping, creatorUser } from '../identifiers'
 
 interface CreateProjectCommand {
   resource: GraphPointer
   store: ResourceStore
+  user: string
 }
 
-export async function createProject({ resource, store }: CreateProjectCommand): Promise<GraphPointer> {
+export async function createProject({ resource, store, user }: CreateProjectCommand): Promise<GraphPointer> {
   const label = resource.out(rdfs.label).term!
 
   const project = store.create(cubeProject(label.value))
     .addOut(rdfs.label, label)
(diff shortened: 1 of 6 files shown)


apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -1,20 +1,22 @@
...
+import { rdfs, rdf, hydra, dcterms } from '@tpluscode/rdf-ns-builders'
 import { cc } from '@cube-creator/core/namespace'
 import { ResourceStore } from '../../ResourceStore'
-import { cubeProject, csvMapping } from '../identifiers'
+import { cubeProject, csvMapping, creatorUser } from '../identifiers'
 
 interface CreateProjectCommand {
   resource: GraphPointer
   store: ResourceStore
+  user: string
```
tpluscode, line 10: I would also make the parameter a named node so that the project creation does not create new resources but depends on a concrete value

```suggestion
  creator: NamedNode
```

Maybe trivial but this way the user id becomes an opaque input. Even if we indirect the URI creation with `identifiers`

apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -1,20 +1,22 @@
 import { GraphPointer } from 'clownface'
-import { rdfs, rdf, hydra } from '@tpluscode/rdf-ns-builders'
+import { rdfs, rdf, hydra, dcterms } from '@tpluscode/rdf-ns-builders'
 import { cc } from '@cube-creator/core/namespace'
 import { ResourceStore } from '../../ResourceStore'
-import { cubeProject, csvMapping } from '../identifiers'
+import { cubeProject, csvMapping, creatorUser } from '../identifiers'
```
tpluscode, line 5: I think the command parameter and factory function should be somewhat reversed. 

```suggestion
import { cubeProject, csvMapping, user } from '../identifiers'
```

tpluscode (changes requested):
Looks good.

In addition to this I would ask you to add an `Expect` statement to [e2e-tests/cube-projects/create.hydra](https://github.com/zazuko/cube-creator/blob/a61552afa7c41775f71d8ec5bcf117714f17b12d/e2e-tests/cube-projects/create.hydra#L24-L30)

lucafurrer:
> Looks good.
> 
> In addition to this I would ask you to add an `Expect` statement to [e2e-tests/cube-projects/create.hydra](https://github.com/zazuko/cube-creator/blob/a61552afa7c41775f71d8ec5bcf117714f17b12d/e2e-tests/cube-projects/create.hydra#L24-L30)

Done with commit 4d1fa74

tpluscode (approved):

PR #47, opened 2020-10-09
refactor: added helper function to manage collections and resources

tpluscode:
This PR adds some handy tweaks and tricks for creating resources, collections and collection members

#### Resources are `rdf:type hydra:Resource` by default when created with `ResourceStore`

Controlled by adding parameter 

```typescript
store.create(resourceId, {
  implicitlyDereferencable: false,
})
```

#### Helper function to create collections

The `hydra:manages` is awkward. Instead, use a new helper. For example:

```typescript
import { manages } from './lib/resources/hydraManages'

store.create($rdf.namedNode('collection'))
  .addOut(...manages({
    property: rdf.type,
    object: schema.Person,
  }))
```

Above will create

```turtle
<collection> hydra:manages [
  hydra:property rdf:type ;
  hydra:object schema.Person
] .
```

#### Creating collection members with resources store

```typescript
const collectionId: NamedNode
const newMemberId: NamedNode

store.createMember(collectionId, newMemberId)
```

This will create triples according to the collection's `hydra:manages` statements


apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -1,24 +1,26 @@
...
-  const project = store.create(cubeProject(label.value))
-    .addOut(rdfs.label, label)
-    .addOut(rdf.type, [cc.CubeProject, hydra.Resource])
+  const project = await store.createMember(projectsCollection, cubeProject(label.value))
+
+  project.addOut(rdfs.label, label)
 
   if (resource.out(cc.projectSourceKind).value === 'CSV') {
     const mapping = store.create(csvMapping(project))
-      .addOut(rdf.type, [cc.CsvMapping, hydra.Resource])
```
lucafurrer, line 21: Why is the Resource not needed anymore?

lucafurrer (approved):
Looks fine for me

apis/core/lib/domain/cube-projects/create.ts
tpluscode: It became the default. Check the first point in the issue description

martinmaillard (approved):

PR #48, opened 2020-10-12
API redirects text/html requests to UI

vhf:


tpluscode (approved):

martinmaillard (approved):
Looks good to me.

One thing I couldn't figure out: where is the `/app` route defined?

tpluscode:
Well it's not. This will apply only to the deployed instance where `/app` is the UI running its own container

martinmaillard:
Ok but where is the redirection defined? Do we have a proxy in front of the API & App?

tpluscode:
> Ok but where is the redirection defined? Do we have a proxy in front of the API & App?

@sandhose ☝️

sandhose:
> > Ok but where is the redirection defined? Do we have a proxy in front of the API & App?
> 
> @sandhose ☝️

@tpluscode @martinmaillard There is `ingress-nginx` running in front that handles all HTTP traffic in the cluster, which consumes Ingress resources. In particular, routing for this app is done here: https://gitlab.ldbar.ch/vshn/gitops-prod/-/blob/master/zazuko/test/cube-creator/ingress.yaml

PR #50, opened 2020-10-12
feat(csv): get csv head via api

lucafurrer:
When making a request on a `CSVSource` with ContentType "text/csv", the header + 20 random(?) rows are returned.
For all other requests the usual Hydra Get is called and returns the node as triples.

`.addOut(cc.s3Key, key)`
Is also contained in the contentUrl, but I think it is better to store it explicitly. If anyone has a better existing property, I am open for suggestions.

### DIFF
--- a/apis/core/lib/domain/csv-source/upload.ts
+++ b/apis/core/lib/domain/csv-source/upload.ts
@@ -29,10 +29,13 @@ export async function uploadFile({
   const csvSource = store
     .create(id.csvSource(csvMapping, fileName))
     .addOut(schema.name, fileName)
-    .addOut(schema.contentUrl, $rdf.namedNode(upload.Location))
-    .addOut(cc.s3Key, key)
     .addOut(rdf.type, [cc.CSVSource, hydra.Resource])
     .addOut(cc.csvMapping, csvMapping)
+    .addOut(schema.associatedMedia, mediaObject => {
+      mediaObject.addOut(rdf.type, [schema.MediaObject, hydra.Resource])
+        .addOut(schema.identifier, key)
+        .addOut(schema.contentUrl, $rdf.namedNode(upload.Location))
+    })
 
   csvMapping.addOut(cc.csvSource, csvSource)
 
--- a/apis/core/lib/domain/csv-source/upload.ts
+++ b/apis/core/lib/domain/csv-source/upload.ts
@@ -32,7 +32,7 @@ export async function uploadFile({
     .addOut(rdf.type, [cc.CSVSource, hydra.Resource])
     .addOut(cc.csvMapping, csvMapping)
     .addOut(schema.associatedMedia, mediaObject => {
-      mediaObject.addOut(rdf.type, [schema.MediaObject, hydra.Resource])
+      mediaObject.addOut(rdf.type, schema.MediaObject)
         .addOut(schema.identifier, key)
         .addOut(schema.contentUrl, $rdf.namedNode(upload.Location))
     })
--- a/apis/core/lib/domain/csv-source/upload.ts
+++ b/apis/core/lib/domain/csv-source/upload.ts
@@ -30,6 +30,7 @@ export async function uploadFile({
     .create(id.csvSource(csvMapping, fileName))
     .addOut(schema.name, fileName)
     .addOut(schema.contentUrl, $rdf.namedNode(upload.Location))
+    .addOut(cc.s3Key, key)
     .addOut(rdf.type, [cc.CSVSource, hydra.Resource])
     .addOut(cc.csvMapping, csvMapping)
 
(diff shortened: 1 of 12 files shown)


vhf:
> the header + 20 random(?) rows are returned.

https://github.com/zazuko/cube-creator/blob/fc633f919d2365905c7caba3c85e60928ef8f6f8/apis/core/lib/storage/s3.ts#L48

20 unique lines including the header

tpluscode (changes requested):

apis/core/lib/handlers/csv-source.ts
```diff
@@ -33,3 +36,25 @@ export const post = protectedResource(
     await res.dataset(fileLocation.dataset)
   }),
 )
+
+export const getCSVSource = labyrinth.protectedResource(
```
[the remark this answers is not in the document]
lucafurrer: No

apis/core/lib/domain/csv-source/upload.ts
```diff
@@ -30,6 +30,7 @@ export async function uploadFile({
     .create(id.csvSource(csvMapping, fileName))
     .addOut(schema.name, fileName)
     .addOut(schema.contentUrl, $rdf.namedNode(upload.Location))
+    .addOut(cc.s3Key, key)
```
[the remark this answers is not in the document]
lucafurrer: I like this idea!

apis/core/lib/handlers/csv-source.ts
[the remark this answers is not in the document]
lucafurrer: No

lucafurrer:
To make the unit-test run properly I made a refactoring for loadFileHeadString
- moved it away from the class s3
- takes a stream as input
```TypeScript
export function loadFileHeadString(
  stream: Readable,
  uniqueLines = 20,
): Promise<string> 
```

The access to S3 is not covered by unit tests anymore. In my opinion this has to be done through integration testing and is covered by the e2e tests.

apis/core/lib/handlers/csv-source.ts
```diff
@@ -37,10 +38,10 @@ export const post = labyrinth.protectedResource(
   }),
 )
 
-export const getCSVSource = labyrinth.protectedResource(
+const getCSVSource = labyrinth.protectedResource(
   shaclValidate,
   asyncMiddleware(async (req, res, next) => {
-    if (req.headers['content-type'] !== 'text/csv') {
+    if ((!req.headers['content-type'] || !is(req.headers['content-type'], 'text/csv'))) {
```
tpluscode, line 44: This would be the simplest approach
```suggestion
    if (!typeis(req, 'text/csv')) {
```

apis/core/lib/handlers/csv-source.ts
```diff
@@ -1,10 +1,11 @@
 import asyncMiddleware from 'middleware-async'
 import * as labyrinth from '@hydrofoil/labyrinth/resource'
+import { parse } from 'content-disposition'
+import clownface from 'clownface'
+import { is } from 'type-is'
```
tpluscode, line 5: If you default-import

```suggestion
import typeis from 'type-is'
```

apis/core/lib/domain/identifiers.ts
```diff
@@ -21,3 +21,7 @@ export function csvSourceCollection(csvMapping:GraphPointer<NamedNode>): NamedNo
 export function csvSource(project: GraphPointer<NamedNode>, fileName: string): NamedNode {
   return $rdf.namedNode(`${project.value}/csv-source/${url.slugify(fileName)}`)
 }
+
+export function associatedMedia(csvSource: GraphPointer<NamedNode>): NamedNode {
+  return $rdf.namedNode(`${csvSource.value}/media/`)
+}
```
tpluscode, line 27: Is this used?

Besides, I think that we can just keep a blank node for the `MediaObject`

apis/core/test/domain/csv/parse.test.ts
```diff
@@ -17,31 +17,19 @@ describe('domain/csv/parse', () => {
...
-
-  it('reads parts of a file from s3', async () => {
+  it('reads parts of a file ', async () => {
     const path = resolve(__dirname, '../../fixtures/CH_yearly_air_immission_basetable.csv')
     const fileContent = await fs.readFile(path)
-    await saveFile('test/CH_yearly_air_immission_basetable.csv', fileContent.toString())
 
-    const input1 = await loadFileString('test/CH_yearly_air_immission_basetable.csv')
-    const input2 = await loadFileHeadString('test/CH_yearly_air_immission_basetable.csv')
+    const input1 = fileContent.toString()// await loadFileString('test/CH_yearly_air_immission_basetable.csv')
```
tpluscode, line 24:
```suggestion
    const input1 = fileContent.toString()
```

apis/core/test/domain/csv/parse.test.ts
```diff
@@ -1,9 +1,9 @@
 import { describe, it } from 'mocha'
 import { expect } from 'chai'
-import { promises as fs } from 'fs'
+import fss, { promises as fs } from 'fs'
```
tpluscode, line 3: I don't see `fss` used, do I?

PR #53, opened 2020-10-13
feat: add post-logout page

### LINKED ISSUE #52: Logout button looks broken because of auto-login
The logout button works, but the user is then immediately automatically logged in again.

martinmaillard:
Will close #52

### DIFF
--- a/ui/src/views/Logout.vue
+++ b/ui/src/views/Logout.vue
@@ -1,0 +1,35 @@
+<template>
+  <section class="section">
+    <b-message type="is-success">
+      <p>
+        You have been successfully logged out.
+      </p>
+      <p>
+        <router-link :to="{ name: 'Home' }">
+          Back to home
+        </router-link>
+      </p>
+    </b-message>
+  </section>
+</template>
+
+<script lang="ts">
+import Vue from 'vue'
+import { mapGetters } from 'vuex'
+
+export default Vue.extend({
+  name: 'Logout',
+
+  mounted () {
+    if (this.isAuthenticated) {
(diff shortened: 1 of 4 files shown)


sandhose (approved):
LGTM.

On why exactly it fixes the issue: SAML and OIDC specs specify two types of logout, frontchannel and backchannel.

`signOutOidcSilent` does a backchannel logout (without redirecting the user), but eIAM does not support it, so keycloak can't perform a logout on eIAM that way. 
`signOutOidc` is a frontchannel logout (by redirecting the user), which does chained logouts from other apps and identity providers, and because it is supported by eIAM, it should log out from eIAM.

ui/src/views/Logout.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <section class="section">
+    <b-message type="is-success">
+      <p>
+        You have been successfully logged out.
+      </p>
+      <p>
+        <router-link :to="{ name: 'Home' }">
+          Back to home
```
vhf, line 9: Following this link will automatically sign you in again, right? If so I'm not sure about keeping this link. Usually my expectation when I sign out and go back to home is to see the home as if I was not signed in.

Another option would be to change the text to something explicit like "Sign in again"?

vhf (approved):

ui/src/views/Logout.vue
martinmaillard: Ok I can change the text, as we don't have a "public" home page (other than the keycloak login page)

tpluscode (approved):
👍

PR #54, opened 2020-10-13
feat: redirect home to cube-projects

martinmaillard:
We don't have a homepage yet. Auto-redirect to the projects list.


vhf (approved):
👍

PR #69, opened 2020-10-15
feat(csv): error if file added twice to same project

### LINKED ISSUE #61: CSV upload: error when uploading the same file twice
The API should return a validation error when trying to upload a file with the same name again and the UI should show it.

lucafurrer:
fixes #61 

When a file with the same name is uploaded twice an error 409 Conflict is thrown.

To test if this is the case, it is checked if there exists an other CSVSource with the same schema:name.

```TypeScript 
const result = await ASK`
      GRAPH ${csvMapping} 
      {
        ${csvMapping} ${cc.csvSource} ?source
      }
      GRAPH ?source
      {
          ?source ${schema.name} "${fileName}"
      }
      `
    .execute(client.query)
```

I added a new folder structure in domain for sparql queries.

### DIFF
--- a/apis/core/lib/domain/queries/csv-source.ts
+++ b/apis/core/lib/domain/queries/csv-source.ts
@@ -1,0 +1,21 @@
+import { NamedNode } from 'rdf-js'
+import { ASK } from '@tpluscode/sparql-builder'
+import { cc } from '@cube-creator/core/namespace'
+import { schema } from '@tpluscode/rdf-ns-builders'
+import { client } from './query-client'
+
+export async function sourceWithFilenameExists(csvMapping: NamedNode, fileName: string): Promise<boolean> {
+  const result = await ASK`
+      GRAPH ${csvMapping} 
+      {
+        ${csvMapping} ${cc.csvSource} ?source
+      }
+      GRAPH ?source
+      {
+          ?source ${schema.name} "${fileName}"
+      }
+      `
+    .execute(client.query)
+
+  return result
+}
--- a/apis/core/lib/domain/queries/csv-source.ts
+++ b/apis/core/lib/domain/queries/csv-source.ts
@@ -5,17 +5,15 @@ import { schema } from '@tpluscode/rdf-ns-builders'
 import { client } from './query-client'
 
 export async function sourceWithFilenameExists(csvMapping: NamedNode, fileName: string): Promise<boolean> {
-  const result = await ASK`
+  return await ASK`
       GRAPH ${csvMapping} 
       {
         ${csvMapping} ${cc.csvSource} ?source
       }
       GRAPH ?source
       {
-          ?source ${schema.name} "${fileName}"
+        ?source ${schema.name} "${fileName}"
       }
       `
     .execute(client.query)
-
-  return result
 }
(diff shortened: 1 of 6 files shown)


e2e-tests/csv-source/upload.hydra
```diff
@@ -57,19 +57,31 @@ With Class cc:CubeProject {
                     }
                 }
 
-                Invoke {
+                                Invoke {
```
vhf, line 60: Identation seems off here

vhf (approved):
👍

apis/core/lib/domain/queries/csv-source.ts
```diff
@@ -5,17 +5,15 @@ import { schema } from '@tpluscode/rdf-ns-builders'
 import { client } from './query-client'
 
 export async function sourceWithFilenameExists(csvMapping: NamedNode, fileName: string): Promise<boolean> {
-  const result = await ASK`
+  return await ASK`
```
vhf, line 8:
```suggestion
  return ASK`
```

since it's not wrapped in try/catch, we can catch this on the caller side if need be

e2e-tests/csv-source/upload.hydra
```diff
@@ -59,17 +59,29 @@ With Class cc:CubeProject {
...
+                Invoke {
+                    Content-Type "text/csv"
+                    Content-Disposition "file; filename='test.jpg'"
+
+                    ```
+                    A,B,C
+                    1,2,test
+                    ```
+                } => {
+                    Expect Status 409
```
martinmaillard, line 83: I never used 409, is this a usual way of telling the user that a file already exists?

lucafurrer: @tpluscode suggested it and I liked it. If you have an other preference, tell me :wink: I am not attached to it.

lucafurrer: By the way I have not touched the front end for the error handling (yet?)

martinmaillard: I pushed a change for the UI

martinmaillard (approved):

martinmaillard:
One weird thing I noticed is that I get the 409 error really fast, but then the Hydra client tries to load `/api` and this second request takes a long time. The fact that it queries `/api` is unrelated to this issue, but it hints to a possibility that the backend process is still busy handling the file upload? Is this possible?

e2e-tests/csv-source/upload.hydra
tpluscode: > [The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource. This code is used in situations where the user might be able to resolve the conflict and resubmit the request. The server SHOULD generate a payload that includes enough information for a user to recognize the source of the conflict.](http://tools.ietf.org/html/rfc7231#section-6.5.8)

We could address the `SHOULD` clause by linking to the already existing resource. In such case the client could direct the user to make a `PUT` instead to update the source. 

I would have the as a separate PR though. Not only we do not have the update operation yet, it would also be nice to make `problem+json` payloads proper RDF

tpluscode:
Let me address the `/api` thing in separate change. We are not changing it while the app is running and might as well skip subsequent calls. This is now a feature of Alcaeus

PR #70, opened 2020-10-15
feat: Extract columns

### LINKED ISSUE #62: CSV: extract columns and sample data
When uploading a new CSV, the backend should extract a list of columns and link them to the source. It should also probably extract a list of unique values for each column.

**DoD**

* [ ] for every column from the csv, create a `cc:CSVColumn` resource
* [ ] `cc:CSVColumn` resources should be stored in the source's graph
    * [ ] annotated sort order of columns
* [ ] if extracting fails, it should not prevent the request from succeeding (file saved to S3, source graph created)
* [ ] sample values stored with each `cc:CSVColumn`

**Example**

Given a `/source/employees.csv` `CSVSource` we should in its very graph create column resources when it is created:

```turtle
graph </source/employees.csv> {
  </source/employees.csv> 
    cc:csvColumn </source/employees.csv/column-id-slug> , 
              </source/employees.csv/column-name-slug> , 
              </source/employees.csv/column-salary-slug> .

  </source/employees.csv/column-id-slug>
    schema:name "id" ;
    cc:csvColumnSample "123", "456", "789" ;
    dtype:order 0 ;
  .

  </source/employees.csv/column-salary-slug>
    schema:name "salary" ;
    cc:csvColumnSample "5k EUR", "5000 AUD", "$6000" 
    dtype:order 1 ;
  .

  # the rest of </source/employees.csv> and its columns
}
```

lucafurrer:
Read columns from CSV File and some sample data and save them to the graph attached to the correspending csv-source

fix #62

### DIFF
--- a/apis/core/lib/domain/csv-source/upload.ts
+++ b/apis/core/lib/domain/csv-source/upload.ts
@@ -67,7 +67,7 @@ export async function uploadFile({
           .addOut(schema.name, columnName)
           .addOut(dtype.order, index)
         rows.forEach(row => {
-          if (row.length() > index) {
+          if (row.length > index) {
             column.addOut(cc.csvColumnSample, row[index])
           }
         })
--- a/apis/core/lib/domain/csv-source/upload.ts
+++ b/apis/core/lib/domain/csv-source/upload.ts
@@ -63,7 +63,7 @@ export async function uploadFile({
     for (let index = 0; index < header.length; index++) {
       const columnName = header[index]
       csvSource.addOut(csvw.column, id.column(csvSource, columnName), column => {
-        column.addOut(rdf.type, [csvw.Column, hydra.Resource])
+        column.addOut(rdf.type, [csvw.Column])
           .addOut(schema.name, columnName)
           .addOut(dtype.order, index)
         rows.forEach(row => {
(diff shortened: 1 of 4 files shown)


apis/core/lib/domain/csv-source/upload.ts
```diff
@@ -59,6 +59,20 @@ export async function uploadFile({
...
+
+    for (let index = 0; index < header.length; index++) {
+      const columnName = header[index]
+      csvSource.addOut(csvw.column, id.column(csvSource, columnName), column => {
+        column.addOut(rdf.type, [csvw.Column])
+          .addOut(schema.name, columnName)
+          .addOut(dtype.order, index)
+        rows.forEach(row => {
+          if (row.length > index) {
+            column.addOut(cc.csvColumnSample, row[index])
```
martinmaillard, line 71: This is good enough for a start, but it would be great to have a sample of multiple unique values of each column.

martinmaillard (approved):

vhf (approved):

tpluscode (approved):

apis/core/lib/domain/csv-source/upload.ts
lucafurrer: Was thinking about this as well. We might also have to reiterate over the extract sample code. Id columns might screw this up.

PR #72, opened 2020-10-16
refactor: use query-client of request

lucafurrer:
Reuse the client provided by the request to execute the SPARQL-query to check if there exists already a file with the same name


martinmaillard (approved):

vhf (approved):

tpluscode:
Having discussed with @lucafurrer I pushed this refactoring one step further, and in the opposite direction.

We were unsatisfied about how the sprawl client and resource store has to be forwarded from top to bottom. 

By making the parameters optional with default value imported from a shared module it is possible to simplify the calls without sacrificing testability.

vhf (approved):

martinmaillard (approved):

PR #73, opened 2020-10-16
use SHACL form component

tpluscode:
I propose to use a form element which builds UI based on SHACL shapes

The shapes are provided by Hydra using the `hydra:expects` property

The UI is defined using [DASH](http://datashapes.org/dash.html) vocabulary. It comes with a handful of editors (identified by URIs). They are selected based on the actual value and Property Shape

In this PR I added our renderings using buefy of a ten box and radio buttons.

---

Slightly unrelated, I refactored the create project request to require one of two IRIs instead of plain text so that additional labels can be displayed on the UI


apis/core/bootstrap/shapes/cube-project.ts
```diff
@@ -1,10 +1,10 @@
-import { cc } from '@cube-creator/core/namespace'
+import { cc, shape } from '@cube-creator/core/namespace'
 import { hydra, rdfs, sh } from '@tpluscode/rdf-ns-builders'
 import { turtle } from '@tpluscode/rdf-string'
 
 export const CubeProjectShape = turtle`
-<shape/cube-project/create> {
-  <shape/cube-project/create> a ${sh.NodeShape}, ${hydra.Resource} ;
+${shape('cube-project/create')} {
```
lucafurrer, line 6: Is it possible to somehow use the predefined strings from '@cube-creator/core/namespaces/shapes.ts'. here

apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -26,7 +26,7 @@ export async function createProject({
   project.addOut(rdfs.label, label)
     .addOut(dcterms.creator, user)
 
-  if (resource.out(cc.projectSourceKind).value === 'CSV') {
+  if (shape('cube-project/create#CSV').equals(resource.out(cc.projectSourceKind).term)) {
```
lucafurrer, line 29: It would be really great to not have strings but a variable here for `'cube-project/create#CSV'` as well :smile:

ui/src/forms/matchers.ts
```diff
@@ -0,0 +1,19 @@
...
+export const projectKindSelector: SingleEditor = {
+  term: editor.RadioButtons,
+  match (shape) {
+    const { in: choices, path } = shape
+    if (choices.length === 0) {
+      return 0
+    }
+    if (choices.length > 5) {
+      return null
+    }
```
lucafurrer, line 14: Do I understand this right that you can not have more than 5 options?

lucafurrer (commented):
I cannot say I understood everything, but it looks nice!

ui/src/store/modules/cube-projects.ts
```diff
@@ -63,23 +63,6 @@ const actions: ActionTree<CubeProjectsState, RootState> = {
...
-      '@type': 'CubeProject',
-      // Must pass this ID for now because of a bug in rdfjs/express-handler
-      '@id': '',
-      [rdfs.label.value]: formData.label,
-      projectSourceKind: formData.projectSourceKind,
-    }
-
-    return api.invokeSaveOperation(operation, data)
-  },
-
```
tpluscode, line 82: Most importantly, this is removed in favor of RDF which the form creates

ui/src/store/modules/api.ts
```diff
@@ -19,6 +21,11 @@ const actions: ActionTree<APIState, RootState> = {
     const entrypoint = await api.fetchResource('/')
     context.commit('storeEntrypoint', entrypoint)
     return entrypoint
+  },
+  async invokeSaveOperation (context, { operation, resource }: {operation: RuntimeOperation; resource: RdfResource | GraphPointer<ResourceIdentifier>}) {
+    const data = 'toJSON' in resource ? resource : RdfResourceImpl.factory.createEntity(resource) as RdfResource
+
+    return api.invokeSaveOperation(operation, data)
```
tpluscode, line 28: A single point for executing operations, using `toJSON` to serialize an refines RdfResouce as simple JSON-LD

ui/src/forms/metadata.ts
```diff
@@ -0,0 +1,7 @@
+import { dataset, literal, quad } from '@rdf-esm/dataset'
+import { editor } from '@/forms/bulma'
+import { rdfs } from '@tpluscode/rdf-ns-builders'
+
+export const Metadata = dataset([
+  quad(editor.RadioButtons, rdfs.label, literal('Radio buttons'))
+])
```
tpluscode, line 7: DASH editor metadata is RDF

ui/src/forms/matchers.ts
tpluscode: If you look at DASH, it predefines a bunch of editors in its namespace. There is an enum selector which applies to `sh:in`. The spec wants it to render a dropdown. 

Here we give a higher score to enum of `<5` choices to prefer a radio button component.

ui/src/forms/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  focusNode: MaterialRenderStrategy.focusNode(DefaultStrategy.focusNode),
+})
+components.pushComponents(Components)
+
+editors.addMatchers(Matchers)
+editors.addMetadata(Metadata)
+
+class Form extends ShaperoneForm {
+  protected createRenderRoot (): Element | ShadowRoot {
+    return this
```
tpluscode, line 22: Disables Shadow DOM

ui/src/forms/components.ts
```diff
@@ -0,0 +1,29 @@
...
+    const items = property.shape.pointer.node(property.shape.in)
+
+    return html`<b-radio .options="${items}" .value="${value.object}" .update="${update}"></b-radio>`
+  },
+  loadDependencies () {
+    return [
+      import('./BulmaRadioButtons.vue').then(createCustomElement('b-radio'))
+    ]
+  }
+}
```
tpluscode, line 29: This is how component rendering is hooked up with the form component.

If a matching component is not customised, a browser-native element would render

ui/src/forms/BulmaRadioButtons.vue
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  label (value: Term) {
+    return this.options?.node(value).out(rdfs.label).values[0] || value.value
+  }
+
+  emit (value: string) {
+    this.message = this.options?.namedNode(value).out(rdfs.comment).values[0] || ''
+    this.update(value)
+  }
```
tpluscode, line 44: By trying `rdfs.label` and `rdfs.comment` we display hints where enum elements are named nodes. Strings are just strings

packages/core/namespaces/shapes.ts
```diff
@@ -0,0 +1,10 @@
+import namespace from '@rdf-esm/namespace'
+import { NamedNode } from 'rdf-js'
+import { baseUri } from '../env'
+
+type ShapeTerms =
+  'cube-project/create' | 'cube-project/create#CSV' | 'cube-project/create#ExistingCube'
+
+type ShapesNamespace = (term: ShapeTerms) => NamedNode
+
+export const shape: ShapesNamespace = namespace(baseUri + 'shape/') as any
```
tpluscode, line 10: Like with `cc`, by defining a `shape` namespace builder we prevent typos where it should be used instead of magic string URIs

apis/core/bootstrap/shapes/cube-project.ts
tpluscode: Precisely, the type declaration prevents strings outside of those declared in `shapes.ts`

apis/core/lib/domain/cube-projects/create.ts
tpluscode: The string also gets type-checked. Typos will raise TS errors

lucafurrer: Code completion and this. That's fine with me then. If I only have a restricted set of possibilities it is always nice to not have to go to the source to know what is allowed.

lucafurrer:
@tpluscode Do you think there is a way to avoid the "Click to add value" and render the fields in a clicked state?

image

tpluscode:
@lucafurrer definitely. we've been discussing this with @martinmaillard too

The nature of SHACL is to define a data structure in open terms. A property can have 0..N values and also no notion of "required" fields. An RDF property with no presence in a graph has no objects and thus no editor to render.

In a form like this it would be easily solved by initialising the form data with blank values

lucafurrer:
> In a form like this it would be easily solved by initialising the form data with blank values

In this case we I think we should do this. As an end user I would be confused about this click to add value.

ui/public/index.html
```diff
@@ -21,6 +21,8 @@
       };
     </script>
     <% } %>
+    <script>try{new EventTarget}catch(e){document.write('<script src="https://unpkg.com/@ungap/event-target@0.1.0/min.js"><\x2fscript>')}</script>
+    <link href="https://fonts.googleapis.com/css?family=Material+Icons&display=block" rel="stylesheet">
```
martinmaillard, line 25: Any way we can avoid a separate set of icons? By re-implementing a set of components for shaperone for example?

ui/src/forms/matchers.ts
```diff
@@ -0,0 +1,19 @@
+import { cc } from '@cube-creator/core/namespace'
+import type { SingleEditor } from '@hydrofoil/shaperone-core'
+import { editor } from '@/forms/bulma'
+
+export const projectKindSelector: SingleEditor = {
```
martinmaillard, line 5: Why is it specific to `projectKind`? Having code that is domain-specific in the matchers will be very confusing. Any way we can hard-set the component to use directly in the shape?

ui/src/store/modules/api.ts
martinmaillard: If it's not modifying the state in any way, I don't think it should be in the store. But from our discussion last week, I gather that you would actually manage forms state in a central place?

ui/src/views/CubeProjectNew.vue
```diff
@@ -63,22 +46,30 @@ export default Vue.extend({
...
-        return 'Add metadata to a Cube resulting of another pipeline'
+  async mounted () {
+    for (const expects of this.operation?.expects) {
+      if ('load' in expects && expects.types.has(sh.Shape)) {
+        const { representation } = await expects.load()
+        const { root: shape } = representation
+        this.shapes = Object.freeze(dataset([...shape?.pointer.dataset
+          .match(null, null, null, shape.id)]
+          .map(({ subject, predicate, object }) => quad(subject, predicate, object))))
+        break
```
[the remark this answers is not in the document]
martinmaillard: It looks like this code is generic-enough to be extracted in a function, right?

I don't understand why `break` though 😄

ui/src/views/CubeProjectNew.vue
```diff
@@ -35,23 +20,21 @@
...
-      project: {
-        projectSourceKind: 'CSV',
-        label: '',
-      },
-      error: null,
-      projectSourceKinds: [
-        { value: 'CSV', label: 'CSV file(s)' },
-        { value: 'Existing cube', label: 'Existing cube' },
-      ],
+      resource: Object.freeze(clownface({ dataset: dataset() }).namedNode('')),
```
martinmaillard, line 35: You said `Object.freeze` is to prevent vue to react to observe changes in the dataset? I'm curious to see how this works

martinmaillard (commented):
There are some code-organisation choices that I would like to change in the UI code, but I can do it later. This is not critical.

ui/src/forms/BulmaRadioButtons.vue
tpluscode: Of course! I don't know why it wasn't obvious to me

ui/src/forms/matchers.ts
tpluscode: I guess initially I want an example of a very specific matcher but radio buttons hardly are unique UI component, I agree 👌

ui/src/store/modules/api.ts
tpluscode: Yes, this would not represent a complete solution yet. Main reason to have this now is only to replace the `create` action I removed from `cube-projects.ts`

That said, there definitely would be some state change involved. At least to toggle a `loading` flag which would show/hide some kind spinner

ui/src/views/CubeProjectNew.vue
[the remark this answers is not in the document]
tpluscode: Yes, generic but should also be greatly simplified. The missing piece is now zazuko/clownface#47

> I don't understand why break though

Huh, neither do I 😆

ui/src/views/CubeProjectNew.vue
tpluscode: I think Vue simply check `Object.isFrozen` and does not add any observers if true

ui/public/index.html
tpluscode: Right now the material design rendering code hard-codes material icons (via `<mwc-icon>`) element.

Yes, It could be part of a refactoring I want in how the rendering is composed to have small pieces of the UI replaced

ui/src/forms/metadata.ts
martinmaillard: What is this metadata?

tpluscode: Right now it's just labels and the editor type (can be `dash:SingleEditor` or `dash:MultiEditor`)

ui/src/views/CubeProjectNew.vue
```diff
@@ -53,10 +53,10 @@ export default Vue.extend({
       if ('load' in expects && expects.types.has(sh.Shape)) {
         const { representation } = await expects.load()
         const { root: shape } = representation
-        this.shapes = Object.freeze(dataset([...shape?.pointer.dataset
-          .match(null, null, null, shape.id)]
-          .map(({ subject, predicate, object }) => quad(subject, predicate, object))))
-        break
+        if (shape) {
+          this.shapes = shape.pointer
+          break
```
martinmaillard, line 58: Break 😆

tpluscode: it is a loop tho

lucafurrer: You could use a filter or similar to avoid the break...

martinmaillard (approved):

PR #76, opened 2020-10-19
feat(ui): display CSV columns with sample values

### LINKED ISSUE #14: Mapper: display CSV file structure
The mapper needs to show the structure of a CSV file.

It shows:
- The name of the CSV file
- The name of each column
- For each column, a small sample of the values it contains

Screenshot 2020-09-15 at 16 16 29

Depends on #62

martinmaillard:
Will close #14

### DIFF
--- a/ui/src/api/index.ts
+++ b/ui/src/api/index.ts
@@ -8,6 +8,7 @@ import ProjectMixin from './mixins/Project'
 import CSVMappingMixin from './mixins/CSVMapping'
 import CSVSourceCollectionMixin from './mixins/CSVSourceCollection'
 import CSVSourceMixin from './mixins/CSVSource'
+import CSVColumnMixin from './mixins/CSVColumn'
 
 const rootURL = window.APP_CONFIG.apiCoreBase
 const segmentSeparator = '!!' // used to replace slash in URI to prevent escaping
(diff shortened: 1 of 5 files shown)


ui/src/api/index.ts
```diff
@@ -25,6 +26,7 @@ Hydra.resources.factory.addMixin(ProjectMixin)
 Hydra.resources.factory.addMixin(CSVMappingMixin)
 Hydra.resources.factory.addMixin(CSVSourceCollectionMixin)
 Hydra.resources.factory.addMixin(CSVSourceMixin)
+Hydra.resources.factory.addMixin(CSVColumnMixin)
```
tpluscode, line 29: As this keeps growing, you might combine all mixins in a single module and add them all at once

```js
import * as Mixins from './mixins'

Hydra.resources.factory.addMixin(...Object.values(Mixins))
```

tpluscode (approved):

ui/src/api/index.ts
martinmaillard: Doesn't that mean I still have to list all the mixins in `mixins/index.ts`?
And TypeScript is not too happy about this, as it doesn't know the type of the values of `Mixins`.

tpluscode: Yes, would still have to `export` from `index.ts`.

It's just a suggestion. Feel free to ignore :)

martinmaillard: I'll keep it here for now, but will probably extract at some point

PR #77, opened 2020-10-19
CSVW data cube pipeline

tpluscode:
This PR starts the pipeline code for transforming CSVs into a data cube

* [x] clean up pipeline code
* [x] implement integration tests for pipeline
* [ ] have pipeline write the `cube:Cube` back to the database


cli/test/support/testData.ts
```diff
@@ -0,0 +1,85 @@
+import { csvw, hydra, rdf, schema } from '@tpluscode/rdf-ns-builders'
```
tpluscode, line 1: This is an interesting part, where I put some test project and other necessary resources to the triple store

cli/test/support/testData.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export const insertTestData = async () => {
+  await removeTestGraphs()
+
+  return INSERT.DATA`
+  graph </project/cli-test> {
+      </project/cli-test>
+          a ${hydra.Resource} , ${cc.CubeProject} ;
+          ${cc.csvMapping} </project/cli-test/mapping> ;
+          ${cc.cube}       </cube/cli-test>
```
tpluscode, line 37: I assume a `cc:cube` which would be the actual instance of project's `qb:Cube`

Now, currently observations will be pushed into its graph but I reconsider and I would propose to have them separated

```sparql
graph <project/foo/cube> {
  <project/foo/cube> a qb:Cube .
}

graph <project/foo/cube/observation> {
   # pipeline output here
}
```

This way we can easily reject the previous transformation result but keep the cube itself

cli/test/support/testData.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  graph </project/cli-test/mapping/table/foo> {
+      </project/cli-test/mapping/table/foo>
+          a ${hydra.Collection}, ${cc.Table} ;
+          ${cc.csvw}       </project/cli-test/mapping/table/foo/csvw> ;
+          ${cc.csvMapping} </project/cli-test/mapping> ;
+          ${cc.csvSource}  </project/cli-test/source/foo> ;
+          ${schema.name}   "Foo table" ;
+      .
+  }
```
tpluscode, line 69: @martinmaillard pleas compare with your work on tables

cli/package.json
```diff
@@ -0,0 +1,76 @@
+{
+  "name": "@cube-creator/cli",
+  "version": "0.0.0",
+  "description": "Runs a pipeline to create an RDF data cube",
+  "main": "index.js",
+  "scripts": {
+    "postinstall": "mkdir -p output",
+    "transform": "dotenv -e ./.env -e ./.test.env -- yarn start",
+    "start": "ts-node index.ts transform --project \"$TRANSFORM_PROJECT\" --to stdout --debug --enable-buffer-monitor",
```
tpluscode, line 9: The run command is simplified where everything is by default provided as environment variables

.github/workflows/ci.yaml
```diff
@@ -13,6 +13,23 @@ jobs:
...
+      - name: Install lando
+        run: sh e2e-tests/install-lando.sh
+      - name: Start site
+        run: sh e2e-tests/start-lando.sh
+      - run: yarn install --ci
+      - run: yarn test:cli
+      - name: logs on fail
+        if: ${{ failure() }}
+        run: lando logs -s core
+
```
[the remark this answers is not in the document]
martinmaillard: I don't mind having a separate step

apis/core/lib/domain/table/csvw.ts
```diff
@@ -27,13 +30,55 @@ export async function createCsvw({
...
+        types: [csvw.Column],
+        suppressOutput: true,
+        title: 'limitvalue',
+      }, {
+        types: [csvw.Column],
+        suppressOutput: true,
+        title: 'value_remark',
+      }],
+    },
+  })
```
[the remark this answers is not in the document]
martinmaillard: I guess it's temporary test data that will guide us to implement the real thing?

packages/model/CsvMapping.ts
```diff
@@ -1,6 +1,25 @@
...
 import { RdfResourceCore } from '@tpluscode/rdfine/RdfResource'
+import { Constructor, property } from '@tpluscode/rdfine'
+import { Collection } from '@rdfine/hydra'
+import { CsvSource } from './CsvSource'
+import { Table } from './Table'
+import { cc } from '@cube-creator/core/namespace'
 
 export interface CsvMapping extends RdfResourceCore {
   sources: CsvSource[]
+  tables: Collection<Table>
```
martinmaillard, line 10: It's a bit weird to have a collection in one case and an array in the other. Is it just because we'll unify the sources in the future?

martinmaillard (approved):

packages/model/CsvMapping.ts
tpluscode: Collection is not a language thing, it's a separate RDF resource.
While the array is simply multiple objects of a property.

We could have tables also directly associated with the mapping resource, not problem there.

apis/core/lib/domain/table/csvw.ts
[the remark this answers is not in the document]
tpluscode: Correct. A CSVW structure like this above will have to be generated from the user mappings between sources and tables

packages/model/CsvMapping.ts
martinmaillard: The only issue I have is naming. It would be less confusing to have either `tables: Table[]` + `tableCollection: Collection<Table>` or change `sources` to also return a collection.

tpluscode: To be honest, this should actually be just the identifier because the collection will never be reachable from a CSV mapping's graph. So preferably

```
tableCollection: RdfResource
```

or something like this, which doesn't exist at the moment but add a little strong typing

```
tableCollection: Link<Collection<Table>>
```

cli/lib/s3.ts
```diff
@@ -0,0 +1,69 @@
+import aws from 'aws-sdk'
```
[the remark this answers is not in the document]
lucafurrer: Is it? I see usage all over the place

lucafurrer (approved):
:+1:

packages/model/CsvMapping.ts
tpluscode: Ok, I changed exactly like that

PR #78, opened 2020-10-20
Extract unique values per column

vhf:
Fixing https://github.com/zazuko/cube-creator/pull/70#discussion_r507517264 .

Since the frontend code seems to only care about 3 unique values per row that's what I went for as well but there's a param to tweak it.


martinmaillard (approved):
Nice! 🚀

PR #82, opened 2020-10-20
feat: added cc:csvw to cc:Table

tpluscode:
This adds a handler for the future `/table/foo/csvw` resource which will generate the CSVW mapping document for tables.

The resource structure will be like:

```
graph </table/foo> {
  </table/foo> a cc:Table ; cc:csvw </table/foo/csvw>
}
```

The handler of `</table/foo/csvw>` I propose here is totally generated. Right now it copies the source URL and dialect. In the finally version we'll have to add all generated the columns according to the mapping users prepare.

---

This PR also introduces `rdfine` to the API code. It's already used indirectly in the UI app. By extracting a `@cube-creator/model` package we can share strongly-typed object abstraction over RDF between the front- and back-end


apis/core/hydra/table.ttl
```diff
@@ -0,0 +1,39 @@
...
+        "GET" ;
+      code:implementedBy
+        [
+          a
+            code:EcmaScript ;
+          code:link
+            <file:handlers/table/csvw#get> ;
+        ]
+    ]
+.
```
tpluscode, line 39: Something different: operation attached to a property. This way any `?table` matching the pattern as below will support a `GET` request

```sparql
graph ?resource {
  ?resource a cc:Table .
  ?resource cc:csvw ?table .
}
```

apis/core/lib/domain/table/csvw.ts
```diff
@@ -0,0 +1,39 @@
...
+    url: source.associatedMedia.contentUrl?.value,
+    dialect: {
+      types: [csvw.Dialect],
+      quoteChar,
+      delimiter,
+      encoding,
+      doubleQuote,
+    },
+  }) as any
+}
```
tpluscode, line 39: By annotating JS classes it is possible to avoid clownface pointers and directly using RDF vocabularies.

This also provides full type checking

packages/model/CsvSource.ts
```diff
@@ -0,0 +1,32 @@
...
+import { csvw, schema } from '@tpluscode/rdf-ns-builders'
+import { cc } from '@cube-creator/core/namespace'
+import { MediaObjectMixin } from '@rdfine/schema'
+import { DialectMixin } from '@rdfine/csvw'
+
+export interface CsvSource extends RdfResourceCore {
+  associatedMedia: Schema.MediaObject
+  name: string
+  dialect: Csvw.Dialect
+}
```
tpluscode, line 14: A strongly typed RDF resource gets an interface...

packages/model/CsvSource.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    @property.literal({ path: schema.name })
+    name!: string
+
+    @property.resource({ path: csvw.dialect, as: [DialectMixin] })
+    dialect!: Csvw.Dialect
+  }
+
+  return Impl
+}
```
tpluscode, line 29: And an annotated implementation. Of course plain properties and methods are also supported.

Note that this is a partial mixin class, which gets combined with potentially other at runtime.

martinmaillard (approved):

PR #86, opened 2020-10-21
feat: add table collection

martinmaillard:


ui/src/store/modules/cube-projects.ts
```diff
@@ -2,23 +2,32 @@ import { ActionTree, MutationTree, GetterTree } from 'vuex'
...
 import * as ns from '@cube-creator/core/namespace'
-import { Project, ProjectsCollection, CSVMapping, SourcesCollection } from '@/types'
+import { Project, ProjectsCollection, CSVMapping, SourcesCollection, TableCollection } from '@/types'
 
 export interface CubeProjectsState {
   collection: null | ProjectsCollection,
   project: null | Project,
   csvMapping: null | CSVMapping,
   sourcesCollection: null | SourcesCollection,
+  tableCollection: null | TableCollection,
```
tpluscode, line 12: Ditto, maybe we could drop the `Collection` suffix in property names

ui/src/api/mixins/TableCollection.ts
```diff
@@ -0,0 +1,19 @@
...
+  return class extends base implements TableCollection {
+    get actions () {
+      return {
+        create: findOperation(this, schema.CreateAction),
+      }
+    }
+  }
+}
+
+Mixin.appliesTo = ns.cc.TableCollection
```
tpluscode, line 19: Not to change this now but wonder if you really need this as a mixin. Food for thought if we indeed use `schema` for common actions

ui/src/api/mixins/Table.ts
```diff
@@ -0,0 +1,27 @@
+import { Resource } from 'alcaeus'
+import * as ns from '@cube-creator/core/namespace'
+import { schema } from '@tpluscode/rdf-ns-builders'
+import { Table, Source } from '@/types'
+import { findOperation } from '../common'
+
+export type Constructor = new (...args: any[]) => Resource;
+
+export default function Mixin<Base extends Constructor> (base: Base) {
```
tpluscode, line 9: I see what you did there 🙂. The patterns goes like this

```suggestion
import { Constructor } from '@tpluscode/rdfine'

export default function Mixin<Base extends Constructor<Resource> (base: Base) {
```

tpluscode (changes requested):

ui/src/types.ts
```diff
@@ -12,9 +12,16 @@ export interface SourcesCollection extends Collection {
...
+  actions: {
+    create: RuntimeOperation | null,
+  };
+}
+
 export interface CSVMapping extends Resource {
   actions: Record<string, unknown>;
-  sourcesCollection: SourcesCollection,
+  sourcesCollection: SourcesCollection;
+  tableCollection: TableCollection;
```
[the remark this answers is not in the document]
martinmaillard: I prefer to keep `tables` for an actual list of tables in the frontend. It's easier to understand.

ui/src/api/mixins/TableCollection.ts
martinmaillard: Yeah I'm not doing anything else that extracting the operations in `actions` at the moment so if we use schema for all of the operations, I can have a generic mixin for collections.

ui/src/api/mixins/Table.ts
martinmaillard: Alright, will change 👍 
I guess it was a pattern I ported from the first project.

apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -50,7 +50,25 @@ export async function createProject({
...
+      .addOut(hydra.manages, manages => {
+        // ?member rdf:type cc:Table
+        manages.addOut([hydra.property], [rdf.type])
+        manages.addOut([hydra.object], [cc.Table])
+      })
+      .addOut(hydra.manages, manages => {
+        // ?member cc:csvMapping <mapping>
+        manages.addOut([hydra.object], mapping)
+        manages.addOut([hydra.property], [cc.csvSource])
+      })
```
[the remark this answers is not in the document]
martinmaillard: Actually I think the comment was correct, but the `cc:csvSource` property should have been `cc:csvMapping`

ui/src/types.ts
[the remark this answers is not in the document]
tpluscode: Fair enough

tpluscode (approved):

PR #87, opened 2020-10-21
Create table

martinmaillard:
Missing things:
- [x] Table creation handler (#92)
- [x] A way to select the source (we don't have a component to show a select box yet)


martinmaillard:
Will address the column mapping feature in a separate PR (#110)

apis/core/lib/domain/table/create.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const table = await store
+    .createMember(tableCollection.term, id.table(csvMapping.term, label.term.value))
+
+  table.addOut(rdf.type, cc.Table)
+  table.addOut(cc.csvSource, resource.out(cc.csvSource))
+  table.addOut(cc.csvMapping, csvMapping.term)
+  table.addOut(schema.name, label)
+  table.addOut(cc.identifierTemplate, resource.out(cc.identifierTemplate))
+  table.addOut(schema.color, resource.out(schema.color))
```
tpluscode, line 38: We'll also need to add a `${table} cc:csvw <${table.id}/csvw>` triple here for the generated CSVW

tpluscode (approved):

PR #90, opened 2020-10-22
feat: redirect error page for 401/403

### LINKED ISSUE #74: UI: redirect to an error page when getting 401/3 from the API
The API now returns a meaningful error message with a link to an eIAM page to ask for access to the app. The UI should have a page to show this.

lucafurrer:
When an error 401/403 is raised the user gets routed to a page with a link (provided by the backend) where he can request access to the site.

resolves #74 

Note: New version of Alceaus was needed


ui/src/App.vue
```diff
@@ -44,6 +45,12 @@ export default Vue.extend({
       this.$store.dispatch('app/dismissMessage', message)
     },
   },
+  errorCaptured (err) {
```
martinmaillard, line 48: Interesting! I would have never thought of doing it that way. I like it :)

ui/src/views/NotAuthorized.vue
```diff
@@ -0,0 +1,22 @@
+<template>
+  <page-content>
+    <h2 class="title">
+      Missing Authorization
+    </h2>
+    <p>
+      You have not the right to access this site. Please request access it via
+      <a :href="link"> {{ link }}</a>
+    </p>
```
martinmaillard, line 9: We could use the `details` provided by the backend in the error response. But I'm also fine with this.

ui/src/App.vue
```diff
@@ -44,6 +45,12 @@ export default Vue.extend({
...
+    if (err instanceof APIErrorAuthorization) {
+      this.$router.push({ name: 'NotAuthorized', params: { link: err.details.link.href } })
+    }
+    return false
```
martinmaillard, line 52: Won't that stop any error from propagating?

martinmaillard (changes requested):

package.json
```diff
@@ -37,5 +37,8 @@
     "require": [
       "mocha-setup.js"
     ]
+  },
+  "dependencies": {
+    "alcaeus": "1.0.0-RC.3"
```
tpluscode, line 42: This must be in `ui/package.json`

tpluscode (changes requested):

ui/src/App.vue
lucafurrer: True, I missread the documentation.

ui/src/views/NotAuthorized.vue
lucafurrer: I was hesitating about this. The link is read from the error response

package.json
lucafurrer: Uups

ui/src/App.vue
lucafurrer: Yes, I misread the documentation, I will fix that.

package.json
```diff
@@ -37,5 +37,6 @@
     "require": [
       "mocha-setup.js"
     ]
-  }
+  },
+  "dependencies": {}
```
tpluscode, line 41:
```suggestion
  }
```

vhf (approved):

tpluscode (approved):
We might wanna work a little on the styling but otherwise it's fine once alcaeus is updated to `1.0.0-RC.4`

martinmaillard (approved):

PR #91, opened 2020-10-22
feat: use vue components are shaperone property/object renderers

martinmaillard:
This will give us more control over the styling of our forms.

One downside I still couldn't figure out is that there are a lot of superfluous HTML tags in the DOM, which makes styling way harder than it should be. But at least we can have a consistent look and change the rendering as needed.

It's also vue-in-webcomponents-in-vue, which makes the rendering even more janky than it was. But we can maybe address this later?


ui/src/forms/index.ts
```diff
@@ -1,15 +1,43 @@
 import { renderer, components, editors } from '@hydrofoil/shaperone-wc/configure'
+import { html } from 'lit-element'
 import { DefaultStrategy } from '@hydrofoil/shaperone-wc/renderer/DefaultStrategy'
-import * as MaterialRenderStrategy from '@hydrofoil/shaperone-wc-material/renderer'
```
tpluscode, line 3: Guess this would no longer be needed as dependency

martinmaillard: Good catch

tpluscode (approved):

Issue #88, opened 2020-10-22
Create Delete API Handlers


lucafurrer:
The following Delete Handlers should be treated by the API

- [x] API handler to delete a project
- [x] API handler to delete a source
- [ ] API handler to delete a table

Open questions: 
- Can a source be deleted if it has associated tables?
- Do we want to hard delete or soft delete? (22.10)
  - hard delete: delete all triples
  - ~~soft delete: flag as deleted~~
- How do we handle cascade delete of associated resources?

lucafurrer:
Discussion on Zulip on Thursday 22.10.2020: 

**We go for hard delete**

lucafurrer:
Open question: 
How to exactly implement the cascade delete?
For the moment @tpluscode and I would go for an approach which every delete code handles the cascade itself. Up to see if it is possible to call the delete on the linked resources itself.

lucafurrer:
@martinmaillard  what do you think about the delete if it has associated tables?
- Not possible
- With warning?

tpluscode:
> How to exactly implement the cascade delete?

Just a not that this side effect on other resources does not have to be delete.

For example, we might allow deleting a source and mark the tables so that they need user's attention. Smth like

```diff
<table>
  cc:source <source> #might keep this, no instant harm in broken links
  schema:error "Source gone. Please update the table"
```

martinmaillard:
Closing this one. I created #121 for table deletion

PR #93, opened 2020-10-23
feat: delete csv-source

lucafurrer:
Delete a CSV-Source
Does:
- Remove file from S3
- Delete graph of csv.source

Does not (yet): 
- Delete associated tables or prevent deletion when tables are present 

Part of #88


apis/core/lib/ResourceStore.ts
```diff
@@ -121,4 +140,11 @@ export default class implements ResourceStore {
 
     return member
   }
+
+  delete(id: NamedNode): void {
+    if (!this.__resources.has(id)) {
+      throw new Error('Resource does not exist')
+    }
+    this.__deletedGraphs.add(id)
```
lucafurrer, line 148: Should we remove all nodes of this graph from the store at this moment?

apis/core/lib/domain/csv-source/delete.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export async function deleteSource({
+  resource,
+  store = resourceStore(),
+}: DeleteSourceCommand): Promise<void> {
+  const csvSource = await store.get(resource)
+
+  // TODO: Delete Tables or prevent it
+  // Find related tables
+  // Delete them
```
lucafurrer, line 20: I suggest to wait until the table implementation is done, to continue on this

apis/core/test/ResourceStore.test.ts
```diff
@@ -165,4 +165,33 @@ describe('ResourceStore', () => {
...
+      const resourceStore = new ResourceStore(client)
+      resourceStore.create(ex.Foo)
+      await resourceStore.save()
+      resourceStore.delete(ex.Foo)
+
+      // when
+      await resourceStore.save()
+
+      // then
+      expect(client.query.update).to.have.been.calledOnce
```
lucafurrer, line 194: @tpluscode what would be the good way to check that the graph does not exist anymore?

apis/core/hydra/csv-source.ttl
```diff
@@ -57,6 +57,6 @@ cc:CSVSource
           a
             code:EcmaScript ;
           code:link
-            <file:handlers/csv-source#delete> ;
+            <file:handlers/csv-source#remove> ;
```
lucafurrer, line 60: Somehow the compiler was not too happy with a function called delete in the handler...

e2e-tests/csv-source/delete.hydra
```diff
@@ -0,0 +1,73 @@
...
+    Expect Operation cc:DeleteCSVSourceAction {
+        Invoke {
+        } => {
+            Expect Status 204
+        }     
+    }
+    Expect Operation cc:GetCSVSourceAction {
+        Invoke {
+        } => {
+            Expect Status 404
```
lucafurrer, line 70: I would expect a 404 but a 405 is returned. Not sure if we should adapt the test or if there is a bug somewhere

tpluscode (changes requested):

apis/core/test/support/TestResourceStore.ts
```diff
@@ -3,13 +3,15 @@ import $rdf from 'rdf-ext'
...
+import TermSet from '@rdfjs/term-set'
 
 export class TestResourceStore implements ResourceStore {
   private readonly resources: TermMap<NamedNode, GraphPointer<NamedNode>>
+  private readonly __deletedGraphs: TermSet
 
   constructor(pointers: GraphPointer<NamedNode>[]) {
     this.resources = new TermMap()
-
+    this.__deletedGraphs = new TermSet()
```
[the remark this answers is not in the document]
lucafurrer: It is private...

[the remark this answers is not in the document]
tpluscode: The underscores is a convention which we don't enforce, but maybe we should. Ok, forget my suggestion for now

e2e-tests/csv-source/delete.hydra
lucafurrer: From a zulip discussion with @tpluscode 
`hydra-box looks up resources+operations by checking a resource named X (that is obviously getting removed) but then it also looks up resources used as objects by other resources. in our case it is a leftover <csv-mapping> cc:csvSource <deleted-source>`

apis/core/lib/domain/csv-source/delete.ts
lucafurrer: Done -> #97

martinmaillard (approved):

tpluscode (approved):

tpluscode:
Will also need a rebase or merge master

lucafurrer:
I did it 1 hour ago, but here we go again

PR #94, opened 2020-10-23
feat: delete project

lucafurrer:
Make it possible to delete a projet

Deletes the project and the linked csvMapping as well as all linked csvSources

Linked to #88


lucafurrer:
Depends on #93

apis/core/hydra/projects.ttl
```diff
@@ -9,7 +9,24 @@ cc:CubeProject
   a
     hydra:Class ;
   query:include
-    cc:csvMapping .
+    cc:csvMapping ;
+  hydra:supportedOperation
+  [
+    a 
+      hydra:Operation, 
+      cc:DeleteProjectAction ;
```
martinmaillard, line 17: I would replace `cc:DeleteProjectAction` with `schema:DeleteAction`

lucafurrer: Should I also remove cc:DeleteCSVSourceAction to be consistent?

martinmaillard: I think I handled it in https://github.com/zazuko/cube-creator/pull/108

apis/core/lib/domain/csv-source/delete.ts
```diff
@@ -15,12 +15,18 @@ export async function deleteSource({
   resource,
   store = resourceStore(),
 }: DeleteSourceCommand): Promise<void> {
+  await deleteSourceWithoutSave(resource, store)
+
+  // Save changes
+  await store.save()
+}
+
+export async function deleteSourceWithoutSave(resource: NamedNode<string>, store: ResourceStore): Promise<void> {
```
lucafurrer, line 24: I think it is preferable that the store.save is called only once when deleting a project.

apis/core/lib/handlers/cube-projects.ts
```diff
@@ -23,3 +24,12 @@ export const post = protectedResource(
     await res.dataset(project.dataset)
   }),
 )
+
+export const remove = protectedResource(
+  shaclValidate,
```
tpluscode, line 29: No body, no validation :)
```suggestion

```

apis/core/lib/domain/queries/csv-source.ts
```diff
@@ -17,3 +17,29 @@ export async function sourceWithFilenameExists(csvMapping: NamedNode, fileName:
...
+        if (value) {
+          sources.push(value as string)
+        }
+      })
+    })
+
+    stream.on('end', () => resolve(sources))
+    stream.on('error', error => reject(error))
+  })
+}
```
tpluscode, line 45: This can be much simpler with a `parsingClient` (also exported from `../../query-client`)

```suggestion
export async function * getSourcesFromMapping(csvMapping: NamedNode, client = parsingClient) {
  const results = await SELECT.DISTINCT`?source`
    .WHERE`
    GRAPH ${csvMapping}
    {
      ${csvMapping} ${cc.csvSource} ?source
    }
    `
    .execute(client.query)

  for (const result of results) {
    const source = result.source
    if (source.termType === 'NamedNode') {
      yield source
    }
  }
}
```

apis/core/lib/domain/csv-mapping/delete.ts
```diff
@@ -0,0 +1,27 @@
...
+import { getSourcesFromMapping } from '../queries/csv-source'
+
+import { deleteSourceWithoutSave } from '../csv-source/delete'
+import { cc } from '@cube-creator/core/namespace'
+
+export async function deleteMapping(csvMapping: NamedNode, store: ResourceStore): Promise<void> {
+  const sources = await getSourcesFromMapping(csvMapping)
+  for (const source of sources) {
+    await deleteSourceWithoutSave(source, store)
+  }
```
tpluscode, line 13: But besides, why not using the store?

```suggestion
  const csvMappingResource = await store.get(csvMapping)
  const sources = csvMappingResource.out(cc.csvSource).terms
  for await (const source of sources) {
    if (source.termType === 'NamedNode') {
      await deleteSource({resource: source, store})
    }
  }
```

tpluscode (changes requested):

apis/core/lib/domain/queries/csv-source.ts
lucafurrer: Yes this is nice, but I go with the other suggestion :wink:

apis/core/lib/domain/csv-mapping/delete.ts
lucafurrer: Actually I did not need to load the graph at the beginning. But it is way cleaner without a query, in my opinion. I go with this one

Issue #92, opened 2020-10-22
API endpoint to create a table


martinmaillard:
I started the form to create a table here https://github.com/zazuko/cube-creator/pull/87 but it's missing the API endpoint.

Issue #10, opened 2020-09-15
Data model


martinmaillard:
High-level overview of the application flow:
![cc_high-level-data-flow](https://user-images.githubusercontent.com/1334124/93220825-4b987500-f76d-11ea-9cb0-dd318e542e94.png)

Raw data model (⚠️ outdated):
![cc_raw-data-model](https://user-images.githubusercontent.com/1334124/93220869-5bb05480-f76d-11ea-93ed-60a3582f326c.png)

martinmaillard:
This is an example of the data we generate with the tool. Almost nothing is set in stone, feel free to comment/edit.

I started with [Adrian's UBD cube example](https://github.com/zazuko/rdf-cube-schema-viz/tree/master/example) and tried to invent how we would get there if we had mapped the cube from 2 CSV files in the tool:
- ubd.csv: contains the measures
- stations.csv: contains station names

## Project

```turtle
<project-1> a cc:CubeProject ;
    cc:cube <foen/ubd/28/pm1> ;
    cc:csvMapping <project-1/mapping> ;
.

<project-1/mapping> a cc:CSVMapping ;
.
```

## Sources

```turtle
<source-1> a cc:CSVSource ;
    cc:csvMapping <project-1/mapping> ;
    schema:name "ubd.csv" ;
    csvw:column <source-1/column-year> ;
    csvw:column <source-1/column-station> ;
    csvw:column <source-1/column-value> ;
    csvw:column <source-1/column-aggregation> ;
.

<source-1/column-year>
    schema:name "YEAR" ;
.

<source-1/column-station>
    schema:name "STATION" ;
.

<source-1/column-value>
    schema:name "VALUE" ;
.

<source-1/column-aggregation>
    schema:name "AGGREGATION" ;
.

<source-2> a cc:CSVSource ;
    cc:csvMapping <project-1/mapping> ;
    schema:name "stations.csv" ;
    csvw:column <source-1/column-year> ;
    csvw:column <source-2/column-station-name-fr> ;
    csvw:column <source-2/column-station-name-de> ;
.

<source-2/column-station-id>
    schema:name "STATION_ID" ;
.

<source-2/column-station-name-fr>
    schema:name "STATION_NAME_FR" ;
.

<source-2/column-station-name-de>
    schema:name "STATION_NAME_DE" ;
.
```

## Tables

```turtle
<table-observation> a cc:Table ;
    cc:csvMapping <project-1/mapping> ;
    cc:source <source-1> ;
[...]

PR #96, opened 2020-10-26
Endpoint to create a table

vhf:
#92

### DIFF
--- a/apis/core/lib/domain/identifiers.ts
+++ b/apis/core/lib/domain/identifiers.ts
@@ -1,6 +1,6 @@
 import UrlSlugify from 'url-slugify'
 import $rdf from 'rdf-ext'
-import { NamedNode } from 'rdf-js'
+import { NamedNode, Term } from 'rdf-js'
 import { GraphPointer } from 'clownface'
 import env from '@cube-creator/core/env'
 
@@ -20,6 +20,10 @@ export function csvSourceCollection(csvMapping: GraphPointer<NamedNode>): NamedN
 
 export function tableCollection(csvMapping: GraphPointer<NamedNode>): NamedNode {
   return $rdf.namedNode(`${csvMapping.value}/tables`)
+}
+
+export function table(csvMapping: Term, label: string): NamedNode {
+  return $rdf.namedNode(`${csvMapping.value}/table/${url.slugify(label)}`)
 }
 
 export function csvSource(project: GraphPointer<NamedNode>, fileName: string): NamedNode {
--- a/apis/core/lib/domain/table/create.ts
+++ b/apis/core/lib/domain/table/create.ts
@@ -1,0 +1,42 @@
+import { GraphPointer } from 'clownface'
+import { schema, rdf } from '@tpluscode/rdf-ns-builders'
+import { cc } from '@cube-creator/core/namespace'
+import { ResourceStore } from '../../ResourceStore'
+import * as id from '../identifiers'
+import { resourceStore } from '../resources'
+import { NamedNode } from 'rdf-js'
+
+interface CreateTableCommand {
+  tableCollection: GraphPointer<NamedNode>
+  resource: GraphPointer
+  store?: ResourceStore
+}
+
+export async function createTable({
+  tableCollection,
+  resource,
+  store = resourceStore(),
+}: CreateTableCommand): Promise<GraphPointer> {
+  const label = resource.out(schema.name)
+  if (!label?.term) {
+    throw new Error('schema:name missing from the payload')
+  }
+
+  const csvMapping = tableCollection.out(cc.csvMapping)
+  if (!csvMapping?.term) {
+    throw new Error('cc:csvMapping missing from the payload')
+  }
+
+  const table = await store
+    .createMember(tableCollection.term, id.table(csvMapping.term, label.term.value))
+
+  table.addOut(rdf.type, cc.Table)
+  table.addOut(cc.csvSource, resource.out(cc.csvSource))
+  table.addOut(cc.csvMapping, csvMapping.term)
+  table.addOut(schema.name, label)
+  table.addOut(cc.identifierTemplate, resource.out(cc.identifierTemplate))
+  table.addOut(schema.color, resource.out(schema.color))
+
+  await store.save()
+  return table
+}
--- a/apis/core/hydra/table.ttl
+++ b/apis/core/hydra/table.ttl
@@ -26,7 +26,7 @@ cc:TableCollection
           a
             code:EcmaScript ;
           code:link
-            <file:handlers/table#post> ;
+            <file:handlers/table/create#post> ;
         ] ;
     ] .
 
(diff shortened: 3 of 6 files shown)


apis/core/lib/domain/table/create.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const table = await store
+    .createMember(tableCollection, id.table(label.value))
+
+  table.addOut(schema.name, label)
+  // TableShape
+  if (shape('table/create').equals(resource.out(cc.Table).term)) {
+    table
+      .addOut(rdf.type, cc.Table)
+  }
```
vhf, line 30: @tpluscode what's the best way to get the triples that were posted to this endpoint and add all of them to the `table` dataset?

apis/core/lib/domain/identifiers.ts
```diff
@@ -22,6 +22,10 @@ export function tableCollection(csvMapping: GraphPointer<NamedNode>): NamedNode
   return $rdf.namedNode(`${csvMapping.value}/tables`)
 }
 
+export function table(label: string): NamedNode {
+  return $rdf.namedNode(`table-${url.slugify(label)}`)
+}
```
tpluscode: I think we'd want to have tables namespaced by the containing project

apis/core/lib/domain/identifiers.ts
```diff
@@ -22,6 +22,10 @@ export function tableCollection(csvMapping: GraphPointer<NamedNode>): NamedNode
   return $rdf.namedNode(`${csvMapping.value}/tables`)
 }
 
+export function table(label: string): NamedNode {
+  return $rdf.namedNode(`table-${url.slugify(label)}`)
```
martinmaillard, line 26: It should probably use the `csvMapping` to build the URI, similar to CSV source

apis/core/lib/domain/table/create.ts
martinmaillard: I tried to build the endpoint and had the exact same question.

martinmaillard (changes requested):

apis/core/lib/domain/table/create.ts
tpluscode: The request payload is parsed from `await req.dataset()`

To add to the dataset you'd likely loop over that to ensure correct graph URI in the table's dataset

```javascript
[...payloadData].forEach(quad => {
  table.addOut(quad.property, quad.subject)
})
```

That would be the simplest way which assumes that the payload is only triples where the tables is the subject of what you create. Obviously a deeper or more complex graph would require a more specialised approach.

apis/core/lib/domain/table/create.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (!label?.term) {
+    throw new Error('schema:name missing from the payload')
+  }
+
+  const csvMapping = tableCollection.out(cc.csvMapping)
+  if (!csvMapping?.term) {
+    throw new Error('cc:csvMapping missing from the payload')
+  }
+
+  if (!tableCollection?.term) {
```
martinmaillard, line 30: It looks like I'm seeing an Open World™ 😄

martinmaillard (approved):

apis/core/lib/domain/table/create.ts
vhf: Yeah but this:
Screenshot 2020-10-27 at 16 10 35
 is so much more useful than this:
Screenshot 2020-10-27 at 16 10 56

😄

martinmaillard: I wasn't complaining about your code. Half my code ends up looking the same: with `if` statements every 2 lines... I was just saying that it's a direct consequence of the "Open World" assumption.

tpluscode, line 30: This is unnecessary because `tableCollection` will not be falsy

apis/core/lib/domain/table/create.ts
```diff
@@ -0,0 +1,45 @@
...
+}: CreateTableCommand): Promise<GraphPointer> {
+  const label = resource.out(schema.name)
+  if (!label?.term) {
+    throw new Error('schema:name missing from the payload')
+  }
+
+  const csvMapping = tableCollection.out(cc.csvMapping)
+  if (!csvMapping?.term) {
+    throw new Error('cc:csvMapping missing from the payload')
+  }
```
tpluscode, line 28: The `cc:csvMapping` is not part of the `table/create` shape.

Get it from the `tableCollection` instance

apis/core/lib/domain/table/create.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (!tableCollection?.term) {
+    throw new Error('Resource is not a valid tableCollection')
+  }
+
+  const table = await store
+    .createMember(tableCollection.term as NamedNode, id.table(csvMapping.term, label.term.value))
+
+  table.addOut(rdf.type, cc.Table)
+  table.addOut(cc.csvSource, resource.out(cc.csvSource))
```
tpluscode, line 38: Please add the `cc:csvMapping` link to the table as we shown in https://github.com/zazuko/cube-creator/issues/10

tpluscode (changes requested):
Please also add an case in e2e-tests to check that `cc:csvMapping` is actually added to the created resource

apis/core/lib/domain/table/create.ts
vhf: The compiler complains that it can be undefined

apis/core/lib/domain/table/create.ts
vhf: Isn't it what this code already does? What do you suggest changing?

tpluscode: Ah shoot, I only looked at the error message where I understood as `payload === request body`

apis/core/lib/domain/table/create.ts
tpluscode: You sure? Do check again

apis/core/lib/domain/table/create.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (!csvMapping?.term) {
+    throw new Error('cc:csvMapping missing from the payload')
+  }
+
+  if (!tableCollection?.term) {
+    throw new Error('Resource is not a valid tableCollection')
+  }
+
+  const table = await store
+    .createMember(tableCollection.term as NamedNode, id.table(csvMapping.term, label.term.value))
```
tpluscode, line 35:
```suggestion
    .createMember(tableCollection.term, id.table(csvMapping.term, label.term.value))
```

This will also be unnecessary when you change the interface typing

apis/core/lib/domain/table/create.ts
```diff
@@ -0,0 +1,45 @@
+import { GraphPointer } from 'clownface'
+import { schema, rdf } from '@tpluscode/rdf-ns-builders'
+import { cc } from '@cube-creator/core/namespace'
+import { ResourceStore } from '../../ResourceStore'
+import * as id from '../identifiers'
+import { resourceStore } from '../resources'
+import { NamedNode } from 'rdf-js'
+
+interface CreateTableCommand {
+  tableCollection: GraphPointer
```
tpluscode, line 10:
```suggestion
  tableCollection: GraphPointer<NamedNode>
```

tpluscode (changes requested):

martinmaillard (approved):

PR #101, opened 2020-10-26
feat: use collection resource to populate dash:InstancesSelectEditor

tpluscode:
I implemented the csv source select by directly linking the required `hydra:Collection` on the property.

A little out-of-band, in a better solution this would have been included on a specific instance of shape, directly linked to the table resource.

### DIFF
--- a/packages/core/namespace.ts
+++ b/packages/core/namespace.ts
@@ -32,6 +32,7 @@ type OtherTerms =
 
 type CubeCreatorNamespace = Record<CubeCreatorClass | CubeCreatorProperty | OtherTerms, NamedNode>
 
+export const hashi = namespace('http://hypermedia.app/shapes#')
 export const cube = namespace('http://ns.bergnet.org/cube/')
 export const cc: CubeCreatorNamespace = namespace('https://cube-creator.zazuko.com/vocab#') as any
 export const editor = namespace(cc.dash.value)
--- a/ui/src/forms/BulmaSelect.vue
+++ b/ui/src/forms/BulmaSelect.vue
@@ -1,5 +1,5 @@
 <template>
-  <b-select placeholder="Select" @input="emit">
+  <b-select placeholder="Select" @input="emit" :value="value">
     <option
       v-for="option in options"
       :value="option.id"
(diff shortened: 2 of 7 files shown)


packages/core/namespace.ts
```diff
@@ -32,6 +32,7 @@ type OtherTerms =
 
 type CubeCreatorNamespace = Record<CubeCreatorClass | CubeCreatorProperty | OtherTerms, NamedNode>
 
+export const hashi = namespace('http://hypermedia.app/shapes#')
```
tpluscode, line 35: This is a namespace I intend for a Hydra-SHACL bridge

ui/src/forms/BulmaSelect.vue
```diff
@@ -0,0 +1,44 @@
...
+    if (this.collection && this.collection.id.termType === 'NamedNode') {
+      const { representation } = await Hydra.loadResource(this.collection.id)
+      this.options = representation?.root?.member || []
+    }
+  }
+
+  emit (value: NamedNode): void {
+    this.update(value)
+  }
+}
```
tpluscode, line 43: An instance select component is aware of Hydra and loads its elements from a hydra Collection resource.

`schema:name` right now hardcoded for the display property but we could have a bunch alternatives or property annotation if we wanted to.

ui/src/views/TableCreate.vue
```diff
@@ -60,13 +60,17 @@ export default Vue.extend({
       const { representation } = await expects.load()
       const { root: shape } = representation
       if (shape) {
+        const sourceProperty = shape.property.find(p => p.class?.equals(ns.cc.CSVSource))
+        sourceProperty[ns.hashi.collection.value] = this.sourcesCollection
+
```
tpluscode, line 65: This part is where the UI now cuts corner a little because there is currently no good way to find the collection URL only by the `sh:class` annotation

ui/src/forms/BulmaSelect.vue
```diff
@@ -0,0 +1,44 @@
...
+import { Hydra } from 'alcaeus/web'
+import { NamedNode } from 'rdf-js'
+import { schema } from '@tpluscode/rdf-ns-builders'
+
+const labelProperty = schema.name
+
+@Component
+export default class extends Vue {
+  @Prop() collection?: RdfResource
+  @Prop() update!: (newValue: string | NamedNode) => void
```
martinmaillard, line 25: It's missing the initial `value`

martinmaillard (changes requested):

martinmaillard (approved):

PR #106, opened 2020-10-27
fix: load linked resources before SHACL validation

martinmaillard:
To be able to validate the type of linked resources with SHACL, we need to load the types of linked resources before validation.


apis/core/lib/domain/queries/resources-types.ts
```diff
@@ -0,0 +1,32 @@
...
+    .WHERE`
+      values ?resource {
+        ${ids}
+      }
+    `
+    .WHERE`GRAPH ?g {
+        ?resource a ?type .
+    }`.execute(client.query)
+
+  return new Promise((resolve, reject) => {
```
martinmaillard, line 17: I know there's an easier way to do this, but I don't remember how...

martinmaillard:
Looks like the SPARQL endpoint is not available during unit tests (makes sense). Not sure what to do about that.

apis/core/lib/domain/queries/resources-types.ts
martinmaillard: Found it

PR #107, opened 2020-10-27
feat: produce cube using new schema from the pipeline

### LINKED ISSUE #100: Adapt the pipeline to produce new cube schema
Right now the pipeline is simply a simplified copy or the old CLI from data-cube-curation

We need to adopt step from [FOEN UBD 28](https://gitlab.ldbar.ch/pipelines/foen-ubd-28/-/blob/master/pipelines/main.ttl) to produce the dataset using the new cube schema

**DoD**

1. The e2e test needs to be updated to check against correct expected data
2. The test must pass

tpluscode:
Fixes #100 by importing barnard59 cube schema steps

Notable changes (or semi-changes):

1. The pipeline assumes a virtual column on table which has produce `?observation cc:cube ?cube` triples. 
   - They are filtered out in the end. 
   - The `?cube` object of those triples is used to construct the `observationSet` IDs
2. This will have the impact that a cube project will have to allow multiple cubes
   - Cube as template means that we will not know the cube URIs up front; they will have to extracted from the result graph
   - Instead I added `cc:cubeGraph` property for the project to direct the result into that named graph


cli/test/support/testData.ts
```diff
@@ -35,7 +35,7 @@ export const insertTestData = async () => {
       </project/cli-test>
           a ${hydra.Resource} , ${cc.CubeProject} ;
           ${cc.csvMapping} </project/cli-test/mapping> ;
-          ${cc.cube}       </cube/cli-test>
+          ${cc.cubeGraph}  </cube/cli-test>
```
tpluscode, line 38: I propose a `cc:cubeGraph` property on project to explicitly prepare the location for generated triples

cli/pipelines/main.ttl
```diff
@@ -77,3 +86,50 @@
...
+                      code:name "observations";
+                                code:value [
+                                             a code:EcmaScript ;
+                                             code:link <file:../lib/cube#getObservationSetId>
+                                         ]
+
+                    ], [
+                         code:name "observation";
+                         code:value "({ dataset, observations }) => ([...dataset][0].subject)"^^code:EcmaScript
+                       ] .
```
tpluscode, line 112: This is the key part of the new pipeline steps.

The `observation` parameter simply uses the generated `aboutUrl` from csvw as the observation URI

The `observations` parameter is code to generate the `cube:observationSet` URI. I propose that the CSVW would implicitly add a mapping to output a `cc:cube` property. The `cube:observationSet` would simply be `${cubeId}/observation/`

martinmaillard (approved):

PR #109, opened 2020-10-28
Fix weird form behavior

martinmaillard:
By default, shaperone wraps a focus node fieldset in a `<form>` tag, which makes it impossible to control the behavior of the form. I just replaced the default renderer with a custom one without the form tag.


tpluscode:
Curious, what is the weird behavior? @martinmaillard

martinmaillard:
For example hitting "enter" in a field will submit the form. We want to be able to control that.

Issue #99, opened 2020-10-26
Integrate with GitLab for running transformation pipeline


tpluscode:
The app should be able to trigger a pipeline job to run a transformation (on GitLab?)

**DoD**

* [x] The UI has a "trigger" button to start a transformation job
* [x] The API has a hander for this "trigger" request
   - a `Job` resource is created in the data store
   - `Job` should link to the CSVW resources to transform, output graph URI, etc; 
   - `Job collection` resource created as child of `csv mapping`
* [x] GitLab is triggered (Job URI as parameter)
* [x] GitLab starts running the CLI (depends on #98) 
* [x] When done, the pipeline runner updates the `Job` resource 
   - success or not
   - something else?
* [x] All data get overwritten in the cube's observations graph
   - I propose to keep observation separated to avoid overwriting the cube itself

lucafurrer:
What information does the trigger button provide?

CSV-Mapping is given from the context. 
- Tables?
- Target Graph?

tpluscode:
The "trigger button" does not really provide any information.

The tables you'd get from the table collection `<csv-mapper> cc:tables <csv-mapper/tables>`
The target graph you'd get from the project `<project> cc:cubeGraph <whatever>`

tpluscode:
In fact, for the `Job` resource I propose a structure like

```
graph <csv-mapping/jobs/xyz> {
  <csv-mapping/jobs/xyz> a cc:Job ;
    cc:csvw <table-1/csvw> , <table-2/csvw> , <table-3/csvw> ;
    cc:cubeGraph <from-project>
}
```

`cc:csvw` would be a property of `Table`

`xyz` could be a short string like generated by [nanoid](https://npm.im/nanoid)

lucafurrer:
After discussion we would have

```
graph <csv-mapping/jobs/xyz> {
  <csv-mapping/jobs/xyz> a cc:Job ;
    cc:tables <csv-mapping/tables> ;
    cc:cubeGraph <from-project>
}
```
Correct?

tpluscode:
Agree @lucafurrer 👌

lucafurrer:
Just thinking aloud, I think a indication when a job was created could be useful for debugging etc.
`schema:dateCreated` or `dcterms:created`

lucafurrer:
@tpluscode you agree that this is done

tpluscode:
Definitely. Thanks!

PR #116, opened 2020-10-29
Cli trigger job

lucafurrer:
Create a job and get it executed by the cli

Links to #99


packages/model/Job.ts
```diff
@@ -0,0 +1,31 @@
...
+export function JobMixin<Base extends Constructor>(base: Base) {
+  class Impl extends base implements Job {
+    @property.resource({ path: cc.tables })
+    tableCollection!: Link<Collection<Table>>
+
+    @property({ path: cc.cubeGraph })
+    cubeGraph?: NamedNode
+
+    @property.literal({ path: schema.label })
+    label!: string
```
tpluscode, line 25:
```suggestion
    label!: string
    
    @property.literal({ path: dcterms.created, type: Date })
    create!: Date
```

apis/core/lib/pipeline/trigger.ts
```diff
@@ -0,0 +1,25 @@
+import { URLSearchParams } from 'url'
+import { NamedNode } from 'rdf-js'
+import fetch from 'node-fetch'
+import env from '@cube-creator/core/env'
+
+const pipelineURI = env.PIPELINE_URI
+
+export async function triggerPipeline(job: NamedNode): Promise<void> {
```
tpluscode, line 8: Let's call this a `triggerLocalPipeline` or similar to make it clear already that it's used in local env

cli/server.ts
```diff
@@ -0,0 +1,46 @@
...
+async function main() {
+  const log = debug('cube-creator')
+
+  const app = express()
+
+  app.enable('trust proxy')
+
+  app.use(cors())
+
+  app.use(bodyParser.urlencoded())
```
tpluscode, line 25: We'll need RDF support here. Please use [@rdfjs/express-handler](https://github.com/rdfjs-base/express-handler) for reading the request body as RDF

tpluscode (changes requested):

apis/core/lib/pipeline/trigger.ts
lucafurrer: In a perfect world we could do the switch via the url in env. Ok, I am not sure this will work

cli/server.ts
lucafurrer: Do we? We send one argument here and in the old pipeline there was only a querystring.

tpluscode: Sorry so maybe I misread. Are you sending the job URI as query string?

apis/core/lib/pipeline/trigger.ts
tpluscode: This will not be a simple URL substitution. The HTTP request to trigger GitHub or GitLab will be completely different.

cli/server.ts
lucafurrer: I expressed myself badly
```typescript
  const form = new URLSearchParams({
    JOB_URI: job.value,
  })

  const res = await fetch(pipelineURI, {
    method: 'POST',
    body: form,
  })
```

apis/core/lib/pipeline/trigger.ts
lucafurrer: What do you think of think of this approach

tpluscode: Yes, I was thinking similar. Can be even simpler by star import and using the env var as function name:

```javascript
import * as triggers from './lib/pipeline/trigger'

const trigger = triggers[env.PIPELINE_TYPE]

if (!trigger) {
  throw new Error(`Trigger ${env.PIPELINE_TYPE} is not implemented`)
}

trigger(job)
```

and in `lib/pipeline/trigger.ts`

```js
export function local() {
}

export function github() {
}

export function gitlab() {
}
```

cli/server.ts
tpluscode: OK, case closed

apis/core/lib/pipeline/trigger.ts
lucafurrer: Fancy 👍

cli/test/lib/commands/transform.test.ts
```diff
@@ -6,7 +6,7 @@ import { CONSTRUCT } from '@tpluscode/sparql-builder'
 import transform from '../../../lib/commands/transform'
 import { log } from '../../support/logger'
 import { setupEnv } from '../../support/env'
-import { client, insertTestData } from '@cube-creator/testing/lib'
+import { client, insertTestData } from '../../../../packages/testing/lib'
```
tpluscode, line 9: Please revert

cli/test/lib/job.test.ts
```diff
@@ -2,15 +2,15 @@ import { describe, before, beforeEach, it } from 'mocha'
 import { expect } from 'chai'
 import { Hydra } from 'alcaeus/node'
 import env from '@cube-creator/core/env'
-import { ProjectIterator } from '../../lib/project'
-import { insertTestData } from '@cube-creator/testing/lib'
+import { JobIterator } from '../../lib/job'
+import { insertTestData } from '../../../packages/testing/lib'
```
tpluscode, line 6: Please revert

tpluscode (changes requested):

PR #124, opened 2020-10-30
Fix: file upload

lucafurrer:
I discovered that we did not upload the entire file and @tpluscode observed some performance problems. this PR fixes both

### DIFF
--- a/apis/core/lib/domain/csv/file-head.ts
+++ b/apis/core/lib/domain/csv/file-head.ts
@@ -8,7 +8,7 @@ export function loadFileHeadString(
   let partialLine = ''
   stream.on('readable', () => {
     let chunk
-    while ((chunk = stream.read()) !== null) {
+    while ((chunk = stream.read(1000)) !== null) {
       let string = chunk.toString()
       if (!string.includes('\n')) {
         // continue reading until we reach the first \n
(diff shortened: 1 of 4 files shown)


apis/core/lib/domain/csv/file-head.ts
```diff
@@ -8,7 +8,7 @@ export function loadFileHeadString(
   let partialLine = ''
   stream.on('readable', () => {
     let chunk
-    while ((chunk = stream.read()) !== null) {
+    while ((chunk = stream.read(1000)) !== null) {
```
lucafurrer, line 11: Force the reader to only read smaller chunks of the file at once.
Yes this number is a little bit random.

apis/core/lib/handlers/csv-source.ts
```diff
@@ -27,7 +27,7 @@ export const post = labyrinth.protectedResource(
       : new Date().toISOString()
 
     const fileLocation = await uploadFile({
-      file: req.read(),
+      file: req,
```
lucafurrer, line 30: With the call of read be actually only got the first chunk of it.

martinmaillard (approved):

vhf (approved):

PR #127, opened 2020-11-02
typescript improvements

tpluscode:
I removed `skipLibCheck: true` and excluded UI from the root tsconfig. This removed a ton of irrelevant TS "issues" reported by Webstorm's TS support and also revealed small fixes which were required

The only thing was that `@types/jest` conflicted with `@types/mocha`. Thus, I removed `@vue/cli-plugin-unit-jest`. Are we going to use jest in UI project @martinmaillard ?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,15 +10,13 @@
     "noImplicitReturns": true,
     "strictNullChecks": true,
     "allowSyntheticDefaultImports": true,
-    "skipLibCheck": true,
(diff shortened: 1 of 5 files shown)


martinmaillard:
> Are we going to use jest in UI project @martinmaillard ?

If we ever have the time to write tests, probably ;) We'll figure it out at this point I guess

martinmaillard (approved):

PR #129, opened 2020-11-02
feat: delete table

### LINKED ISSUE #121: API: delete a Table

lucafurrer:
Adds an API point to delete a table

Tables will deleted, when a project is deleted

fix #121

### DIFF
--- a/apis/core/hydra/table.ttl
+++ b/apis/core/hydra/table.ttl
@@ -51,7 +51,7 @@ cc:Table
         hydra:Operation,
         schema:DeleteAction ;
       hydra:title
-        "Delete Project" ;
+        "Delete Table" ;
       hydra:method
         "DELETE" ;
       code:implementedBy
--- a/apis/core/hydra/table.ttl
+++ b/apis/core/hydra/table.ttl
@@ -45,6 +45,24 @@ cc:Table
       hydra:property
         cc:csvw ;
     ] ;
+  hydra:supportedOperation
+    [
+      a
+        hydra:Operation,
+        schema:DeleteAction ;
+      hydra:title
+        "Delete Project" ;
+      hydra:method
+        "DELETE" ;
+      code:implementedBy
+        [
+          a
+            code:EcmaScript ;
+          code:link
+            <file:handlers/table/delete#remove> ;
+        ] ;
+    ]
+  
 .
 
 cc:csvw
--- a/apis/core/lib/domain/csv-mapping/delete.ts
+++ b/apis/core/lib/domain/csv-mapping/delete.ts
@@ -1,4 +1,3 @@
-
 import { NamedNode } from 'rdf-js'
 import { ResourceStore } from '../../ResourceStore'
 import { deleteSourceWithoutSave } from '../csv-source/delete'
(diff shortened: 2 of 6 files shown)


apis/core/hydra/table.ttl
```diff
@@ -45,6 +45,24 @@ cc:Table
       hydra:property
         cc:csvw ;
     ] ;
+  hydra:supportedOperation
+    [
+      a
+        hydra:Operation,
+        schema:DeleteAction ;
+      hydra:title
+        "Delete Project" ;
```
martinmaillard, line 54: Should probably be "Delete table"

martinmaillard (approved):

apis/core/lib/domain/csv-mapping/delete.ts
```diff
@@ -21,6 +21,13 @@ export async function deleteMapping(csvMapping: NamedNode, store: ResourceStore)
     store.delete(sourceCollection)
   }
 
+  const tables = getTablesFromMapping(csvMapping)
+  for await (const table of tables) {
+    if (table.termType === 'NamedNode') {
+      await deleteTableWithoutSave(table, store)
+    }
```
tpluscode, line 28: I'd move the check inside the delete function (no-op if not named node)

```suggestion
    await deleteTableWithoutSave(table, store)
```

The parameter would become `Term`

apis/core/lib/domain/queries/table.ts
```diff
@@ -0,0 +1,23 @@
+import { NamedNode } from 'rdf-js'
+import { SELECT } from '@tpluscode/sparql-builder'
+
+import { parsingClient } from '../../query-client'
+import { cc } from '@cube-creator/core/namespace'
+
+export async function * getTablesFromMapping(csvMapping: NamedNode, client = parsingClient) {
```
tpluscode, line 7: A tiny rename maybe?

```suggestion
export async function * getTablesForMapping(csvMapping: NamedNode, client = parsingClient) {
```

apis/core/lib/domain/table/delete.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function deleteTableWithoutSave(resource: NamedNode<string>, store: ResourceStore): Promise<void> {
+  const table = await store.get(resource)
+  if (!table) return
+
+  // Delete in columnMappings
+  const columnMappings = table.out(cc.columnMapping).terms
+  for await (const columnMapping of columnMappings) {
+    if (columnMapping?.termType === 'NamedNode') {
+      await deleteColumnMapping(columnMapping, store)
+    }
```
tpluscode, line 30: Ditto
```suggestion
    await deleteColumnMapping(columnMapping, store)
```

tpluscode (changes requested):
Small changes + please fix what Martin pointed out

apis/core/hydra/table.ttl
lucafurrer: Yes

apis/core/lib/domain/csv-mapping/delete.ts
lucafurrer: Should I do the same thing with the sources?

tpluscode (approved):

PR #130, opened 2020-11-02
Create column mapping

vhf:
#119


apis/core/lib/handlers/column-mapping.ts
```diff
@@ -0,0 +1,16 @@
+import asyncMiddleware from 'middleware-async'
+import { protectedResource } from '@hydrofoil/labyrinth/resource'
+import { shaclValidate } from '../middleware/shacl'
+import { createColumnMapping } from '../domain/column-mapping/create'
+
+export const post = protectedResource(shaclValidate, asyncMiddleware(async (req, res) => {
+  const project = await createColumnMapping({
```
martinmaillard, line 7: Probably not `project`

apis/core/lib/handlers/column-mapping.ts
```diff
@@ -0,0 +1,16 @@
...
+    table: null,
```
martinmaillard, line 8: You should be able to get the table from the URI

apis/core/lib/domain/column-mapping/create.ts
```diff
@@ -0,0 +1,34 @@
...
+  resource: GraphPointer
+  store?: ResourceStore
+}
+
+export async function createColumnMapping({
+  table,
+  resource,
+  store = resourceStore(),
+}: CreateColumnMappingCommand): Promise<GraphPointer> {
+  const label = resource.out(schema.label).term!
```
martinmaillard, line 20: Not sure we're going to have a label. We haven't put that in the sample data.

apis/core/lib/domain/column-mapping/create.ts
```diff
@@ -0,0 +1,34 @@
...
+  const columnMapping = id.columnMapping(table, label.value)
+
+  const graph = table.out(columnMapping)
```
martinmaillard, line 23: `ColumnMapping` should be a resource. Check  https://github.com/zazuko/cube-creator/blob/master/apis/core/lib/domain/table/create.ts#L78, I already started a function to create a column mapping

apis/core/lib/domain/column-mapping/create.ts
vhf: I think it's required here: https://github.com/zazuko/cube-creator/blob/ff091d19307a0283134455aea97bd6c44b1851ce/apis/core/lib/domain/identifiers.ts#L30-L32

martinmaillard: It's using the `targetColumn` name. But I'm not even sure that was a good idea.

apis/core/lib/domain/column-mapping/create.ts
```diff
@@ -0,0 +1,36 @@
...
+  const columnName = column.out(schema.name).value!
+
+  const columnMapping = store.create(id.columnMapping(table, columnName))
+
+  columnMapping
+    .addOut(rdf.type, cc.ColumnMapping)
+    .addOut(cc.sourceColumn, column)
+    .addOut(cc.targetProperty, defaultProperty(columnName))
+
+  return columnMapping
```
martinmaillard, line 30: Now you're probably missing a `store.save()` somewhere. Look at what Luca did in #129, I think it's probably a good pattern

e2e-tests/column-mapping/create.hydra
```diff
@@ -0,0 +1,63 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
tpluscode, line 63: By using the seed data you will be able to start the test already at the source itself rather than recreating all the way from the top

Use the `ENTRYPOINT` keyword to declare the initial request to make

```suggestion
PREFIX api: <https://cube-creator.zazuko.com/vocab#>
PREFIX hydra: <http://www.w3.org/ns/hydra/core#>
PREFIX cc: <https://cube-creator.zazuko.com/vocab#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX schema: <http://schema.org/>

ENTRYPOINT "cube-project/ubd/csv-source/ubd"

HEADERS {
  x-user "john-doe"
  x-permission "pipelines:write"
}


With Class api:CSVSource {
    Expect Operation cc:CreateColumnMappingAction {
        Invoke {
            Content-Type "application/ld+json"

            ```
            ```
        } => {
            Expect Status 201
        }
    }
}
```

tpluscode (changes requested):

martinmaillard:
E2e tests are now failing because of the `sh:class` constraint on `cc.sourceColumn` but I have no idea why

e2e-tests/column-mapping/create.hydra
```diff
@@ -0,0 +1,43 @@
...
+            ```
+        } => {
+            Expect Status 201
+
+            Expect Type <https://cube-creator.zazuko.com/vocab#ColumnMapping>
+            Expect Property cc:targetProperty
+            Expect Property cc:sourceColumn
+            Expect Property cc:datatype
+            Expect Property cc:language
+            Expect Property cc:defaultValue
```
tpluscode, line 40: Wight wanna check for those precise values as sent above

packages/model/ColumnMapping.ts
```diff
@@ -9,6 +9,9 @@ import { initializer } from './lib/initializer'
 export interface ColumnMapping extends RdfResourceCore {
   sourceColumn: Link<CsvColumn>
   targetProperty: Term
+  datatype: NamedNode | null
+  language: string | null
+  defaultValue: Term | null
```
tpluscode, line 14: Slightly less verbose option

```suggestion
  datatype?: NamedNode
  language?: string
  defaultValue?: Term
```

tpluscode (changes requested):
> E2e tests are now failing because of the sh:class constraint on cc.sourceColumn but I have no idea why

The linked source column is loading empty. Guess the URI is incorrect?

packages/model/ColumnMapping.ts
martinmaillard: These are not equivalent if I'm not mistaken 🤔

martinmaillard:
Now the tests are passing for `push`, but not `pull_request` O_o

I have no idea why. But I also don't understand why we're running all the tasks for `push` *and* `pull_request`

lucafurrer (approved):
Looks fine to me

Issue #122, opened 2020-10-29
API: edit Cube metadata


martinmaillard:

tpluscode:
I think we can close this issue

PR #135, opened 2020-11-03
feat: meta data shape

lucafurrer:
Shape for Cube Meta Data

part of #122


apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -0,0 +1,90 @@
...
+      ${sh.order} 90 ;
+    ] ;
+    ${sh.property} [
+      ${sh.name} "Last Update" ;
+      ${sh.path} ${dcterms.modified} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${sh.datatype} xsd:date
+      ${sh.order} 100 ;
+    ] ;
```
tpluscode, line 87: I expect the dates to be automatically updated. The creation (I think there is `dcterm:created` by the way) would be set only once and `modified` should just update when making any changes

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -0,0 +1,90 @@
...
+      ${sh.order} 20 ;
+    ] ;
+    ${sh.property} [
+      ${sh.name} "Version" ;
+      ${sh.path} ${schema.version} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${sh.minLength} 1 ;
+      ${sh.order} 30 ;
+    ] ;
```
tpluscode, line 33: This one too, I suppose it would be set behind the scenes and only when we actually publish a cube. Might not even have it as part of the metadata at all

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -0,0 +1,90 @@
...
+      ${sh.order} 40 ;
+    ] ;
+    ${sh.property} [
+      ${sh.name} "Publish on Swiss Open Data" ;
+      ${sh.path} ${cc.PublishOnOpendata} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${sh.datatype} xsd:boolean
+      ${sh.order} 50 ;
+    ] ;
```
tpluscode, line 49: This is duplicated

```suggestion
    ${sh.property} [
      ${sh.name} "Publish on Swiss Open Data" ;
      ${sh.path} ${cc.PublishOnOpendata} ;
      ${sh.minCount} 1 ;
      ${sh.maxCount} 1 ;
      ${sh.datatype} xsd:boolean
      ${sh.order} 40 ;
    ] ;
```

Also, the property `cc.PublishOnOpendata` looks like a class (capital letter)

apis/core/bootstrap/shapes/dataset.ts
lucafurrer: Daniel used consequently issued in his document...

tpluscode: > Daniel used consequently issued in his document...

I don't understand

> I was wondering if the user should be able to overwrite it

Most certainly not. Isn't the update simply the moment in time when it happens? Why would the users set it themselves?

lucafurrer: Document of Dani with an example from Adrian? <https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/wikis/Dataset-Metadata#draft>

Since this date is displayed to the end users on opendata etc. I would be not surprised if somebody wants to adapt them. But I am fine with the fact, that they could be hidden or read only.

tpluscode: Ok, interesting point. Let's keep them then.

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -0,0 +1,104 @@
...
+      ${sh.datatype} xsd:boolean
+      ${sh.order} 40 ;
+    ] ;
+    ${sh.property} [
+      ${sh.name} "Publish on Visualize" ;
+      ${sh.path} ${dcterms.publishOnVisualize} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${sh.datatype} xsd:boolean
+      ${sh.order} 50 ;
```
lucafurrer, line 48: Just wondering, should those two actually be part of the cube project definition. They are not really meta data.

martinmaillard: I think they are metadata. Where would you put them?

lucafurrer: At the same level as the decision if it is a csv or cube. The question is who handles the publication, the cube project or the data set.

martinmaillard: These are information for the consumers of the cube, not for us. They basically allow IXT to know in which tool the cube should be shown.

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -0,0 +1,104 @@
...
+      ${sh.maxCount} 1 ;
+      ${sh.minLength} 1 ;
+      ${sh.order} 100 ;
+    ] ;
+    ${sh.property} [
+      ${sh.name} "Creation Date" ;
+      ${sh.path} ${dcterms.issued} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${sh.datatype} xsd:date
```
martinmaillard, line 91: It would be more consistent to use `${xsd.date}`. Same for the other datatypes.

lucafurrer:
I followed the latest state on the Document on https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/wikis/Dataset-Metadata

There are some special types like vcard.Organization or PeriodOfTime. Not sure if we have to do something more for those.

I also found something for the languages, no clue how this renders in the UI
```turtle
    ${sh.property} [
      ${sh.name} "Title" ;
      ${sh.path} ${dcterms.title} ;
      ${sh.minCount} 1 ;
      ${sh.minLength} 1 ;
      ${sh.datatype} ${rdf.langString} ;
      ${sh.languageIn} (en, de, fr, it) ;
      ${sh.order} 10 ;
    ] ;
```

apis/core/lib/domain/dataset/update.ts
```diff
@@ -0,0 +1,39 @@
+import { dcat, hydra, rdf, schema, _void } from '@tpluscode/rdf-ns-builders'
+import { GraphPointer } from 'clownface'
+import { NamedNode } from 'rdf-js'
+import { NotFoundError } from '../../errors'
+import { ResourceStore } from '../../ResourceStore'
+import { resourceStore } from '../resources'
+
+interface AddMetaDatatCommand {
```
tpluscode, line 8:
```suggestion
interface AddMetaDataCommand {
```

apis/core/lib/domain/dataset/update.ts
```diff
@@ -0,0 +1,39 @@
...
+  resource,
+  store = resourceStore(),
+}: AddMetaDatatCommand): Promise<GraphPointer> {
+  const datasetResource = await store.get(dataset.term)
+
+  if (!datasetResource) {
+    throw new NotFoundError(dataset)
+  }
+
+  datasetResource.deleteOut()
```
tpluscode, line 26: This will not be enough because any deep blank nodes will stay in the graph. You'd need to be more thorough, to descend the blank nodes children. Something like this, without getting too sophisticated

```suggestion
  datasetResource.out().forEach(child => {
    if (child.term === 'BlankNode') {
      child.deleteOut()
    }
  })
  datasetResource.deleteOut()
```

apis/core/lib/domain/dataset/update.ts
```diff
@@ -0,0 +1,39 @@
...
+  resource: GraphPointer
+  store?: ResourceStore
+
+}
+
+export async function update({
+  dataset,
+  resource,
+  store = resourceStore(),
+}: AddMetaDatatCommand): Promise<GraphPointer> {
```
tpluscode, line 19:
```suggestion
}: AddMetaDataCommand): Promise<GraphPointer> {
```

apis/core/lib/handlers/dataset.ts
```diff
@@ -0,0 +1,20 @@
...
+export const put = protectedResource(
+  shaclValidate,
+  asyncMiddleware(async (req, res) => {
+    const dataset = await update({
+      dataset: clownface(req.hydra.resource),
+      resource: await req.resource(),
+    })
+
+    res.status(200)
+    res.header('Location', dataset.term.value)
```
tpluscode, line 17: `Location` is only used with `201` status
```suggestion
```

apis/core/test/domain/dataset/update.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    expect(result.out(dcterms.title).value).to.eq(title)
+    expect(result.out(dcterms.description).value).to.eq(description)
+    expect(result.out(dcat.keyword).value).is.undefined
+    expect(dataset).to.matchShape({
+      property: {
+        path: rdf.type,
+        [sh.hasValue.value]: [schema.Dataset, _void.Dataset, dcat.Dataset],
+      },
+    })
+  })
```
tpluscode, line 42: Might wanna add a test to check that we remove the dataset triples but not the `cube:Cube`.

martinmaillard (approved):

tpluscode (changes requested):

apis/core/hydra/dataset.ttl
```diff
@@ -16,7 +16,7 @@ void:Dataset
         hydra:Operation,
         schema:UpdateAction ;
       hydra:title
-        "Update dataset" ;
+        "Edit cube metadata" ;
```
tpluscode, line 19: This is not exactly cube metadata either 🤔 🤷 
```suggestion
        "Edit metadata" ;
```

apis/core/test/domain/dataset/update.test.ts
```diff
@@ -5,12 +5,14 @@ import $rdf from 'rdf-ext'
...
 import { TestResourceStore } from '../../support/TestResourceStore'
 import { update } from '../../../lib/domain/dataset/update'
+import { cube } from '@cube-creator/core/namespace'
 
 describe('domain/dataset/update', () => {
   let store: TestResourceStore
   const dataset = clownface({ dataset: $rdf.dataset(), term: $rdf.namedNode('dataset') })
     .addOut(dcterms.title, 'My Title')
     .addOut(dcat.keyword, 'Test')
+  const cubeNode = dataset.namedNode('cube').addOut(rdf.type, cube.Cube)
```
tpluscode, line 15: Almost there. I would only just add one more triple to ensure that the cube and dataset remain connected
```suggestion
  const cubeNode = dataset.namedNode('cube')
    .addOut(rdf.type, cube.Cube)
    .addIn(schema.hasPart, dataset)
```

apis/core/test/domain/dataset/update.test.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    // then
+    expect(result.out(dcterms.title).value).to.eq(title)
+    expect(result.out(dcterms.description).value).to.eq(description)
+    expect(result.out(dcat.keyword).value).is.undefined
+    expect(dataset).to.matchShape({
+      property: {
+        path: rdf.type,
+        [sh.hasValue.value]: [schema.Dataset, _void.Dataset, dcat.Dataset],
+      },
```
tpluscode, line 42: And a matching assertion

```suggestion
      property: [{
        path: rdf.type,
        [sh.hasValue.value]: [schema.Dataset, _void.Dataset, dcat.Dataset],
      }, {
        path: schema.hasPart,
        minCount: 1,
      }],
```

tpluscode (changes requested):
Thank you @lucafurrer. Just final touches, please bear with me 🐻

lucafurrer:
> Thank you @lucafurrer. Just final touches, please bear with me bear

Thanks for the review :+1:

apis/core/hydra/dataset.ttl
martinmaillard: I just changed that!
We always talked about "cube metadata" with customers, so that's why I think it makes sense to use the same wording here.

tpluscode (approved):

PR #133, opened 2020-11-02
refactor: use OOP and more tests for creating csv source

tpluscode:
* completes the `CsvSource` model interface according to the seed data
* implements proper tests for created data
* turns the domain logic into more OOP

PR #136, opened 2020-11-03
Edit source settings

martinmaillard:


tpluscode:
@martinmaillard are we good to merge #133 and rebase here to continue on the API end?

martinmaillard:
Missing:
- [ ] API endpoint to update source

It should update the source dialect and re-parse the CSV file.

apis/core/lib/domain/csv-mapping/update.ts
```diff
@@ -0,0 +1,58 @@
...
+  store = resourceStore(),
+  fileStorage = s3,
+}: UpdateCsvSourceCommand): Promise<GraphPointer> {
+  const changed = RdfResource.factory.createEntity<CsvSource>(resource)
+  const csvSource = await store.getResource<CsvSource>(resource.term)
+  if (!csvSource) {
+    throw new NotFoundError(resource)
+  }
+
+  csvSource.name = changed.name
```
tpluscode, line 33: Now that you removed the name from the shape, we'd need to remove this assignment too, lest a request clears the name

Or we could bring back the property shape with `dash:hidden`

tpluscode (approved):

apis/core/lib/domain/csv-mapping/update.ts
martinmaillard: Ok I'll re-add the property with `dash:hidden`

lucafurrer (approved):

PR #137, opened 2020-11-03
feat: github action to run pipeline

tpluscode:
This PR extends the transformation pipeline and the Job class, and adds a GitHub Action to run the tranformation


fuseki/sample-ubd.trig
```diff
@@ -326,6 +326,8 @@ graph <project/ubd/csv-mapping/jobs/test-job> {
     cc:tables <cube-project/ubd/csv-mapping/tables> ;
     schema:label "cli-test" ;
     cc:cubeGraph <cube-project/ubd/cube-data> ;
-    dcterms:created "2020-10-29T12:01:54"^^xsd:dateTime
+    dcterms:created "2020-10-29T12:01:54"^^xsd:dateTime ;
+    rdfs:seeAlso <https://github.com/zazuko/cube-creator/actions/runs/345020887> ;
+    schema:actionStatus schema:PotentialActionStatus ;
```
tpluscode, line 331: I propose to extend `cc:Job` from [`schema:Action`](http://schema.org/Action) and use its properties to keep track of the job's state

.github/workflows/ci.yaml
```diff
@@ -35,6 +35,10 @@ jobs:
       - name: logs on fail
         if: ${{ failure() }}
         run: lando logs -s core
+      - name: Codecov
+        uses: codecov/codecov-action@v1.0.5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
lucafurrer, line 41: How comes that this is a change?

apis/core/lib/storage/s3.ts
```diff
@@ -1,5 +1,5 @@
 import aws from 'aws-sdk'
-import { isReadable } from 'isstream'
+import { readable } from 'is-stream'
```
lucafurrer, line 2: What is the difference between those two libraries?

lucafurrer (approved):
I have not run the code, but it looks fine :+1:

apis/core/lib/storage/s3.ts
tpluscode: Not much I guess other than the new is only 2 years old while the other was last released 6 yrs ago :)

.github/workflows/ci.yaml
tpluscode: Missed that in #152 and now it was bringing the % down

PR #138, opened 2020-11-03
feat(ui): add URI input

martinmaillard:
Bare bones URI input. It does nothing special apart from creating a named node from a text input.

### DIFF
--- a/ui/src/forms/URIInput.vue
+++ b/ui/src/forms/URIInput.vue
@@ -5,7 +5,7 @@
 <script lang="ts">
 import { Prop, Component, Vue } from 'vue-property-decorator'
 import { NamedNode } from 'rdf-js'
-import $rdf from '@rdfjs/data-model'
+import * as $rdf from '@rdf-esm/data-model'
 
 @Component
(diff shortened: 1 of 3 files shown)


ui/src/forms/URIInput.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+  <b-input :value="textValue" @input="onUpdate" />
+</template>
+
+<script lang="ts">
+import { Prop, Component, Vue } from 'vue-property-decorator'
+import { NamedNode } from 'rdf-js'
+import $rdf from '@rdfjs/data-model'
```
tpluscode, line 8:
```suggestion
import * as $rdf from '@rdf-esm/data-model'
```

martinmaillard: Ok, can you please explain this to me? So that I don't make the mistake again ;)

tpluscode: `@rdf-esm/(.+)` are web-optimized forks of their `@rdfjs/$1` counterparts

PR #139, opened 2020-11-03
Setup Cypress e2e tests for UI

martinmaillard:


martinmaillard:
Seems like my env variables hack doesn't work. I'll have to check that.

vhf (approved):
Nice but I couldn't figure how to run the tests locally. I tried `yarn test:e2e` in `./ui` and `yarn --cwd ui test:e2e` at the root, I'm probably missing a step somewhere.

Probably worth adding here as well: https://github.com/zazuko/cube-creator/blob/master/README.md#e2e-tests

PR #141, opened 2020-11-03
feat: use cc:identifier template for csvw:aboutUrl

tpluscode:
With this PR the generated csvw mapping will combine the mapping's `cc:namespace` and table's `cc:identifierTemplate` to create the observation URL

### DIFF
--- a/fuseki/sample-ubd.trig
+++ b/fuseki/sample-ubd.trig
@@ -206,7 +206,8 @@ graph <project/ubd/csv-mapping/table-observation> {
     cc:csvSource <cube-project/ubd/csv-source/ubd> ;
     schema:name "Observations" ;
     schema:color "#AAAAAA" ;
-    cc:identifierTemplate "ammonia/observation/{STATION}-{YEAR}-annualmean" ;
+    # cc:identifierTemplate "ammonia/observation/{STATION}-{YEAR}-annualmean" ;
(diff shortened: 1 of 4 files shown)


fuseki/sample-ubd.trig
```diff
@@ -206,7 +206,8 @@ graph <project/ubd/csv-mapping/table-observation> {
     cc:csvSource <cube-project/ubd/csv-source/ubd> ;
     schema:name "Observations" ;
     schema:color "#AAAAAA" ;
-    cc:identifierTemplate "ammonia/observation/{STATION}-{YEAR}-annualmean" ;
+    # cc:identifierTemplate "ammonia/observation/{STATION}-{YEAR}-annualmean" ;
+    cc:identifierTemplate "{station_id}-{year}-{aggregation_id}" ;
```
tpluscode, line 210: Keeping the "correct" template around while the csvw is still fake and not built from all the mappings

martinmaillard (commented):
This means we don't support absolute identifier templates, right? I think this may be an issue for non-observation tables, as we may actually want to reference existing resources.

tpluscode:
I have not thought this through yet but may be like you say. Note how the test is called

> constructs observation table aboutUrl by concatenating namespace and id template

While the alternative test cases are not yet implemented, I have not considered the other possibilities

martinmaillard (approved):

PR #142, opened 2020-11-03
fix: use cookie to store OIDC state

sandhose:
In private browsing session, Firefox treats localStorage as sessionStorage, meaning that it gets emptied whenever the site is leaved. This is a problem with the oidc-client, because it uses the localStorage to store informations about the current OIDC flow.

This patch changes the state store to use `cookie-storage`, which is a WebStorage-compatible layer on top of cookies.


martinmaillard (approved):

tpluscode:
Wait, isn't that the purpose of private mode?

vhf (approved):

PR #145, opened 2020-11-04
fix(ui): enable auto silent token refresh

### LINKED ISSUE #144: Auth token refresh
For some reason, I have to re-authenticate every few minutes. I'm guessing it's caused by:
- a very short ttl for the auth token
- a broken refresh mechanism

@sandhose do you have any idea how we can fix that?

martinmaillard:
Will close #144


sandhose (approved):
Haven't tested but there is no reason this shouldn't work

PR #151, opened 2020-11-05
fix(shacl): ability to use sh:targetNode to match request URI

tpluscode:
Requests such as `PUT` or especially `PATCH` may not explicitly include a class triple and thus `sh:targetClass` makes no sense for validation. However, without a target no validation happens.

For these scenarios the SHACL middleware will now assume `sh:targetNode` to be the request URI (`req.hydra.term`)


martinmaillard:
That makes sense.

I feel like it would even make sense for POST requests. Isn't it weird to require the client to provide a type for the payload?

tpluscode:
I like that too. The only scenario would be a POST which accepts multiple different classes, each with its own shape.

But with `targetClass` I would propose something opposite - to inject the `rdf:type` into the payload dataset

martinmaillard (approved):

PR #155, opened 2020-11-09
Pipeline GitHub workflow

tpluscode:
Some tweaks to the GitHub pipeline workflow

1. fix the yaml definition
1. have the process actually fail when pipeline fails
1. added a `shape/job/trigger` to provide GitHub Personal Access token. we will only use GitHub on test environment, where it should be good enough for anyone to use their own token. might store that in local storage to make it a little less problematic


apis/core/hydra/job.ttl
```diff
@@ -41,6 +44,8 @@ cc:JobCollection
         "Start transformation" ;
       hydra:method
         "POST" ;
+      hydra:expects
+        <shape/job/trigger> ;
```
martinmaillard, line 48: This file is missing `<shape/job/trigger> a sh:Shape .`

tpluscode: good catch 🏉

martinmaillard (approved):

PR #162, opened 2020-11-10
fix: shacl validator should not always fail on empty bodies

tpluscode:


martinmaillard (approved):

apis/core/lib/middleware/shacl.ts
```diff
@@ -17,6 +16,13 @@ interface ShaclMiddlewareOptions {
 export const shaclMiddleware = (options: ShaclMiddlewareOptions) => asyncMiddleware(async (req, res, next) => {
   const resources = options.createResourceStore()
 
+  let resource: GraphPointer
+  if (!req.dataset) {
+    resource = clownface({ dataset: $rdf.dataset() }).node(req.hydra.term)
+  } else {
+    resource = await req.resource()
+  }
```
tpluscode, line 24: Ensure there always is a "body resource", even if the request body was empty

apis/core/lib/middleware/shacl.ts
```diff
@@ -28,22 +34,15 @@ export const shaclMiddleware = (options: ShaclMiddlewareOptions) => asyncMiddlew
       if (pointer.out([sh.targetClass, sh.targetNode, sh.targetObjectsOf, sh.targetSubjectsOf]).values.length === 0) {
         shapes.add($rdf.quad(pointer.term, sh.targetNode, req.hydra.term))
       }
+
+      resource.addOut(rdf.type, pointer.out(sh.targetClass))
```
tpluscode, line 38: I put this in the loop to ensure that only the root shape's `sh:target` gets added to the resource.

PR #166, opened 2020-11-10
Fix missing content-type when uploading CSV

martinmaillard:


lucafurrer (approved):
Does the back end do something with this information

martinmaillard:
Yeah, at some point the backend started trying to parse the CSV as JSON because the `content-type` was wrong. I'm not sure why it wasn't doing that before.

tpluscode:
It's the shacl middleware which tried to get the body for validation and it tries to parse csv as RDF

PR #167, opened 2020-11-11
feat: remove source link from table when deleting

### LINKED ISSUE #97: Remove link from Table when a CSV-Source is deleted
When a CSV-Source is deleted, the corresponding table should remove its link to the source and add an warning `schema:warning`

Follow up task to #88

lucafurrer:
Removes the link to the csvSource from the table when the source is deleted.

I did a refactoring to make the code testable and wrote a test

fixes #97

### DIFF
--- a/apis/core/lib/domain/csv-source/delete.ts
+++ b/apis/core/lib/domain/csv-source/delete.ts
@@ -6,6 +6,7 @@ import { schema } from '@tpluscode/rdf-ns-builders'
 import { cc } from '@cube-creator/core/namespace'
 import $rdf from 'rdf-ext'
 import * as TableQueries from '../queries/table'
+import { Table } from '@cube-creator/model'
 
 interface DeleteSourceCommand {
   resource: NamedNode | Term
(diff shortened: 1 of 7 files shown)


apis/core/lib/domain/csv-source/delete.ts
```diff
@@ -1,40 +1,54 @@
...
 
-  const csvSource = await store.get(csvSourceTerm)
+  const csvSource = await store.get(resource)
   if (!csvSource) return
 
-  // TODO: Delete Tables or prevent it
-  // Find related tables
-  // Delete them
+  // Remove links from tables
+  const tables = await getLinkedTablesForSource(csvSource.term)
```
tpluscode, line 41: Is it necessary to await here?

```suggestion
  const tables = getLinkedTablesForSource(csvSource.term)
```

apis/core/lib/domain/csv-source/delete.ts
```diff
@@ -1,40 +1,54 @@
...
+  for await (const table of tables) {
+    const tableGraph = await store.get(table)
+    tableGraph?.dataset.delete($rdf.quad(tableGraph.term, cc.csvSource, csvSource.term))
+  }
```
tpluscode, line 45: I would shift this into table itself. At least if it was assigning `undefined`

```suggestion
  for await (const tableId of tables) {
    const table = await store.getResource<Table>(tableId)
    if (table) {
      table.csvSource = undefined
    }
  }
```

Which also reveals that `Table#csvSource` is not typed to allow falsy values. Removing this link without adjust the interface risks undetected runtime errors

apis/core/lib/domain/csv-source/delete.ts
lucafurrer: True, we have that in the `for await`

lucafurrer: Or not ` TypeError: tables is not async iterable`

martinmaillard (approved):

PR #168, opened 2020-11-11
refactor: unify models

tpluscode:
I saw that the UI project received a reference to the shared models so I went for putting all eggs in one basket

This PR removes much of what was duplicated in the UI code keeping only the necessary extensions


lucafurrer (approved):
:+1:

martinmaillard:
I would have loved to test that before merged. I hope you tested it 😅

PR #169, opened 2020-11-11
feat: start implementation of csvw builder

tpluscode:


ui/src/api/mixins/ApiResource.ts
```diff
@@ -32,6 +32,10 @@ export function apiResourceMixin (rootUrl: string, separator: string): Mixin {
       get actions (): Actions {
         return commonActions(this, this._additionalActions)
       }
+
+      get inlineRepresentation () {
```
martinmaillard, line 36: Not sure I understand this

martinmaillard:
If I put a space in the `targetProperty` of a column mapping, it crashes the pipeline. Not sure if we should prevent that or just escape spaces.

ui/src/api/mixins/ApiResource.ts
tpluscode: Right, this is a little annoying technicality, for which I could not find a good solution. 

Take [Table#columnMappings](https://github.com/zazuko/cube-creator/pull/169/files#diff-952eebae7242509177c8aa88b60ec294a4c0fd9f04e5a20554993a7f2464e184R51) for an example. It is typed as an array of `Link<ColumnMapping>`. A link is deliberately typed as a plain resource with a `load` method to prevent accidentally adding triples about the linked resource inside the Table's graph.

But reconsidering this, an explicit cast can work just as well without adding meaningless properties

martinmaillard: TIL what a `Link` is :)

fuseki/sample-ubd.trig
```diff
@@ -216,7 +216,7 @@ graph <cube-project/ubd/csv-source/ubd> {
 
   <cube-project/ubd/csv-source/ubd/column/year>
     a csvw:Column ;
-    schema:name "YEAR" ;
+    schema:name "year" ;
```
martinmaillard, line 219: Why this change?

fuseki/sample-ubd.trig
```diff
@@ -386,8 +386,7 @@ graph <cube-project/ubd/csv-mapping/table-station> {
     cc:csvw <cube-project/ubd/csv-mapping/table-station/csvw> ;
     schema:name "Station" ;
     schema:color "#BBBBBB" ;
-    # cc:identifierTemplate "station/{STATION_ID}" ;
-    cc:identifierTemplate "{station_id}-{year}-{aggregation_id}" ;
+    cc:identifierTemplate "station/{STATION_ID}" ;
```
martinmaillard, line 389: The `STATION_ID` variable is wrong now that you downcased the column names

packages/model/Table.ts
```diff
@@ -18,7 +19,7 @@ export interface Table extends RdfResource {
   csvMapping: Link<CsvMapping>
   identifierTemplate: string
   color: string
-  columnMappings: ColumnMapping[]
+  columnMappings: Array<Link<ColumnMapping>>
```
martinmaillard, line 22: In the UI, they are not really `Link`s right? I see that you fixed it by "convincing" typescript that they are full resources but I wonder if we are not already hitting some of the limits of sharing these models.

apis/core/lib/csvw-builder/index.ts
```diff
@@ -0,0 +1,102 @@
...
+      warning(`Column ${columnMapping.sourceColumn.id} not found`)
+      continue
+    }
+
+    yield mappedColumn({ csvMapping, columnMapping, column, source, resources })
+  }
+
+  for (const [, column] of unmappedColumns) {
+    yield unmappedColumn({ column })
+  }
```
lucafurrer, line 68: I find it rather strange that the function `buildColumns` returns the ignored columns.

apis/core/lib/csvw-builder/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  const source = await resources.getResource<CsvSource>(table.csvSource?.id)
+  if (!source) {
+    throw new Error(`Source not found for table ${table.id.value}`)
+  }
+  const csvMapping = await resources.getResource<CsvMapping>(table.csvMapping.id)
+  if (!csvMapping) {
+    throw new Error(`CSV Mapping resource not found for table ${table.id.value}`)
+  }
+
+  const column: Initializer<Csvw.Column>[] = [{
```
lucafurrer, line 81: since it is an array column**s**?

packages/model/Table.ts
tpluscode: This is just annotation so yes, the properties are actual resource objects.

That aside, I have been thinking about this and it is true that this is a contention point between the two usages of a shared model. In the long run I would rather that the UI model interface are for example generated from a SHACL shapes.

This would especially useful for the designer view.

apis/core/lib/csvw-builder/index.ts
lucafurrer: Does column not contain several columns? 
`Initializer<Csvw.Column>[]` and  `column.push(csvwColumn)` made me think it is an array of columns

packages/model/Table.ts
martinmaillard: Yeah... let's not go into this for now ;)

fuseki/sample-ubd.trig
tpluscode: Done. Although we might want to sync the test dataset and the sample file

martinmaillard (approved):

PR #172, opened 2020-11-12
Input field for RDF properties

martinmaillard:
Merge after #130


lucafurrer:
grafik
The generated URL does not look correct. I think there should be a `/` before test, no?

martinmaillard:
I think it's the responsibility of the namespace to include a separator. @tpluscode ?

Maybe we should add a trailing slash in the default namespace?

tpluscode:
Good idea 👍 . I wanted to say about hash URIs but maybe they don't make sense

PR #173, opened 2020-11-12
feat: mapping basic properties to csvw columns

tpluscode:
Add tests and code to produce basic column CSVW mappings:

1. suppressed columns where there is not column mapping
2. concatenate `cc:namespace` and `cc:targetProperty` to build `csvw:propertyUrl`


apis/core/lib/csvw-builder/index.ts
```diff
@@ -1,17 +1,91 @@
...
+
+  initializer.suppressOutput = true
+
+  return initializer
+}
+
+function mappedColumn({ columnMapping, column, source, namespace }: { column: CsvColumn; columnMapping: ColumnMapping; source: CsvSource; namespace: NamedNode }): Initializer<Csvw.Column> {
+  return {
+    title: $rdf.literal(column.name),
+    propertyUrl: `${namespace.value}${columnMapping.targetProperty.value}`,
```
lucafurrer, line 30: Wouldn't this be something for the identifiers module?

tpluscode: Not sure. This is about output URIs and not resources we have in our db

lucafurrer (approved):

PR #174, opened 2020-11-13
fix: force identifier template

lucafurrer:
Forces a identifier template since we do not have logic for auto generation.

Changed e2e test for table create such that it actually does something.

### DIFF
--- a/apis/core/bootstrap/shapes/table.ts
+++ b/apis/core/bootstrap/shapes/table.ts
@@ -36,10 +36,11 @@ ${shape('table/create')} {
     ] ;
     ${sh.property} [
       ${sh.name} "Identifier template" ;
-      ${sh.description} "Used to build a unique identifier for each row of this table. Leave empty to get an auto-generated identifier." ;
+      ${sh.description} "Used to build a unique identifier for each row of this table. Not supported yet: Leave empty to get an auto-generated identifier." ;
       ${sh.path} ${cc.identifierTemplate} ;
       ${sh.minCount} 1 ;
       ${sh.maxCount} 1 ;
+      ${sh.minLength} 1 ;
       ${sh.defaultValue} "" ;
       ${sh.order} 30 ;
     ] ;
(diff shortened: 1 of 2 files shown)


apis/core/bootstrap/shapes/table.ts
```diff
@@ -36,10 +36,11 @@ ${shape('table/create')} {
     ] ;
     ${sh.property} [
       ${sh.name} "Identifier template" ;
-      ${sh.description} "Used to build a unique identifier for each row of this table. Leave empty to get an auto-generated identifier." ;
+      ${sh.description} "Used to build a unique identifier for each row of this table. Not supported yet: Leave empty to get an auto-generated identifier." ;
       ${sh.path} ${cc.identifierTemplate} ;
       ${sh.minCount} 1 ;
       ${sh.maxCount} 1 ;
+      ${sh.minLength} 1 ;
```
tpluscode, line 43: Tell you what, a `sh:pattern "[\w\d]+"` would be event better, to prevent whitespace-only

lucafurrer: Pattern is happy when the RegExp finds a matching pattern in the string or does the entire string match

lucafurrer: Either `[\w\d]` is sufficient or it would not work since there will be `-` and  `{}`in the entry

tpluscode: Aha, good point. For entire value you'd need `^` and `$`

lucafurrer: Until we know exactly what is needed i'll go with `sh:pattern "[\w\d]"`

tpluscode: >Either [\w\d] is sufficient or it would not work since there will be - and {}in the entry

Good point. Might as well run with `.+` to avoid going into template syntax specific.

tpluscode (approved):

PR #176, opened 2020-11-16
fix: form updates

tpluscode:
Bringing in a host of improvements to the form

1. Click ➕ multiple times and handle those fields independently
   * **CAUTION** field `object` now an be undefined, when in a "no value state"
2. Update how elements are lazy-loaded
3. Fix checkbox (`update` must be called with an RDF/JS literal)
4. Nested `cc-form` should just work, honouring the shapes graph pointer
5. Nested blank nodes generated by `sh:minCount` are correctly connected the parent focus node

### DIFF
--- a/ui/src/forms/components.ts
+++ b/ui/src/forms/components.ts
@@ -1,115 +1,96 @@
-import { html, SingleEditorComponent } from '@hydrofoil/shaperone-wc'
+import { html, SingleEditorComponent, Lazy } from '@hydrofoil/shaperone-wc'
 import * as ns from '@cube-creator/core/namespace'
 import { dash, xsd } from '@tpluscode/rdf-ns-builders'
 import $rdf from '@rdfjs/data-model'
 import { createCustomElement } from '@/forms/bulma'
 
-export const textField: SingleEditorComponent = {
+export const textField: Lazy<SingleEditorComponent> = {
   editor: dash.TextFieldEditor,
-  render ({ value }, { update }) {
-    return html`<b-textbox .value="${value.object.value}" .update="${update}"></b-textbox>`
-  },
-  loadDependencies () {
-    return [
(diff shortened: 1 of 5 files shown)


ui/src/forms/components.ts
```diff
@@ -1,115 +1,96 @@
...
-    return html`<b-textbox .value="${value.object.value}" .update="${update}"></b-textbox>`
-  },
-  loadDependencies () {
-    return [
-      import('./BulmaTextBox.vue').then(createCustomElement('b-textbox'))
-    ]
+  async lazyRender () {
+    await import('./BulmaTextBox.vue').then(createCustomElement('b-textbox'))
+
+    return ({ value }, { update }) => html`<b-textbox .value="${value.object?.value || ''}" .update="${update}"></b-textbox>`
```
tpluscode, line 12: I ditched the `loadDependencies` and replaced it with a `lazyRender` which returns the render function. This way it is better to put the whole implementation in a separate module. Should produce more useful bundles

martinmaillard:
I get this on the column-mapping form:
Screenshot 2020-11-16 at 11 42 14

lucafurrer:
> I get this on the column-mapping form:
> Screenshot 2020-11-16 at 11 42 14

I had the impression, that I clicked everywhere and all worked...

lucafurrer (approved):
Fixes my problem. I was not able to reproduce the problem @martinmaillard had. But I think it would be good to understand it

tpluscode:
It happens sometimes when you open that form right away. If you close it and open again it would mostly disappear.

The issue has nothing to do with the changes in this dataset. It appears to be caused by clashing blank node identifiers, coming from multiple data factories

martinmaillard (approved):
If it's not related, we can fix the blank node issue in a separate PR. This one looks good

PR #179, opened 2020-11-16
feat(ui): improve form error messages

martinmaillard:
This is still not very user friendly, but at least it will show some details about the error that happened when submitting the form.


lucafurrer (approved):
Not perfect yet, but still a huge improvement.

PR #184, opened 2020-11-16
perf: reduce bundles, improve operation lookup

tpluscode:
**rdfine improvements**

This PR bundles an improved alcaeus which is more efficient doing operation lookups.

**Bundle optimisations**

A bunch of import optimisations to reduce the bundle. Vue's setup is super eager to bundle a top of unused modules. With a little effort I reduced the current size of the `vendor` bundle by 60%

```diff
File                            Size                         Gzipped

-dist/js/chunk-vendors.js       9624.72 KiB                  830.79 KiB
+dist/js/chunk-vendors.js       1929.90 KiB                  498.58 KiB
```

The size is till substantial though and might require more careful refactoring of how the `@rdfine/*` packages manage their imports


martinmaillard (approved):
nice

tpluscode:
Unfortunately this breaks imports in node so it'll need some more effort to refactor the `@rdfine/*` packages.

On the other hand `sideEffects: false` also helps heaps to curb the unnecessary bundling

tpluscode:
It can go further down to 2400 kB uncompressed, at which point the biggest module is font awesome SVG :)

martinmaillard (approved):

PR #190, opened 2020-11-17
Handle "delete column mapping" and "edit table" operations

martinmaillard:
The operations don't edit in the API yet, but they should work right away in the UI once we add them.

### DIFF
--- a/apis/core/hydra/table.ttl
+++ b/apis/core/hydra/table.ttl
@@ -62,6 +62,18 @@ cc:Table
             <file:handlers/table/delete#remove> ;
         ] ;
     ] ,
+    # [
+    #   a
+    #     hydra:Operation,
+    #     schema:UpdateAction ;
+    #   hydra:title
+    #     "Edit table" ;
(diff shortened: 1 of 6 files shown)


apis/core/hydra/table.ttl
```diff
@@ -62,6 +62,18 @@ cc:Table
...
+    #     hydra:Operation,
+    #     schema:UpdateAction ;
+    #   hydra:title
+    #     "Edit table" ;
+    #   hydra:method
+    #     "PUT" ;
+    #   hydra:expects
+    #     cc:Table,
+    #     <shape/table/update> ;
+    # ],
```
tpluscode, line 76: Maybe we can implement it already? It will fail if called but at least we can verify the UI bind correctly

martinmaillard: I already verified the UI bind. I commented it out to avoid users reporting bugs before the feature is actually implemented.

tpluscode (approved):

PR #193, opened 2020-11-17
Various UI improvements

martinmaillard:


lucafurrer (approved):
The orange scares from clicking the buttons :wink: 
Otherwise the mouse travel to click on Project Options and then edit the options is huge. I preferred it the old way, but I think I can get used to it.

martinmaillard:
That's a good remark, I'm not a big fan of using orange as primary color either. I just wanted to avoid using the default Bulma color, but not sure how to chose another color. We don't have any guidelines on that so I just used Zazuko's official orange ;)
Maybe I'll just revert that for now... it's not really important.

lucafurrer:
I can provide you with some Trivadis colors if you want :laughing: We have plenty of colors
grafik

martinmaillard:
I just reverted the change ;)

(this "reserve" red looks even more scary than the orange)

PR #196, opened 2020-11-18
refactor: re-remove raw-loader and copy expand into a local module

tpluscode:
Removed the need for raw-loader by copying the expand function from `@zazuko/rdf-vocabularies`. ~~Also memoize the properties so that the datasets are not processed every time one hits `loadCommonProperties`~~


martinmaillard:
Wait, they are already "memoized" in the store. There's no need to put them in the store if we memoize them in the load function.

tpluscode:
Of course. Let me simplify then

PR #199, opened 2020-11-18
test(ui): add more e2e tests

martinmaillard:


ui/src/components/MapperTable.vue
```diff
@@ -21,6 +21,7 @@
             <hydra-operation-button
               :operation="table.actions.delete"
               @click="deleteTable(table)"
+              data-testid="delete-table"
```
tpluscode, line 24: that's a rather ugly thing to do

martinmaillard: Just following [good practices](https://docs.cypress.io/guides/references/best-practices.html#Selecting-Elements). Test IDs are not always required, but they tend to avoid breaking tests that should not break (style change, wording change, etc.).

tpluscode: Interesting. Thanks

PR #200, opened 2020-11-18
feat: endpoint to update dimension metadata

tpluscode:


apis/core/hydra/dimension-metadata.ttl
```diff
@@ -0,0 +1,43 @@
...
+  hydra:supportedProperty
+    [ hydra:property
+        schema:hasPart ] ;
+  hydra:supportedOperation
+    _:DimensionUpdateOperation ;
+.
+
+schema:hasPart
+  hydra:supportedOperation
+    _:DimensionUpdateOperation .
```
tpluscode, line 20: It's slightly weird how the operation is attached to both the collection as well as the `schema:hasPart` property.

Only the latter should be necessary (request to the objects of this property) but hydra-box does not recognise URIs with query strings `?dimension=XYZ` as an independent resource and instead looks for operations on an ID with the query parameters removed.

cc @bergos

fuseki/sample-ubd.trig
```diff
@@ -46,16 +46,16 @@ graph <cube-project/ubd/dataset> {
...
     schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/station> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/aggregation> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/pollutant> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/unit-id> ;
   .
 
-  <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/shape/dimension%20year>
-    schema:about <https://environment.ld.admin.ch/foen/ubd/28/dimension%20year> ;
+  <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/dimension-year>
+    schema:about <https://environment.ld.admin.ch/foen/ubd/28/dimension-year> ;
```
tpluscode, line 58: Unfortunately (RDF/)JS is rather bad at comparing same URIs with mixed up escaping. We will have to avoid escaped characters in the property identifiers

packages/core/namespace.ts
```diff
@@ -55,8 +56,13 @@ type MetaDataProperty =
 
 type CubeCreatorNamespace = Record<CubeCreatorClass | CubeCreatorProperty | OtherTerms | MetaDataProperty, NamedNode>
 
+prefixes.scale = 'http://ns.bergnet.org/cube/scale/'
+prefixes.freq = 'http://purl.org/cld/freq/'
+prefixes.cube = 'http://ns.bergnet.org/cube/'
```
tpluscode, line 61: I completely forgot. It is possible to add a prefix to `@zazuko/rdf-vocabularies`. `rdf-string` will also honor them when interpolating named nodes

apis/core/hydra/dimension-metadata.ttl
martinmaillard: Back with my question of the other day then: do we really need these URIs with query string? What do they bring us over URIs with a random value for example?

fuseki/sample-ubd.trig
martinmaillard: I think that makes sense anyway. We just have to make sure we handle auto-generated properties correctly, but I doubt users need escaped characters in properties they choose manually.

apis/core/hydra/dimension-metadata.ttl
tpluscode: Yes, you do make a compelling case. There is nothing in favour of `?dimension=foo` vs any URI constructed without a query string

fuseki/sample-ubd.trig
```diff
@@ -46,21 +46,21 @@ graph <cube-project/ubd/dataset> {
...
-    schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/station> ;
-    schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/aggregation> ;
-    schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/pollutant> ;
-    schema:hasPart <cube-project/ubd/dimensions-metadata?dimension=https://environment.ld.admin.ch/foen/ubd/28/unit-id> ;
+  <cube-project/ubd/dimensions-metadata> a cc:DimensionMetadataCollection , hydra:Resource ;
+    schema:hasPart <cube-project/ubd/dimensions-metadata/dimension-year> ;
+    schema:hasPart <cube-project/ubd/dimensions-metadata/station> ;
+    schema:hasPart <cube-project/ubd/dimensions-metadata/aggregation> ;
+    schema:hasPart <cube-project/ubd/dimensions-metadata/pollutant> ;
+    schema:hasPart <cube-project/ubd/dimensions-metadata/unit-id> ;
```
martinmaillard, line 54: So you are using the slugified column name for the dimension? It will add a random value at the end right?

tpluscode: Technically it can be anything. Do we have to make it forcefully unique? I thought to simply append the `targetProperty` string or shrunk URI like `cube-project/ubd/dimensions-metadata/schema:age`

martinmaillard: Yeah that works for me. I just wanted to make sure you were not intending to use the CSV column name because one column can definitely be mapped to multiple properties.

martinmaillard (approved):

PR #203, opened 2020-11-18
fix(ui): memoize API documentation

martinmaillard:
I applied the same hack that we did in the first project. @tpluscode let me know if there's a better way.

### DIFF
--- a/ui/src/api/index.ts
+++ b/ui/src/api/index.ts
@@ -1,7 +1,7 @@
 import { Hydra } from 'alcaeus/web'
 import { RdfResource, RuntimeOperation } from 'alcaeus'
 import { RdfResourceCore } from '@tpluscode/rdfine/RdfResource'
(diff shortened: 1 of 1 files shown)


tpluscode:
`Hydra.cacheStrategy.shouldLoad` is a bool function hook which lets you decide if previously loaded resource should be returned without hitting the API

```typescript
Hydra.cacheStrategy.shouldLoad = (previous) => {
  return previous.representation.root?.types.has(hydra.ApiDocumentation)
}
```

martinmaillard:
Nice, done

tpluscode:
The failed cypress test is just a fluke?

martinmaillard:
I don't understand why:
1. We have a test run for "push" *and* one for "pull_request"
2. One of the two sometimes fails when the other one doesn't

Issue #187, opened 2020-11-17
Dimension metadata resources


tpluscode:
Any dimension (column mapping) added to the observation table must have a metadata resource added to the dimension metadata graph (#185)

**DoD**

* [x] when a column mapping is created on **Observation Table**, create a [minimal metadata resource](https://github.com/zazuko/cube-creator/blob/f8cfa737c6cdc017a9591ec9ee5d05110b5ab5f9/fuseki/sample-ubd.trig#L61-L63)
   * the `schema:about` points to the dimension property, to be calculated from mapping' `cc:namespace`+`cc:targetProperty` or concrete URL is `cc:targetProperty` is a named node
   * attach to the `cc:DimensionsMetadata` resource
* [x] when a column mapping is removed, remove the metadata resource
* [x] when a column mapping's `cc:targetProperty` changes, the mapping resource has to be updated
   * `schema:about` point to correct URI
   * the `?dimension={dimension}` resources renamed
* [x] when the CsvMapping's `cc:namespace` changes all dimension mapping resources have to be updated

**Considerations**

The template `/dimensions-metadata?dimension={dimension}` does not necessarily have to be constructed from the full URI. In case of column mappings (in contrast to imported cube) the `cc:targetProperty` could be used alone. For example

```turtle
<project/mapping>
  cc:namespace <http://example.com/cube/> .

<project/table/observation>
  a cc:ColumnMapping ;
  cc:targetProperty "year" .

<project/dimension-metadata?dimension=year>
  schema:about <http://example.com/cube/year> .
```

lucafurrer:
>the ?dimension={dimension} resources renamed
Is this still relevant with the non-query naming?

tpluscode:
Technically not, but unless they are truly unique you would still have to rename using some naming scheme

PR #204, opened 2020-11-19
Feat/create column meta data

lucafurrer:
Create a dimensionMetadata entry whenever a column mapping is created

Part of #187


martinmaillard:
One remark: we only need to create `DimensionMetadata` for column mappings on the `ObservationTable`(s)

apis/core/lib/domain/column-mapping/create.ts
```diff
@@ -39,6 +42,16 @@ export async function createColumnMapping({
     defaultValue: resource.out(cc.defaultValue).term,
   })
 
+  const dimensionMetaDataCollectionPointer = await getDimensionMetaDataCollection(table.csvMapping.id)
+  const dimensionMetaDataCollection = await store.getResource<DimensionMetadataCollection>(dimensionMetaDataCollectionPointer)
+  if (!dimensionMetaDataCollection) {
+    throw new Error(`DimensionMetadataCollection ${dimensionMetaDataCollectionPointer} not found`)
```
tpluscode, line 48: You can simpler
```suggestion
    throw new NotFoundError(`{dimensionMetaDataCollectionPointer`)
```

apis/core/lib/domain/identifiers.ts
```diff
@@ -55,3 +55,7 @@ export function user(user: string): NamedNode {
 export function job(jobCollection: GraphPointer<NamedNode>): NamedNode {
   return $rdf.namedNode(`${jobCollection.value}/${nanoid()}`)
 }
+
+export function dimensionMetadata(dimensionMetadataCollection: DimensionMetadataCollection, mapping: ColumnMapping): NamedNode {
+  return $rdf.namedNode(`${dimensionMetadataCollection.pointer}?dimension=${mapping.pointer}`)
```
tpluscode, line 60: We discussed with @martinmaillard and query params are not such a great choice after all.

Check the updated seed graph in [#200](https://github.com/zazuko/cube-creator/pull/200/files#diff-ee0d6661710ff05549447fce5560bf3fde360e65f8ea766d8a1ab77c1e40ebe9R50-R54)

I propose you simply concatenate the collection ID with the `targetProperty`. Keep the string value unchanged and maybe use `import { shrink } from '@zazuko/rdf-vocabularies'` to shorten URI target properties

apis/core/lib/domain/dimension-metadata/DimensionMetadata.ts
```diff
@@ -0,0 +1,57 @@
...
+  return class Impl extends Resource implements ApiDimensionMetadataCollection {
+    addDimensionMetadata(params: CreateColumnMetadata): DimensionMetadata.DimensionMetadata {
+      const dimensionMetadata = DimensionMetadata.create(this.pointer.node(id.dimensionMetadata(this, params.columnMapping)), {
+        about: createIdentifier(params.columnMapping.targetProperty),
+      })
+
+      this.pointer.addOut(schema.hasPart, dimensionMetadata.id)
+
+      return dimensionMetadata
+    }
```
tpluscode, line 37: Please throw if another metadata resource already exists for the same`schema:about`

apis/core/lib/domain/queries/dimension-metadata.ts
```diff
@@ -0,0 +1,27 @@
...
+        GRAPH ?dataset
+        {
+            ?dataset ${cc.dimensionMetadata} ?dimensionMetadata .
+        }
+        `
+    .LIMIT(1)
+    .execute(client.query)
+  if (results.length === 0) {
+    throw new Error(`No DimensionMetadata for table ${csvMapping} found`)
+  }
```
tpluscode, line 25: Why `LIMIT`?
I would actually throw for anything which is `length !== 1`

apis/core/lib/domain/table/create.ts
```diff
@@ -52,6 +55,12 @@ export async function createTable({
   })
 
   // Create default column mappings for provided columns
+  const dimensionMetaDataCollectionPointer = await getDimensionMetaDataCollection(csvMapping.id)
+  const dimensionMetaDataCollection = await store.getResource<DimensionMetadataCollection>(dimensionMetaDataCollectionPointer)
+  if (!dimensionMetaDataCollection) {
+    throw new Error(`DimensionMetadataCollection ${dimensionMetaDataCollectionPointer} not found`)
```
tpluscode, line 61: Ditto

```suggestion
    throw new NotFoundError(dimensionMetaDataCollectionPointer)
```

packages/model/DimensionMetadata.ts
```diff
@@ -4,14 +4,23 @@ import { schema } from '@tpluscode/rdf-ns-builders'
...
 
+export function DimensionMetadataMixin<Base extends Constructor>(Resource: Base) {
+  class Impl extends Resource implements Partial<DimensionMetadata> {
+    @property.resource({ path: schema.about })
+    about!: NamedNode
+  }
+
+  return Impl
+}
+
```
tpluscode, line 23: re. the CI error, look you will need to annotate the intended target for this mixin. This one does not have a class associated nor any other way to use it implicitly. Maybe try this

```suggestion
export function DimensionMetadataMixin<Base extends Constructor>(Resource: Base) {
  class Impl extends Resource implements Partial<DimensionMetadata> {
    @property.resource({ path: schema.about })
    about!: NamedNode
  }

  return Impl
}

DimensionMetadataMixin.shouldApply = false
```

apis/core/test/domain/column-mapping/create.test.ts
```diff
@@ -104,4 +116,23 @@ describe('domain/column-mapping/create', () => {
...
+
+    // when
+    await createColumnMapping({ resource, store, tableId: table.term, dimensionMetadataQueries })
+
+    // then
+    expect(dimensionMetadata).to.matchShape({
+      property: [{
+        path: schema.hasPart,
+        minCount: 1,
+      }],
```
tpluscode, line 135: You can also use `sh:node` to check deeper into the graph

```suggestion
      property: [{
        path: schema.hasPart,
        minCount: 1,
        node: {
          property: {
            path: schema.about,
            minCount: 1,
            maxCount: 1,
          }
        }
      }],
```

apis/core/test/domain/table/create.test.ts
```diff
@@ -153,9 +164,33 @@ describe('domain/table/create', () => {
...
+
+    // when
+    await createTable({ resource, store, tableCollection, dimensionMetadataQueries })
+
+    // then
+    expect(dimensionMetadata).to.matchShape({
+      property: [{
+        path: schema.hasPart,
+        minCount: 2,
+      }],
```
tpluscode, line 193: Ditto. And btw, `property` does not have to an array

apis/core/lib/domain/dataset/update.ts
```diff
@@ -41,6 +43,7 @@ export async function update({
     .addOut(rdf.type, dcat.Dataset)
 
   hasPart.forEach(child => datasetResource.addOut(schema.hasPart, child))
+  dimensionMetadata.forEach(child => datasetResource.addOut(cc.dimensionMetadata, child))
```
tpluscode, line 46: You don't need to modify this file. The dataset resource is in a different graph from the metadata

tpluscode (changes requested):
You can now rebase master. The `createIdentifier` got merged

apis/core/lib/domain/dataset/update.ts
lucafurrer: But the triple `dataset cc.dimensionMetadata metadatacollection` has been deleted by the deleteOut by an update in the end to end tests.

tpluscode: Aha, my mistake. I mistook these for the `schema:hasPart` triples. Sorry

lucafurrer:
I think I addressed all the points

packages/model/DimensionMetadata.ts
```diff
@@ -42,3 +42,9 @@ DimensionMetadataCollectionMixin.appliesTo = cc.DimensionMetadataCollection
 export const createCollection = initializer<DimensionMetadataCollection>(DimensionMetadataCollectionMixin, {
   types: [cc.DimensionMetadataCollection],
 })
+
+DimensionMetadataCollectionMixin.appliesTo = cc.DimensionMetadataCollection
```
tpluscode, line 46: This got duplicated in merge

apis/core/lib/domain/dimension-metadata/DimensionMetadata.ts
```diff
@@ -0,0 +1,46 @@
+import { Constructor } from '@tpluscode/rdfine'
```
tpluscode, line 1: Uh, oh., I created the same model in `apis/core/lib/domain/dimension` directory. Sorry 😊 

Please merge `dimension` and `dimension-metadata`. Flip a coin to decide on the name :)

apis/core/lib/domain/dimension-metadata/DimensionMetadata.ts
```diff
@@ -0,0 +1,46 @@
...
+  interface DimensionMetadataCollection extends ApiDimensionMetadataCollection {
+  }
+}
+
+export default function Mixin<Base extends Constructor<DimensionMetadataCollection>>(Resource: Base) {
+  return class Impl extends Resource implements ApiDimensionMetadataCollection {
+    addDimensionMetadata(params: CreateColumnMetadata): DimensionMetadata.DimensionMetadata {
+      const identifier = params.csvMapping.createIdentifier(params.columnMapping.targetProperty)
+
+      if (this.hasPart && this.hasPart.some((dimMeta) => dimMeta.about === identifier)) {
```
tpluscode, line 31: As discussed on zulip, checking `this.hasPart` for truthiness will not be necessary

apis/core/lib/domain/dimension-metadata/DimensionMetadata.ts
```diff
@@ -0,0 +1,46 @@
...
+        throw new Error(`Dimension Metadata with identifier ${identifier.value} already exists`)
+      }
+
+      const dimensionMetadata = DimensionMetadata.create(this.pointer.node(id.dimensionMetadata(this, params.columnMapping)), {
```
tpluscode, line 35: I've drifted a little away from really needing the `identifiers` module. In this case for example, `id.dimensionMetadata` will likely only ever be used here. 

I propose to move into this module instead

apis/core/lib/domain/identifiers.ts
```diff
@@ -55,3 +56,12 @@ export function user(user: string): NamedNode {
...
 }
+
+export function dimensionMetadata(dimensionMetadataCollection: DimensionMetadataCollection, mapping: ColumnMapping): NamedNode {
+  if (mapping.targetProperty.termType === 'Literal') {
+    return $rdf.namedNode(`${dimensionMetadataCollection.id.value}/${mapping.targetProperty.value}`)
+  } else {
+    const prefixedProperty = shrink(mapping.targetProperty.value) || mapping.targetProperty.value
+    return $rdf.namedNode(`${dimensionMetadataCollection.id.value}/${prefixedProperty}`)
+  }
+}
```
tpluscode, line 67: 1. May be slightly more readable if there was only one `return`?
2. Please `urlencode` the URI if did not shrink

tpluscode (changes requested):

tpluscode (approved):

PR #209, opened 2020-11-19
Initial version of cube designer

martinmaillard:


packages/model/DimensionMetadata.ts
```diff
@@ -19,6 +20,9 @@ function DimensionMetadataMixin<Base extends Constructor>(Resource: Base) {
   class Impl extends Resource implements DimensionMetadata {
     @property()
     about!: NamedNode
+
+    @property({ values: 'array' })
+    name?: Literal[]
```
tpluscode, line 25: This will never be undefined. No triples = empty array
```suggestion
    name: Literal[]
```

packages/model/DimensionMetadata.ts
```diff
@@ -1,13 +1,14 @@
...
-import { NamedNode } from 'rdf-js'
+import { Literal, NamedNode } from 'rdf-js'
 import { schema } from '@tpluscode/rdf-ns-builders'
 import { cc } from '@cube-creator/core/namespace'
 import { initializer } from './lib/initializer'
 
-interface DimensionMetadata extends RdfResourceCore {
+export interface DimensionMetadata extends RdfResourceCore {
   about?: NamedNode
+  name?: Literal[]
```
tpluscode, line 11:
```suggestion
  name: Literal[]
```

tpluscode (commented):
Other than a minor nitpick this works quite nice.

I only noticed that the data `Missing cube title` text does not update when the dataset is edited. Not when form is said, not when the language is changed

martinmaillard:
Good catch, that's fixed

tpluscode (approved):
LGTM

PR #210, opened 2020-11-19
feat: prevent same property mapped twice

### LINKED ISSUE #189: Enforce unique `cc:targetProperty` for a table
Values of `cc:targetProperty` map to observation dimensions (properties). It makes no sense for two columns to map to the same dimension.

While it's not necessarily a strict requirement for the Csv Mapping resource itself, it will be easier to manage dimension metadata and prevent invalid observation sets if we forbid duplicate column mappings

**DoD**

* [ ] column mapping is rejected if another has the same `cc:targetProperty`
   * it can be both a string and URI value (needs to be checked by the effective URI)

**Examples**

Given column mappings

```turtle
<mapping> cc:namespace <http://example.com/cube/> .

<table/column/foo> cc:targetProperty "year" .
<table/column/bar> cc:targetProperty schema:age .
<table/column/baz> cc:targetProperty <http://example.com/cube/salary> .
```

These will be invalid

```turtle
# obvious duplicates
[] cc:targetProperty "year" .
[] cc:targetProperty schema:age .
[] cc:targetProperty <http://example.com/cube/salary> .

# string will map to explicit constructed property
[] cc:targetProperty "salary" .
```

tpluscode:
fixes #189

### DIFF
--- a/apis/core/lib/errors/domain.ts
+++ b/apis/core/lib/errors/domain.ts
@@ -1,0 +1,22 @@
+import { ErrorMapper } from 'http-problem-details-mapper'
+import { ProblemDocument } from 'http-problem-details'
+
+export class DomainError extends Error {
+  // eslint-disable-next-line @typescript-eslint/no-useless-constructor
+  constructor(message: string) {
+    super(message)
+  }
+}
+
+export class DomainErrorMapper extends ErrorMapper {
+  public constructor() {
+    super(DomainError)
+  }
+
+  mapError(error: Error): ProblemDocument {
+    return new ProblemDocument({
+      status: 400,
+      detail: error.message,
+    })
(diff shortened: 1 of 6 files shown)


apis/core/lib/errors/domain.ts
```diff
@@ -0,0 +1,22 @@
+import { ErrorMapper } from 'http-problem-details-mapper'
+import { ProblemDocument } from 'http-problem-details'
+
+export class DomainError extends Error {
+  // eslint-disable-next-line @typescript-eslint/no-useless-constructor
```
tpluscode, line 5: Not really useless because in `Error` it is optional

martinmaillard (approved):

PR #211, opened 2020-11-19
fix: only allow certain characters in target property

### LINKED ISSUE #197: If I put a space in the `targetProperty` of a column mapping, it crashes the pipeline. Not sure if we should prevent that or just escape spaces.
If I put a space in the `targetProperty` of a column mapping, it crashes the pipeline. Not sure if we should prevent that or just escape spaces.

_Originally posted by @martinmaillard in https://github.com/zazuko/cube-creator/issues/169#issuecomment-728216684_

tpluscode:
fixes #197


martinmaillard:
I think we also need to change this https://github.com/zazuko/cube-creator/blob/master/apis/core/lib/domain/table/Table.ts#L37

tpluscode:
> I think we also need to change this

Done

martinmaillard (approved):

PR #213, opened 2020-11-20
build: prevent bundling of all vocabularies

tpluscode:
Importing from `@zazuko/rdf-vocabularies/datasets` causes the entire 40 odd megabytes being precessed by the build.

Instead, we must only fetch the actually necessary modules


martinmaillard (approved):

lucafurrer (approved):
Whenever we add a vocabulary we need to add it on both places, correct?

tpluscode:
Good thinking @lucafurrer. The import does not really have to be destructured anymore

PR #219, opened 2020-11-23
feat: delete columnmapping

vhf:


apis/core/lib/domain/column-mapping/delete.ts
```diff
@@ -0,0 +1,57 @@
...
+import { ResourceStore } from '../../ResourceStore'
+import { resourceStore } from '../resources'
+import { NamedNode } from 'rdf-js'
+import { ColumnMapping, CsvMapping, DimensionMetadataCollection, Table } from '@cube-creator/model'
+import * as DimensionMetadataQueries from '../queries/dimension-metadata'
+import * as TableQueries from '../queries/table'
+import { NotFoundError } from '../../errors'
+
+interface DeleteColumnMappingCommand {
+  resource: GraphPointer
```
martinmaillard, line 12: I'm not sure you need a `GraphPointer` since you're only using `.term`.

apis/core/lib/domain/column-mapping/delete.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    const dimensionMetaDataCollectionPointer = await getDimensionMetaDataCollection(table.csvMapping.id)
+    const dimensionMetaDataCollection = await store.getResource<DimensionMetadataCollection>(dimensionMetaDataCollectionPointer)
+    if (!dimensionMetaDataCollection) {
+      throw new NotFoundError(dimensionMetaDataCollectionPointer)
+    }
+
+    const dimension = dimensionMetaDataCollection.findDimension({ csvMapping, targetProperty: columnMapping.targetProperty })
+    if (dimension) {
+      store.delete(dimension.about)
```
martinmaillard, line 51: I think it would make sense to completely delete the dimension

apis/core/test/domain/column-mapping/delete.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      getTablesForMapping,
+      getTableForColumnMapping,
+    }
+  })
+
+  it('deletes a column mapping', async () => {
+    const resource = $rdf.namedNode('columnMapping')
+    const columnMapping = deleteColumnMapping({ resource, store, dimensionMetadataQueries, tableQueries })
+
+    expect(columnMapping).to.be.fulfilled
```
martinmaillard, line 118: It would be nice to check that the resource, and potential related dimension, are actually deleted

apis/core/lib/domain/column-mapping/delete.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const tableId = await getTableForColumnMapping(columnMapping.id)
+  const table = await store.getResource<Table>(tableId)
+
+  if (!table) {
+    throw new NotFoundError(tableId)
+  }
+
+  if (table.types.has(cc.ObservationTable)) {
```
martinmaillard, line 36:
```suggestion
  if (table.isObservationTable) {
```

apis/core/lib/domain/dimension/DimensionMetadataCollection.ts
```diff
@@ -46,6 +47,16 @@ export default function Mixin<Base extends Constructor<Omit<DimensionMetadataCol
       }
     }
 
+    deleteDimension(dimension: DimensionMetadata.DimensionMetadata): void {
+      const found = this.find(dimension.pointer.out(schema.about).term!)
+      if (!found) {
+        throw new Error('Dimension not found')
+      }
+      found.pointer.deleteOut()
```
martinmaillard, line 55: Since we already have the `dimension`, I don't think it's useful to `.find` it again.

```suggestion
      dimension.pointer.deleteOut()
```

The other solution is to take a `NamedNode` in this method and to not `.find` the dimension it beforehand.

martinmaillard:
I think the `<table> cc:columnMapping <the-column-mapping>` link is not properly deleted. I still get it from the API after deleting the column mapping.

apis/core/lib/domain/column-mapping/delete.ts
```diff
@@ -32,6 +33,8 @@ export async function deleteColumnMapping({
     throw new NotFoundError(tableId)
   }
 
+  table.pointer.deleteOut(cc.columnMapping)
```
martinmaillard, line 36: This will unlink all the column mappings from the table

vhf: Fixed, I don't think there's a better way with clownface…

martinmaillard (approved):
🚀

PR #224, opened 2020-11-24
feat: observations endpoint

### LINKED ISSUE #216: API endpoint to get Cube observations
For the cube preview, we need a way to get the cube observations. This endpoint should also support filters.

tpluscode:
This adds a link to cubes in the `/dataset` resource which will fetch observations. Filters included 🎉 

closes #216 

**Details**

`observations` property to the Cube is a template. Populate it with turtle string carrying [`view:view` filters](https://github.com/zazuko/rdf-cube-view-schema/blob/master/view.ttl#L89-L97)

Compared to the linked example the filters will not need the `rdf:type` and `view:dimension` would be the cube dimension predicate.

The actual link to `observations` is obtained from the link like `dataset.cubes[0].observations.expand(params)` where `params` would be a graph to put into the template. An optional `hydra:limit` is also possible to control the number of results to return (no paging yet, though). Default page size = `20`

```
[
  view:view "turtle as string as mentioned above" ;
  hydra:limit 10
]
```

**Less related changes**

* Added minimal TS type declarations for `rdf-cube-view-query`
* I had to customise the `toCubeShape` step and so I coped their code and remove the `ssz-data` git dependency


martinmaillard:
I get a 400 before I run the pipeline for the first time. Is that expected?

martinmaillard:
I get a 400 `Missing 'cube' query parameter` today

martinmaillard (approved):

PR #214, opened 2020-11-20
fix: only add selected columns to id-template

lucafurrer:
Create identifier template only from selected columns

PR #232, opened 2020-11-25
Fix/default id template 2

lucafurrer:
I screw up the rebase on #214 

Here is a clean version

### DIFF
--- a/apis/core/lib/domain/table/create.ts
+++ b/apis/core/lib/domain/table/create.ts
@@ -46,16 +46,14 @@ export async function createTable({
 
   const columns = [...csvSource.columns]
 
-  const columnsToAdd : CsvColumn[] = []
-
-  resource.out(csvw.column)
-    .forEach(({ term: columnId }) => {
+  const columnsToAdd = resource.out(csvw.column)
+    .map(({ term: columnId }) => {
       const column = columns
         .find(({ id }) => id.equals(columnId))
       if (!column) {
         throw new Error(`Column ${columnId} not found`)
       }
-      columnsToAdd.push(column)
+      return column
     })
 
   const table = await csvMapping.addTable(store, {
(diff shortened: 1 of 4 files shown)


apis/core/lib/domain/table/create.ts
```diff
@@ -46,10 +46,22 @@ export async function createTable({
...
+  resource.out(csvw.column)
+    .forEach(({ term: columnId }) => {
+      const column = columns
+        .find(({ id }) => id.equals(columnId))
+      if (!column) {
+        throw new Error(`Column ${columnId} not found`)
+      }
+      columnsToAdd.push(column)
+    })
+
```
martinmaillard, line 60: If you use `.map` instead of `.forEach`, you can avoid this weird "build an empty list and then populate it" dance.

Something like:
```typescript
  const columnsToAdd: CsvColumn[] = resource
    .out(csvw.column)
    .map(({ term: columnId }) => {
      const column = columns
        .find(({ id }) => id.equals(columnId))
      if (!column) {
        throw new Error(`Column ${columnId} not found`)
      }
      return column
    })
```

lucafurrer: Nice, why have I not thought of this

martinmaillard (approved):
This looks good to me as a first step.

Ultimately, I think it would be better to generate the auto-identifier template on the fly because with this implementation, adding a column mapping doesn't update the identifier template. And we would only want to update it if it was not provided manually.

lucafurrer:
> This looks good to me as a first step.
> 
> Ultimately, I think it would be better to generate the auto-identifier template on the fly because with this implementation, adding a column mapping doesn't update the identifier template. And we would only want to update it if it was not provided manually.

I agree, that this should be the end goal, but this will require quite some business logic and until then, we have at least something that works as intended

PR #223, opened 2020-11-24
feat: update column mapping

lucafurrer:
Update a column mapping

fix #118

PR #237, opened 2020-11-25
Links between tables

martinmaillard:
TODO:
- [x] Store identifierMapping
- [x] It works when creating. Make it work when updating.
- [x] Fix typescript issues with new ColumnMapping
- [x] tests
- [x] Adapt csvw generation

Depends on #223


apis/core/lib/domain/table/Table.ts
```diff
@@ -55,8 +66,26 @@ export default function mixin<Base extends Constructor<Table>>(Resource: Base):
       return columnMapping
     }
 
-    addColumnMappingFromColumn({ store, column }: CreateColumnMappingFromColumn): ColumnMapping.ColumnMapping {
-      return this.addColumnMapping({
+    addReferenceColumnMapping({ store, targetProperty, referencedTable, identifierMappings }: CreateReferenceColumnMapping): ColumnMapping.ReferenceColumnMapping {
+      const columnMapping = ColumnMapping.createReference(store.create(id.columnMapping(this, targetProperty.value)), {
```
martinmaillard, line 70: I'm not sure about using the target property to build the identifier.

apis/core/bootstrap/shapes/column-mapping.ts
```diff
@@ -66,12 +67,77 @@ ${shape('column-mapping/create')} {
...
+    ${sh.property} [
+      ${sh.name} "Identifier mapping" ;
+      ${sh.path} ${cc.identifierMapping} ;
+      ${sh.node} ${$rdf.blankNode('identifier-mapping')} ;
+      ${dash.editor} ${dash.DetailsEditor} ;
+      ${sh.order} 30 ;
+    ] ;
+  .
+
+  ${$rdf.blankNode('identifier-mapping')} a ${sh.NodeShape} ;
```
tpluscode, line 118: This is what stardog 6 hates. In case will make it wok after all

```suggestion
      ${sh.node} ${$rdf.blankNode('identifierMapping')} ;
      ${dash.editor} ${dash.DetailsEditor} ;
      ${sh.order} 30 ;
    ] ;
  .

  ${$rdf.blankNode('identifierMapping')} a ${sh.NodeShape} ;
```

apis/core/hydra/column-mapping.ttl
```diff
@@ -5,33 +5,70 @@ prefix code:  <https://code.described.at/>
...
+      "DELETE" ;
+    code:implementedBy [
+      a
+        code:EcmaScript ;
+      hydra:expects
+        cc:ColumnMapping ;
+      code:link
+        <file:handlers/column-mapping#remove> ;
+    ]
+  ] .
```
tpluscode, line 44: I think we could have only one DELETE operation supported by a `cc:ColumnMapping` - superclass to the other two

apis/core/lib/csvw-builder/index.ts
```diff
@@ -47,6 +48,54 @@ function mappedColumn({ csvMapping, columnMapping, column }: CsvwBuildingContext
...
 }
 
+async function mappedReferenceColumn({ csvMapping, columnMapping, resources }: CsvwBuildingContext & { columnMapping: ReferenceColumnMapping }): Promise<Initializer<Csvw.Column>> {
+  const csvwColumn: Initializer<Csvw.Column> = {
+    propertyUrl: csvMapping.createIdentifier(columnMapping.targetProperty).value,
+  }
+
+  const identifierMappings = columnMapping.identifierMapping
+  const columnNameMap = await identifierMappings.reduce<Promise<Map<string, string>>>(async (mapP, mapping) => {
+    const map = await mapP
```
tpluscode, line 58: `mapP`?

apis/core/lib/csvw-builder/index.ts
```diff
@@ -30,7 +31,7 @@ interface CsvwBuildingContext {
   resources: ResourceStore
 }
 
-function mappedColumn({ csvMapping, columnMapping, column }: CsvwBuildingContext & { column: CsvColumn; columnMapping: ColumnMapping }): Initializer<Csvw.Column> {
+function mappedLiteralColumn({ csvMapping, columnMapping, column }: CsvwBuildingContext & { column: CsvColumn; columnMapping: LiteralColumnMapping }): Initializer<Csvw.Column> {
```
tpluscode, line 34: Maybe this file is getting big enough to split the `mappedLiteralColumn` and `mappedReferenceColumn` into their own modules?

apis/core/lib/domain/column-mapping/create.ts
```diff
@@ -73,3 +62,71 @@ export async function createColumnMapping({
...
 }
+
+interface CreateLiteralColumnMappingCommand {
+  targetProperty: Term
+  table: Table
+  resource: GraphPointer
+  store: ResourceStore
+}
+
+async function createLiteralColumnMapping({ targetProperty, table, resource, store }: CreateLiteralColumnMappingCommand): Promise<LiteralColumnMapping> {
```
tpluscode, line 73: Here also I would consider moving these two functions in separate module or modules

apis/core/lib/domain/table/Table.ts
tpluscode: You expect it would cause problems?

apis/core/test/csvw-builder/index.test.ts
```diff
@@ -198,12 +198,12 @@ describe('lib/csvw-builder', () => {
...
       CsvColumn.fromPointer(csvSource.pointer.namedNode('jahr-column'), { name: 'JAHR' }),
     ]
-    const columnMapping = ColumnMapping.fromPointer(clownface({ dataset: $rdf.dataset() }).namedNode('year-mapping'), {
+    const columnMapping = ColumnMapping.literalFromPointer(clownface({ dataset: $rdf.dataset() }).namedNode('year-mapping'), {
       sourceColumn: $rdf.namedNode('jahr-column') as any,
       targetProperty: schema.yearBuilt,
       language: 'de-DE',
     })
-    resources.push(columnMapping.pointer)
+    resources.push(columnMapping.pointer as any)
```
tpluscode, line 206: With the change in `TestResourceStore` all those `as any` would be unnecessary?

apis/core/lib/domain/table/Table.ts
martinmaillard: I don't. But then we may as well use `targetProperty` for all column mappings?

apis/core/lib/csvw-builder/index.ts
martinmaillard: "P" is for "Promise". I had a lot of issues making Typescript happy with the async reduce... it's probably improvable

apis/core/lib/csvw-builder/index.ts
martinmaillard: I really prefer long modules than modules with one function in general, but it's up to you... you're kind of the owner of the this package ;)

apis/core/test/csvw-builder/index.test.ts
martinmaillard: Yeah, it was yelling at me and I couldn't figure out how to fix it

PR #248, opened 2020-11-27
fix: remove datatype for frequency

### LINKED ISSUE #242: Cube metadata: Data refresh interval not accepted
```
The request payload does not conform to the SHACL description of this endpoint.

http://purl.org/dc/terms/accrualPeriodicity
Value does not have datatype <http://purl.org/dc/terms/PeriodOfTime>
```
Applies to "annual" and all other options I've tested

lucafurrer:
Since the list of elements if provided by the in clause, I think the data type is redundant and can be omitted instead of adding all the information that the type can correctly be resolved.

fixes #242


martinmaillard (approved):

martinmaillard:
I think `sh:datatype` was not correct anyway. It should have been `sh:class`.

PR #258, opened 2020-11-30
feat: middleware to disambiguate operations

tpluscode:
Waiting for zazuko/hydra-box#86 to be released as package.

This PR adds a middleware which removes operation ambiguity, allowing for example multiple `POST` operations differing in the expected class.

```turtle
cc:Table a hydra:Class ;
  hydra:supportedOperation [
    hydra:expects cc:LiteralColumn ;
    hydra:method "POST"
  ] , [
    hydra:expects cc:ReferenceColumn ;
    hydra:method "POST"
  ]
.

cc:LiteralColumn a hydra:Class .
cc:ReferenceColumn a hydra:Class .
```

When POSTing to the table a resource like

```
<> a cc: ReferenceColumn .
```

The latter operation will be selected where currently it will fail, because both are found based on the HTTP method alone.

**Nitty gritty**

The types must be explicitly typed as `hydra:Class`. Anything else (like a `sh:Shape`) will be ignored.


martinmaillard (approved):
🎉

lucafurrer (approved):

apis/core/index.ts
```diff
@@ -50,6 +51,12 @@ async function main() {
       password: env.maybe.STORE_ENDPOINTS_PASSWORD,
     },
     errorMappers,
+    middleware: {
+      operations: [
+        preferHydraCollection,
```
tpluscode, line 56: This one is to select the collection operation in case when resource is `<> a hydra:Collection , hydra:Resource`. Causes ambiguity with the updated hydra-box/labyrinth

apis/core/index.ts
```diff
@@ -50,6 +51,12 @@ async function main() {
...
+        expectsDisambiguate,
```
tpluscode, line 57: Discards operation which do not match the payload's rdf:type

apis/core/lib/middleware/operations.ts
```diff
@@ -0,0 +1,41 @@
...
+      .has(rdf.type, hydra.Class)
+      .terms
+
+    return new Set(expectedClasses.filter(type => !resourceTypes.has(type))).size === 0
+  })
+  next()
+})
+
+export const preferHydraCollection: RequestHandler = (req, res, next) => {
+  if (req.hydra.operations.length < 2) {
```
martinmaillard, line 27: Is there a reason for using `<= 1` above and `< 2` here?

tpluscode: Certainly not. The top one is how webstorm did a "flip `if`" refactoring ;)

martinmaillard (approved):

PR #259, opened 2020-11-30
feat: gitlab trigger

lucafurrer:
Code to trigger gitlab

### DIFF
--- a/apis/core/lib/pipeline/trigger.ts
+++ b/apis/core/lib/pipeline/trigger.ts
@@ -33,9 +33,19 @@ export const local = trigger(job => {
   }
 })
 
-export async function gitlab(job: NamedNode): Promise<void> {
-  throw new Error('Not implemented')
-}
+export const gitlab = trigger(job => {
+  const form = new URLSearchParams({
+    token: env.PIPELINE_TOKEN,
+    ref: env.maybe.PIPELINE_BRANCH || 'master',
+    'variables[JOB]': job.value,
+    'variables[ENV]': env.PIPELINE_ENV,
+  })
+
+  return {
+    method: 'POST',
+    body: form,
+  }
+})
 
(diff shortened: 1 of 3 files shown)


apis/core/lib/pipeline/trigger.ts
```diff
@@ -33,9 +33,19 @@ export const local = trigger(job => {
   }
 })
 
-export async function gitlab(job: NamedNode): Promise<void> {
-  throw new Error('Not implemented')
-}
+export const gitlab = trigger(job => {
+  const form = new URLSearchParams({
+    token: env.PIPELINE_TOKEN,
+    ref: env.maybe.PIPELINE_BRANCH || 'master',
```
tpluscode, line 39: I don't think we want this here. There will probably be only one branch on GitLab
```suggestion
    ref: 'master',
```

packages/core/env.ts
```diff
@@ -40,6 +40,9 @@ type ENV_VARS =
   | 'AWS_SECRET_ACCESS_KEY'
   | 'PIPELINE_TYPE'
   | 'PIPELINE_URI'
+  | 'PIPELINE_TOKEN'
+  | 'PIPELINE_BRANCH'
```
tpluscode, line 44:
```suggestion
```

apis/core/lib/pipeline/trigger.ts
lucafurrer: If you think so

tpluscode: On GitHub the container executes directly from sources. Thus, having the branch option is good to be able to run different code version.

GitLab runs from pre-built containers. So either we could make the branch a trigger variable. Otherwise, if the job start command changes, we can differentiate that in the gitlab yaml without creating branches

PR #266, opened 2020-12-02
fix: dup. column mapping property check on update

martinmaillard:
Fix duplicate target property column mapping check on update

### DIFF
--- a/apis/core/lib/domain/column-mapping/update.ts
+++ b/apis/core/lib/domain/column-mapping/update.ts
@@ -50,10 +50,12 @@ export async function updateColumnMapping({
   }
 
   if (targetProperty.value !== columnMapping.targetProperty.value) {
-    const mappingExists = await findMapping(table, targetProperty, store)
-    if (mappingExists) {
-      const column = await store.getResource<CsvColumn>(mappingExists.sourceColumn?.id)
-      throw new DomainError(`Target property already mapped from column ${column?.name}`)
+    if (table.isObservationTable) {
+      const mappingExists = await findMapping(table, targetProperty, store)
+      if (mappingExists) {
+        const column = await store.getResource<CsvColumn>(mappingExists.sourceColumn?.id)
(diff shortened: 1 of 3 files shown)


apis/core/lib/domain/column-mapping/update.ts
```diff
@@ -50,10 +50,12 @@ export async function updateColumnMapping({
...
-      throw new DomainError(`Target property already mapped from column ${column?.name}`)
+    if (table.isObservationTable) {
+      const mappingExists = await findMapping(table, targetProperty, store)
+      if (mappingExists) {
+        const column = await store.getResource<CsvColumn>(mappingExists.sourceColumn?.id)
+        throw new DomainError(`Target property already mapped from column ${column?.name}`)
+      }
     }
 
     if (table.types.has(cc.ObservationTable)) {
```
lucafurrer, line 61: If we check for the observation table, I would merge both checks for the observation tables , no?
```suggestion
    if (table.isObservationTable) {
      const mappingExists = await findMapping(table, targetProperty, store)
      if (mappingExists) {
        const column = await store.getResource<CsvColumn>(mappingExists.sourceColumn?.id)
        throw new DomainError(`Target property already mapped from column ${column?.name}`)
      }
```

lucafurrer (commented):
We do not need unicity for the targetProperty outside observation tables?

martinmaillard:
Judging by the discussions, I'm not sure we need unicity at all. But here I'm just replicating what was implemented in "createColumnMapping".

apis/core/lib/domain/column-mapping/update.ts
martinmaillard: Be my guest ;) I don't have much time to work on this. I just noticed that the behavior was not consistent between "create" and "update" and quickly wanted to fix it

lucafurrer: Here you go :wink:

PR #269, opened 2020-12-02
feat: publish cube

lucafurrer:
Publish a cube to lindas, CLI Part


apis/core/test/domain/job/create.test.ts
```diff
@@ -31,11 +31,11 @@ describe('domain/job/create', () => {
     ])
   })
 
-  it('creates a job', async () => {
+  it('creates a transformation job', async () => {
```
tpluscode, line 34: Might also add a `expect` to verify `rdf:type` as in the publish job case below

cli/lib/commands/publish.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    log.enabled = debug
+    const authConfig = {
+      params: command.authParam,
+    }
+
+    const pipelinePath = (filename: string) => path.join(basePath, `./pipelines/${filename}.ttl`)
+    const dataset = $rdf.dataset()
+      .merge(await fileToDataset('text/turtle', pipelinePath('publish')))
+      // .merge(await fileToDataset('text/turtle', pipelinePath(`to-${to}`)))
```
tpluscode, line 54:
```suggestion
```

cli/test/lib/commands/publish.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    )
+
+    // then
+    const expectedGraph = ns.targetCube()
+    const dataset = await $rdf.dataset().import(await CONSTRUCT`?s ?p ?o`
+      .FROM(expectedGraph)
+      .WHERE`?s ?p ?o`
+      .execute(client.query))
+
+    const cubePointer = clownface({ dataset })
```
tpluscode, line 51: Please wrap the test setup code in an additional `describe+before` and make put the expectations in their own `it` blocks

```suggestion
  describe('produces triples', async function () {
    // given
    const runner = publish($rdf.namedNode('urn:pipeline:cube-creator#Main'), debug('test'))

    // when
    await runner({
      debug: true,
      job: `${env.API_CORE_BASE}cube-project/ubd/csv-mapping/jobs/test-publish-job`,
      executionUrl,
    })
    await new Promise((resolve) =>
      setTimeout(resolve, 100),
    )

    // then
    const expectedGraph = ns.targetCube()
    const dataset = await $rdf.dataset().import(await CONSTRUCT`?s ?p ?o`
      .FROM(expectedGraph)
      .WHERE`?s ?p ?o`
      .execute(client.query))

    const cubePointer = clownface({ dataset })
```

cli/test/lib/commands/publish.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    expect(cubePointer.has(rdf.type, sh.NodeShape).terms.length).to.eq(1)
```
tpluscode, line 52: Might actually add an explicit target to check tf the shape is what we expect and not only that "a shape" exists


```suggestion
    expect(cubePointer.has(rdf.type, sh.NodeShape).term).to.deep.eq(ns.targetCube('shape-id'))
```

cli/test/lib/commands/publish.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    expect(cubePointer).to.matchShape({
```
tpluscode, line 54: This will not do any validation. You will have to select a specific node from the pointer like you did below

```suggestion
    expect(cubePointer.node(/* ... */)).to.matchShape({
```

.github/workflows/pipeline.yaml
```diff
@@ -20,13 +21,22 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1.4.4
         with:
           node-version: 14
       - run: yarn install --ci
-      - run: yarn ts-node index.ts transform --job "${{ github.event.inputs.job }}" --execution-url "https://${{ github.repository }}/actions/runs/${{ github.run_id }}" --to ${{ github.event.inputs.to }} --debug
+      - name: transform pipeline
+        if: github.event.inputs.transform_job != ''
+        run: yarn ts-node index.ts transform --job "${{ github.event.inputs.transform_job }}" --execution-url "https://${{ github.repository }}/actions/runs/${{ github.run_id }}" --to ${{ github.event.inputs.to }} --debug
```
tpluscode, line 35: No biggie but you might extract the `https://${{ github.repository }}/actions/runs/${{ github.run_id }}` into the `env`section. These command are already long enough

apis/core/lib/pipeline/trigger.ts
```diff
@@ -37,24 +43,39 @@ export const gitlab = trigger(job => {
...
+        transform_job: job.value,
+      },
+    }
+  } else {
+    body = {
+      ref: 'master',
+      inputs: {
+        publish_job: job.value,
+      },
+    }
```
tpluscode, line 77: Slightly more readable IMO

```suggestion
  const jobParam = job.has(rdf.type, cc.TransformJob).values.length > 0 ? 'transform_job' : 'publish_job'

    body = {
      ref: 'master',
      inputs: {
        [jobParam]: job.value,
      }
    }
```

cli/server.ts
```diff
@@ -38,17 +39,24 @@ async function main() {
 
   app.get('/', (req, res) => res.status(204).end())
 
-  const tranform = transform.default(pipelines.TransformFiles, log)
+  const doTransform = transform.default(pipelines.TransformFiles, log)
+  const doPublish = publish.default(pipelines.TransformFiles, log)
```
tpluscode, line 43: is it in fact `pipelines.TransformFiles`?

ui/src/store/serializers.ts
```diff
@@ -83,8 +83,13 @@ export function serializeDimensionMetadata (dimension: DimensionMetadata): Dimen
...
   return Object.freeze({
     ...serializeResource(collection),
+    actions: {
+      ...serializeActions(collection.actions),
+      createTransform: collection.actions.createTransform,
+      createPublish: collection.actions.createPublish,
+    },
     member: collection.member.map(Object.freeze),
-  }) as JobCollection
+  }) as unknown as JobCollection
```
tpluscode, line 92: Why was `unknown` needed here now?

ui/src/components/JobForm.vue
```diff
@@ -41,10 +42,6 @@ export default class JobForm extends Vue {
     this.shape = await api.fetchOperationShape(this.operation)
   }
 
-  get hasNoOptions (): boolean {
-    return this.shape?.property.length === 0
-  }
-
```
tpluscode, line 47: Is this no longer relevant?

tpluscode (changes requested):

ui/src/store/serializers.ts
lucafurrer: @martinmaillard I let you answer this :wink:

ui/src/components/JobForm.vue
lucafurrer: @martinmaillard this as well

apis/core/lib/pipeline/trigger.ts
lucafurrer: :1st_place_medal:

.github/workflows/pipeline.yaml
lucafurrer: Like this https://github.com/zazuko/cube-creator/pull/269/commits/4d92b392eebf7fb28fac40d647c6f7fc410e9064 ?

tpluscode: Well, almost. Still gotta pass `--execution-url` to the commands :)

ui/src/store/serializers.ts
martinmaillard: It was telling me that the intersection between both types was not sufficient to just use `as ...` (because of all the missing properties of Collection).

ui/src/components/JobForm.vue
martinmaillard: nope

.github/workflows/pipeline.yaml
```diff
@@ -20,13 +21,23 @@ jobs:
       GRAPH_STORE_ENDPOINT: ${{ secrets.GRAPH_STORE_ENDPOINT }}
       GRAPH_STORE_USER: ${{ secrets.GRAPH_STORE_USER }}
       GRAPH_STORE_PASSWORD: ${{ secrets.GRAPH_STORE_PASSWORD }}
+      PUBLISH_GRAPH_STORE_ENDPOINT: ${{ secrets.GRAPH_STORE_ENDPOINT }}
```
martinmaillard, line 24: Isn't it going to be an issue to use the same variable names for both jobs?

martinmaillard (approved):

.github/workflows/pipeline.yaml
lucafurrer: As long as we store the graph into the same graph store, as we want to do in test, as far as i know, this should not be a problem. The cube ends up in the cube namespace which is different from everything in the in cube project.
Since we will use gitlab for integ and production, I let it like this.

cli/lib/metadata.ts
```diff
@@ -31,11 +32,13 @@ export async function injectMetadata(jobUri: string) {
...
 
+  const datasetTriples = dataset.pointer.dataset.match(null, null, null, dataset.id)
+
   return obj(function (quad: Quad, _, callback) {
-    const visited = new Set()
+    const visited = new TermSet()
     const copyChildren = (subject: QuadObject) => {
       if (subject && subject.termType !== 'Literal' && !visited.has(subject)) {
-        [...dataset.pointer.dataset.match(subject)].forEach(item => {
+        [...datasetTriples.match(subject, null, null, dataset.id)].forEach(item => {
```
tpluscode, line 41: Not necessary since you filter out the correct triples above

```suggestion
        [...datasetTriples.match(subject)].forEach(item => {
```

PR #273, opened 2020-12-03
fix: add missing "." in shape

martinmaillard:


tpluscode:
The triples are correct. Superseded by #278

PR #278, opened 2020-12-04
fix: use only alpha characters in blank nodes

tpluscode:


lucafurrer (approved):
Is there a problem with non alpha characters?

tpluscode:
Stardog 6 does not parse them correctly from `n-triples`, which is how the resources are bootstrapped with graph store PUT

PR #281, opened 2020-12-04
refactor: throw on not found resources by default

tpluscode:
This changes the default of `ResourceStore` to throw when a resource is not found. A second parameter can be added where an `undefined` is acceptable. Types adjust automatically which removes the need to the repetitive `if (!resource) throw new Error()` boilerplate

### DIFF
--- a/apis/core/lib/ResourceStore.ts
+++ b/apis/core/lib/ResourceStore.ts
@@ -10,7 +10,6 @@ import { warn } from '@hydrofoil/labyrinth/lib/logger'
 import { sparql } from '@tpluscode/rdf-string'
 import TermSet from '@rdfjs/term-set'
 import RdfResource, { RdfResourceCore } from '@tpluscode/rdfine/RdfResource'
-import { NotFoundError } from './errors'
 
 interface ResourceCreationOptions {
   implicitlyDereferencable?: boolean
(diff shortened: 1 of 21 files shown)


martinmaillard (approved):

lucafurrer (approved):
:+1:

lucafurrer (approved):

apis/core/lib/ResourceStore.ts
```diff
@@ -110,31 +116,40 @@ export default class implements ResourceStore {
...
-    if (!this.__resources.has(term)) {
-      const resource = await this.__storage.loadResource(term)
-      if (resource) {
-        this.__resources.set(term, resource)
+
+    if (!resource) {
+      if (opts?.allowMissing) {
+        return undefined as any
       }
+      throw new NotFoundError(term)
```
martinmaillard, line 145: If I'm not mistaken, `NotFoundError` results in a 404 response from the API. I think in a lot of cases that's not what we want when loading a resource from the store. 404 should only happen when the resource at the requested URI doesn't exist, but not when another resource is not found while processing the request.

tpluscode: Fair point 👌

martinmaillard: I see your fix, but now how about all the cases that *should* be 404? :D

tpluscode: hydra-box will do that earlier, before it gets to our handlers

martinmaillard: Oh, alright then :)

martinmaillard (approved):

PR #312, opened 2020-12-09
Default Cube ID did not end in slash

### LINKED ISSUE #272: Add "/" at the end of generated project namespace
When namespace is not provided, a default one is generated. It should have a trailing "/".

tpluscode:
Also improves appending the version to published cube (somewhat caused by the namespace issue)

fixes #272


martinmaillard (approved):
🚀

PR #323, opened 2020-12-10
fix: rename cube when namespace changes

tpluscode:
When a project update changes the `cc:namespace` the cube should also be updated


lucafurrer (commented):
What happens with the objects that have a name based on the namespace, p.ex DimensionMetadata `<dimMeta> schema:about <namespace/blabla>

tpluscode:
Good point, thanks! I was looking at the generated graph, which can be ignored. But the metadata needs to be updated too

lucafurrer (approved):

PR #325, opened 2020-12-10
chore: add a default value for publish graph

### LINKED ISSUE #317: Default value for `publish graph` and hide field

tpluscode:
..and hide it from the project creation form

fixes #317

### DIFF
--- a/apis/core/bootstrap/shapes/cube-project.ts
+++ b/apis/core/bootstrap/shapes/cube-project.ts
@@ -1,5 +1,5 @@
 import { cc, shape } from '@cube-creator/core/namespace'
-import { hydra, rdfs, sh } from '@tpluscode/rdf-ns-builders'
+import { dash, hydra, rdfs, sh } from '@tpluscode/rdf-ns-builders'
 import { turtle } from '@tpluscode/rdf-string'
 
 export const CubeProjectShape = turtle`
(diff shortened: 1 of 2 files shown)


apis/core/bootstrap/shapes/cube-project.ts
```diff
@@ -46,6 +46,8 @@ ${shape('cube-project/create')} {
       ${sh.maxCount} 1 ;
       ${sh.nodeKind} ${sh.IRI} ;
       ${sh.order} 40 ;
+      ${dash.hidden} true ;
+      ${sh.defaultValue} <https://lindas.admin.ch/foen/cube> ;
```
lucafurrer, line 50: Just wondering if it would not be a good thing to have this in a env variable?

lucafurrer (approved):

apis/core/bootstrap/shapes/cube-project.ts
tpluscode: Well ackchully, I think this would fit most in some kind of system-wide configuration, which we don't have.

The difference being that env requires redeployment, and ideally this would not be necessary for this.

lucafurrer: I agree, but still I think that it would be useful to at least be able to have different values for different instances

martinmaillard (approved):

PR #326, opened 2020-12-10
Cube versions and timestamps

### LINKED ISSUE #318: Add version number in published cube
When publishing a cube, we are adding a version number in the URIs.

Add this same number as a property on the cube.

### LINKED ISSUE #321: Meaning and Behaviour of Dataset date and time Definition [WIP]
We have multiple date(-times) added to each Dataset. I like to clarify the meaning and behaviour here.

Based on https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/wikis/Dataset-Metadata

*dct:issued* and *schema:dateCreated* (creation of the dataset)
* A value to be defined through the user. (Potentially propose the current date in the user interface if empty.)

*schema:datePublished*
* A value automatically added to the dataset first time a publishing is done, with the current date. (stays from then on for all versions the same)
* Can be changed by user if needed.

*dct:modified* and *schema:dateModified*
* A value automatically added on every publishing process. Not visible in the UI.
* Can not be changed by the user.

tpluscode:
Adds version and timestamp metadata to published cubes

* [x] `schema:version` added to published cube
* [x] modified date added to published cube
* [x] `schema:validThrough` added to previous cube version
   * this is not only simple `version - 1` so could be wrong if gaps occur
* `schema:datePublished`
   * [x] added to cube on first publish
   * [x] also stored with the cube metadata in Creator
   * [x] added to cube metadata shape
*  `dcterms:issued` and `schema:dateCreated`
   * [x] added to cube metadata shape
   * [x] initial value set to current timestamp

will fix #321 
will fix #318


apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -189,19 +189,25 @@ ${shapeId} {
...
       ${sh.order} 160 ;
     ] ;
+    ${sh.property} [
+      ${sh.name} "Creation Date" ;
+      ${sh.path} ${schema.dateCreated} ;
+      ${sh.maxCount} 1 ;
+      ${sh.datatype} ${xsd.date} ;
+      ${dash.hidden} true ;
+      ${sh.equals} ${dcterms.issued} ;
+    ] ;
```
tpluscode, line 210: I added a feature to shaperone which automatically keeps properties in sync when one has `dash:hidden true` and `sh:equals`

By restricting this behavior only to `dash:hidden` it will not break "repeat password" kind of forms

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -189,19 +189,25 @@ ${shapeId} {
...
       ${sh.path} ${dcterms.issued} ;
-      ${sh.minCount} 1 ;
       ${sh.maxCount} 1 ;
       ${sh.datatype} ${xsd.date} ;
       ${sh.order} 150 ;
     ] ;
     ${sh.property} [
-      ${sh.name} "Last Update" ;
-      ${sh.path} ${dcterms.modified} ;
-      ${sh.minCount} 1 ;
```
tpluscode, line 200: Removed as per #321

ui/src/forms/FormProperty.vue
```diff
@@ -26,7 +26,7 @@
 import { Prop, Component, Vue } from 'vue-property-decorator'
 import { PropertyObjectState, PropertyState } from '@hydrofoil/shaperone-core/models/forms'
 import { TemplateResult } from 'lit-element'
-import { dash } from '@tpluscode/rdf-ns-builders'
+import '@rdfine/dash/extensions/sh/PropertyShape'
```
tpluscode, line 29: Makes the `hidden` property visible to typescript
Gotta fix shaperone so that will not be necessary in next version.

martinmaillard: I think it would make sense to put this comment in the code

ui/src/forms/FormProperty.vue
```diff
@@ -39,7 +39,7 @@ export default class extends Vue {
   @Prop() renderMultiEditor!: () => TemplateResult;
 
   get isHidden (): boolean {
-    return !!this.property.shape.pointer.out(dash.hidden).value
+    return this.property.shape.hidden === true
```
martinmaillard, line 42: Is the `=== true` part necessary?

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -189,19 +189,25 @@ ${shapeId} {
     ${sh.property} [
       ${sh.name} "Creation Date" ;
```
martinmaillard, line 190: If we're going to keep these dates around in the form, we need to add some `sh:description` to them.

ui/src/forms/FormProperty.vue
tpluscode: Technically not but `hidden` returns `boolean | undefined` which is not exactly the same as strict truth

tpluscode: Might as well change the return type of `isHidden`

martinmaillard: I would keep it that way then. Boolean is a better return type than `boolean | undefined`.

apis/core/bootstrap/shapes/dataset.ts
tpluscode: Ok, TBH I'm not sure what those should be

martinmaillard (approved):

Issue #329, opened 2020-12-11
"Value does not have shape Blank node b18056_b11" when saving Project Settings


danielmittag:
Steps to reproduce (INTEG, Edge)
- New Project (Publish Graph: "http://test")
- Go to `Project Settings`, save

```
The request payload does not conform to the SHACL description of this endpoint.

csvMapping
Value does not have shape Blank node b18236_b11
```

martinmaillard:
This horrible error message happens when the `namespace` doesn't have a correct format (e.g. missing trailing "/"). I'll see what I can do to improve that.

PR #337, opened 2020-12-14
feat: edit project namespace inline

### LINKED ISSUE #313: Update project: namespace disappears
When a project is modified, the namespace does not appear in the form any more until browser refresh (it is actually save correctly)
Dec-09-2020 13-53-04

martinmaillard:
Edit project namespace inline instead of in a nested form.

It should fix #313 and #329

### DIFF
--- a/ui/src/views/CubeProjectEdit.vue
+++ b/ui/src/views/CubeProjectEdit.vue
@@ -38,6 +38,7 @@ import { Project } from '@cube-creator/model'
 import HydraOperationForm from '@/components/HydraOperationForm.vue'
 import { GraphPointer } from 'clownface'
 import { Shape } from '@rdfine/shacl'
+import { cc } from '@cube-creator/core/namespace'
 import { api } from '@/api'
 import { APIErrorValidation, ErrorDetails } from '@/api/errors'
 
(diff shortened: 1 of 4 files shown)


ui/src/views/CubeProjectEdit.vue
```diff
@@ -61,7 +62,8 @@ export default class CubeProjectEditView extends Vue {
       this.shape = await api.fetchOperationShape(operation)
     }
 
-    this.resource = Object.freeze(this.project?.pointer) ?? null
+    const namespace = this.project.csvMapping?.namespace ?? ''
+    this.resource = Object.freeze(this.project.pointer.addOut(cc.namespace, namespace))
```
tpluscode, line 66: I'd rather we could avoid this kind of tweaks.

tpluscode (commented):
The validation can be improved by adding `sh:message` to the property shape of `cc:csvMapping`. Not perfect but enough

For the disappearing namespace, what is the actual cause?

ui/src/views/CubeProjectEdit.vue
martinmaillard: I understand and I don't like it either, but having a nested form in the UI is not a good option either. There are multiple solutions to this problem:
- Have a separate screen to edit Mapping settings
- Magically provide `namespace` on the Project from the API
- Support complex sh:path in shaperone (e.g. `sh:path (cc:csvMapping cc:namespace)`)

For the disappearing namespace, I didn't analyze in detail, but I'm pretty sure it depends on if the full csvMapping is present in the dataset of the clownface pointer used in the form or not. In some cases, `project.csvMapping` is just a `link`. I think it worked by chance when first loading the form.

tpluscode: No good options. Fair, enough, let's have it like this for now

Issue #340, opened 2020-12-14
More Available Data Types


danielmittag:
Feature Request by Fabian:

The current list of the most useful data types is goot, but it isn't possible to add another type if needed (f. ex. Boolean)

![GitHubIssue_340](https://user-images.githubusercontent.com/17119538/102081466-911e1a00-3e10-11eb-9975-d2cfd17d91f4.png)

PR #343, opened 2020-12-14
Add more datatypes

martinmaillard:
Addresses #340 

@tpluscode Is it correct to set the default datatype to `xsd:string` or will it cause issues that I didn't think about? (maybe with the language)


tpluscode:
I'm not sure what CSVW will do but technically it's either datatype or language.

martinmaillard:
Ok, then I'll just make `datatype` optional. It probably makes more sense.

tpluscode (approved):

PR #348, opened 2020-12-15
Store refactor

tpluscode:
Two important refactoring, before the resource store bites back:

1. move `store.save` to the handlers so that domain code does not have to care about it.
2. store will not write back unchanged graphs. this will reduce the risk for potential graphs being overwritten as we don't have optimistic locking


martinmaillard (approved):
Haven't tested but code looks fine

lucafurrer (approved):
I like this!

PR #350, opened 2020-12-15
fix: use schema.workExample to indicate where to publish

### LINKED ISSUE #319: Change publishOn metadata
We need to replace publishOnOpendata and publishOnVisualize with schema:workExample <https://ld.admin.ch/application/opendataswiss> and schema:workExample <https://ld.admin.ch/application/visualize>

Please see for the definition in https://github.com/zazuko/rdf-cube-schema-viz/issues/2

If possible have only the two mentioned entries available to be selected.
(But it needs to be possible to select multiple.)

lucafurrer:
fix #319 

grafik


martinmaillard (approved):
Doesn't look great, but good enough for now ;)

lucafurrer:
> Doesn't look great, but good enough for now ;)

Agreed!

tpluscode:
How hard is it going to be to render a component where both buttons can be selected?

PR #360, opened 2020-12-16
fix: allow more error logging in Sentry

tpluscode:


martinmaillard (approved):

sandhose (approved):
Splitting the handler in two is better :+1:

PR #361, opened 2020-12-16
fix: dcterms:descritption instead of comment

lucafurrer:
Jeremy Stucky found an error 

There was `dcterms:comment` instead of `dcterms:description`

Don't ask me where this came from...

### DIFF
--- a/apis/core/bootstrap/shapes/dataset.ts
+++ b/apis/core/bootstrap/shapes/dataset.ts
@@ -39,7 +39,7 @@ ${shapeId} {
     ] ;
     ${sh.property} [
       ${sh.name} "Description" ;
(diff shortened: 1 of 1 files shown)


martinmaillard:
We should probably also fix existing cubes in integ. The only one I found with a description is https://pipelines-integ.lindas.admin.ch/app/cube-projects/cube-project!!jeremys-penguins-z72jadwl37w/metadata

martinmaillard (approved):

lucafurrer:
@martinmaillard Do we have already think about how to do this?

martinmaillard:
No, but we're gonna have to now that people are using the app

PR #363, opened 2020-12-16
fix: more selectively accept dataset metadata

tpluscode:
The dataset metadata update endpoint was all too eager to save anything it received.

This PR changes this so that only saved quads will be those which has the dataset for subject and blank node children.

Specifically, any pre-existing triples about `<dataset> schema:hasPart ?cube` will be left intact


martinmaillard (approved):

lucafurrer (approved):
It would be really nice to be able to simply not remove everything not copied by copySubgraph. But this is for an other moment

PR #364, opened 2020-12-16
feat: add real paging to cube view query

tpluscode:
Apply page size to observations query (right now still only first page)

### DIFF
--- a/apis/core/lib/domain/observations/index.ts
+++ b/apis/core/lib/domain/observations/index.ts
@@ -23,16 +23,16 @@ export async function getObservations({ cubeId, sourceGraph, filters, templatePa
     throw new DomainError(`Cube not found: '${cubeId}'`)
   }
 
-  const view = createView(cube)
+  const view = createView(cube, pageSize || DEFAULT_PAGE_SIZE)
   if (filters) {
     populateFilters(view, filters)
   }
 
   const observations = await view.observations()
   return createHydraCollection({
-    observations: observations.slice(0, pageSize || DEFAULT_PAGE_SIZE),
+    observations,
     templateParams,
     template,
-    totalItems: observations.length,
+    totalItems: 0,
   })
 }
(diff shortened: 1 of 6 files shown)


apis/core/lib/domain/observations/index.ts
```diff
@@ -23,16 +23,16 @@ export async function getObservations({ cubeId, sourceGraph, filters, templatePa
...
   }
 
   const observations = await view.observations()
   return createHydraCollection({
-    observations: observations.slice(0, pageSize || DEFAULT_PAGE_SIZE),
+    observations,
     templateParams,
     template,
-    totalItems: observations.length,
+    totalItems: 0,
```
tpluscode, line 36: @bergos any suggestions on getting a total count for a given view.

Like, I could create two views: one with projection and the other without it, and then do a `?COUNT( * )` query on the latter. Is that possible?

apis/core/lib/domain/observations/lib/index.ts
```diff
@@ -38,8 +38,19 @@ export function populateFilters(view: View, filters: AnyPointer): void {
...
 
-export function createView(cube: Cube): View {
-  return CubeQuery.View.fromCube(cube)
+export function createView(cube: Cube, pageSize: number): View {
+  const view = CubeQuery.View.fromCube(cube)
+
+  view.ptr.addOut(ns.view.projection, projection => {
+    const order = projection.blankNode()
+      .addOut(ns.view.dimension, view.dimensions[0].ptr)
+      .addOut(ns.view.direction, ns.view.Ascending)
```
tpluscode, line 47: For now hardcodes ascending order over the first dimension (the observation ID I believe)

martinmaillard (approved):
Cool!

Is this going to make it easier to have actual pagination or is the feature still missing from the query lib?

tpluscode:
No no, definitely possible. Would just need `view:offset` and add that to the URL template

PR #372, opened 2020-12-22
fix(cli): add more logging to transformation pipeline

tpluscode:
For some reason the pipelines failed on gitlab when they get to parsing the CSV

I appears that they get XML from S3 which might indicate that something is off about the download link...


tpluscode:
I have no idea. Simply running the jobs from this branch works without changing anything else.

Something must be off about the latest `master` branch builds 🤔

tpluscode:
AHA!. There are no `master` docker tags 🤦‍♂️. Only `edge`

martinmaillard (approved):

PR #379, opened 2021-01-05
Extract some project settings to Organisation

### LINKED ISSUE #359: Extract organisation settings
We have been discussing ways to make it simpler for users to set up a project.

Currently we appear to have two setting which are per-organisation (BAR, BAFU), both of which are currently defined on every project:

1. Cube namespace
2. Publish graph

## Cube namespace

The cube namespace is a base URI for all cube identifiers (observations, dimension predicates). While a part of the namespace would remain project-specific, we would like to move the `http//` portion to a shared location for all projects within an organisation.

## Publish graph

This one is simpler, and could only be moved from the project to the organization.

## Before and after (proposal)

**Before**

```turtle
<project> a cc:CubeProject ;
  cc:publishGraph  <https://lindas.admin.ch/foen/cube> ;
  cc:csvMapping <project/csv-mapping> ;
.

<project/csv-mapping>
  a cc:CsvMapping ;
  cc:namespace <https://environment.ld.admin.ch/foen/ubd/28/> 
.
```

**After**

```diff
<project> a cc:CubeProject ;
- cc:publishGraph  <https://lindas.admin.ch/foen/cube> ;
  cc:csvMapping <project/csv-mapping> ;
+ schema:maintainer <organization/bafu> ;
+ cc:cubeIdentitifer "ubd/28/" ;
.

<project/csv-mapping>
  a cc:CsvMapping ;
- cc:namespace <https://environment.ld.admin.ch/foen/ubd/28/> ;
.

+<organization/bafu>
+ a schema:Organization ;
+ cc:publishGraph  <https://lindas.admin.ch/foen/cube> ;
+ cc:namespace <https://environment.ld.admin.ch/foen/> .
```

## Impact

Currently the `cc:namespace` is derived from the project URI (also created semi-randomly to ensure uniqueness). I think this does not really make sense and that that value should only ever be provided by the user.

Thus, for any given project, transformation would be impossible until the users set it through the UI. This will prevent "surprise cubes" with strange URIs.

---

The project creation/update form would change: namespace and publish graph (hidden field) would be replaced by an element (dropdown) to select an organisation from among the instances of `schema:Organization`

## Open questions

1. Would we allow any arbitrary URI as the cube identifier or should they always be rooted in the `cc:namespace`?

tpluscode:
fixes #359


apis/core/lib/domain/cube-projects/create.ts
```diff
@@ -26,30 +26,35 @@ export async function createProject({
...
     label,
-    publishGraph,
+    maintainer,
+    cubeIdentifier,
   })
 
   project.initializeJobCollection(store)
-  const dataset = Dataset.create(store.create(project.dataset.id))
+  const dataset = Dataset.create(store.create(project.dataset.id), {
+    [dcterms.identifier.value]: project.cubeIdentifier,
```
martinmaillard, line 47: The `identifier` field in the dataset form needs a description. I find it extremely confusing that we have 2 fields named "identifier". It's even more confusing now that the dataset one defaults to being the same value as the cube one, but when I change one it doesn't change the other.

martinmaillard (approved):
Apart from by 2 comments, everything seems to work 👍

apis/core/lib/domain/cube-projects/create.ts
tpluscode: Any suggestions? I don't know what I would write there

martinmaillard: Not really because I myself don't understand the difference

tpluscode: The value you provide when creating the project is used for the URL and initial metadata value

When editing, those values are separated because the URL has to be stricter than the metadata identifier (pretty much any string allowed I guess)

martinmaillard: Ok but what is the dataset one used for? That's what we need to explain in the UI

tpluscode: Everything about the dataset is just metadata. Stuff needed to be published with the cube. Not unlike any other of those dataset properties

martinmaillard: Metadata still has an effect somewhere, or a usefulness for someone. We need to explain to users "why" they have to provide this value, and what will happen with it (if applicable). Is it the "cube identifier on opendata.swiss"? Or something like that?

l00mi: https://www.w3.org/TR/vocab-dcat-2/#Property:resource_identifier seems to be optional first of all.

Is it possible to change the "project" identifier? If so I would actually remove the dataset.identifier and re-use the "project" identifier. Because to have a dataset with a URI and then added another Identifier will be confusing.

l00mi: Will the "metadata" identifier prepopulated in the current version?

PR #385, opened 2021-01-11
feat(ui): display version in footer

### LINKED ISSUE #327: Show the deployed version in the user interface
For end-user testing purposes it will be useful if the user can report the deployed version a potential bug was reported on.

Show the the deployed version on the user interface. (E.g. in a footer.)

martinmaillard:
Fixes #327 

The version number will currently be the hash of the deployed commit. We could improve this by adding the latest tag, but I didn't want to change the build process in this PR without analyzing the consequences first.


l00mi (approved):
That is perfect! Thank you.

PR #389, opened 2021-01-11
feat(ui): show version tag and commit

martinmaillard:
Show version as: `version 1.5.1 (0dk4ix8)`


ui/src/App.vue
```diff
@@ -49,7 +49,8 @@ export default class App extends Vue {
   @appNS.State('messages') messages!: Message[]
 
   get release (): string {
-    return process.env.VUE_APP_RELEASE || '?'
+    const commit = process.env.VUE_APP_COMMIT?.slice(0, 7) ?? 'dev'
```
l00mi, line 52: 8.. okay at least they are not random (-:

l00mi (approved):

ui/src/App.vue
martinmaillard: 7 characters is quite standard when referencing git commits. And no, not random ;)

PR #392, opened 2021-01-11
fix: delete table when source removed

### LINKED ISSUE #309: CSV Mapping: Deleting a Input CSV doesn't delete its Output tables
Steps to reproduce:
- Create a new project
- add a CSV
- Create a table "Tabelle1"
- Delete the Output table "Tabelle1" (proof for correct behaviour)
- Create a table "Tabelle2"
- Try "Link to another table" 
  - --> OK, the "Link to table"-Dropdown will only contain "Tabelle2"
- Delete the whole Input CSV
  - --> OK, the associated output table disappears too
- Upload another CSV
- Create a table "Tabelle3"
- Try "Link to another table" 
  - --> BUG, the "Link to table"-Dropdown will contain "Tabelle2" and  "Tabelle3".  "Tabelle2" should've been deleted.

Remarks
- Having the table itself on the list would create a recursion, I don't think this should be allowed (cf. #310)

tpluscode:
@martinmaillard @lucafurrer do you remember why we only unset the table property and not actually remove the table?

fixes #309

### DIFF
--- a/apis/core/test/domain/csv-source/delete.test.ts
+++ b/apis/core/test/domain/csv-source/delete.test.ts
@@ -7,16 +7,13 @@ import { rdf, schema } from '@tpluscode/rdf-ns-builders'
 import { TestResourceStore } from '../../support/TestResourceStore'
 import clownface from 'clownface'
 import type { FileStorage } from '../../../lib/storage/s3'
-import type * as TableQueries from '../../../lib/domain/queries/table'
+import * as TableQueries from '../../../lib/domain/queries/table'
 import '../../../lib/domain'
 import { deleteSource } from '../../../lib/domain/csv-source/delete'
+import * as DeleteTable from '../../../lib/domain/table/delete'
 
 describe('domain/csv-sources/delete', () => {
   let fileStorage: FileStorage
-  let tableQueries: typeof TableQueries
-  let getLinkedTablesForSource: sinon.SinonStub
-  let getTablesForMapping: sinon.SinonStub
-  const getTableForColumnMapping = sinon.stub()
 
   beforeEach(() => {
     fileStorage = {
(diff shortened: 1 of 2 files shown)


martinmaillard:
Not really. Probably because "maybe in the future we will X". I also think it's safer to delete the table.

lucafurrer:
If I remember right, there was the idea, that you can delete the source upload a new one and reconnect it with the new source. The idea at the time was that the table is not strongly linked to the source.

I could look through my discussions etc, but I think it is not worth the time.

apis/core/test/domain/csv-source/delete.test.ts
```diff
@@ -25,17 +22,16 @@ describe('domain/csv-sources/delete', () => {
...
 
-    getLinkedTablesForSource = sinon.stub().returns(tableGenerator())
-    getTablesForMapping = sinon.stub().returns(tableGenerator())
-    tableQueries = {
-      getLinkedTablesForSource,
-      getTablesForMapping,
-      getTableForColumnMapping,
-    }
+    sinon.restore()
+    sinon.stub(TableQueries, 'getLinkedTablesForSource').returns(tableGenerator())
```
martinmaillard, line 30: What made you decide to stub stuff instead of injecting dependencies?

martinmaillard (approved):

apis/core/test/domain/csv-source/delete.test.ts
tpluscode: I think it's cleaner on the code which actually uses them. What do you think?

martinmaillard: I'm fine with both. Mocking in this way makes the code a bit simpler, but the test a bit more brittle. The tradeoff is probably worth it.

PR #396, opened 2021-01-12
feat: cube resource preview

### LINKED ISSUE #358: Cube preview: click on linked resource to see properties
When a dimension contains resources, clicking on one of them should open a tab/page that displays the properties of this resource.

martinmaillard:
Display properties of a resource in Cube Preview.

One thing left to fix: the endpoint seems to load all cube data in memory, which makes it really slow. It's not related to the code of the handler (I tried with an empty handler).

Will close #358


tpluscode (approved):
Looks good

We would need to change hydra-box and labyrinth's loader to reduce the database traffic separately

martinmaillard:
I added a message about the slowness for now so that we can merge.

But we should really find a solution to consider this feature complete. Worst case, we just use another resource/url instead of the cube-data.

PR #410, opened 2021-01-13
chore: remove rdf-vocabularies copied function

martinmaillard:


yarn.lock
```diff
@@ -3009,6 +3009,18 @@
...
+  resolved "https://registry.yarnpkg.com/@zazuko/rdf-vocabularies/-/rdf-vocabularies-2020.12.16.tgz#aa98fcb536a55b01c8b37f4512440b769d039737"
+  integrity sha512-XBWDIY9ZcoOb7Ft+sFFec1qH7X16y4zcjoSXV4Tels+8yy2ExwCuwoCwnefsOtUULwxAeyTqgILMybFv5X8ADA==
+  dependencies:
+    "@rdfjs/parser-n3" "^1.1.2"
+    commander "^5.0.0"
+    pkg-dir "^5.0.0"
+    rdf-ext "^1.3.0"
+    readable-stream "^3.6.0"
+    string-to-stream "^3.0.1"
+
```
tpluscode, line 3023: Please remove duplicate of `@zazuko/rdf-vocabularies`

martinmaillard: I think it's done. There's still a duplicate because of some other libraries depending on it.

tpluscode: Versions `2020.11.2` and `2020.12.16` are compatible ranges. They should also be compatible with each other.

I don't know why but yarn likes to add them twice and you have to manually combine the lock entries. Would look like 

```
"@zazuko/rdf-vocabularies@^2020.11.2", "@zazuko/rdf-vocabularies@^2020.12.16":
```

martinmaillard: Does that make a difference?

tpluscode: Probably not with rdf-vocabularies, if you think it's good to bundle libraries multiple times.

martinmaillard: Well I'm not really sure what the implications are, that's why I'm asking. Doesn't webpack only take the functions it needs anyway in the end?

tpluscode: Sure, but it won't dedupe between different versions of same package

martinmaillard: I tried combining the two. Let's see

tpluscode (approved):

PR #412, opened 2021-01-13
fix: precalculate page links

tpluscode:
Adds `next`/`previous` links to the collection which will remove the need for UI to manage the page index


martinmaillard:
Great! Can we merge the other PR first and I'll look into this one later?

tpluscode:
Yea, however you prefer

martinmaillard:
Shit, that was not on purpose

PR #416, opened 2021-01-18
Update hydra-box and labyrinth

tpluscode:
This will improve the performance of API requests by optimising the loading of resources:

1. `hydra-box`/`labyrinth` only loads minimal triples necessary to find operations and not entire graphs
   - complete representation available as `async req.hydra.resource.dataset()`, `req.hydra.resource.quadStream()` and `async req.hydra.resource.clownface()`
2. `labyrinth` combines what previously was multiple queries into a single SELECT to reduce SPARQL round-trips

### DIFF
--- a/apis/core/lib/handlers/dataset.ts
+++ b/apis/core/lib/handlers/dataset.ts
@@ -15,7 +15,7 @@ export const put = protectedResource(
   shaclValidate,
   asyncMiddleware(async (req, res) => {
     const dataset = await update({
-      dataset: clownface(req.hydra.resource),
+      dataset: await req.hydra.resource.clownface(),
       resource: await req.resource(),
       store: req.resourceStore(),
     })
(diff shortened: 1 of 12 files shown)


apis/core/lib/handlers/dataset.ts
```diff
@@ -42,7 +42,7 @@ export const loadCubes: Enrichment = async (req, dataset) => {
     return
   }
 
-  dataset.any().has(rdf.type, cube.Cube).forEach(cube => {
+  clownface(dataset).any().has(rdf.type, cube.Cube).forEach(cube => {
```
tpluscode, line 45: Necessary because `hydra-box` still uses `0.x` of clownface which does not have `.any()`

cue zazuko/hydra-box#92

martinmaillard:
Getting `req.hydra.resource.clownface is not a function` when creating a new table

martinmaillard (approved):

PR #423, opened 2021-01-19
fix(ui): broken property full-URI hover

### LINKED ISSUE #420: Fix wrong value in hover on properties
Fix the wrongly concatenated hover in the case of an external property:
Screenshot 2021-01-18 at 16 21 40

martinmaillard:
Fix #420

### DIFF
--- a/ui/src/components/PropertyDisplay.vue
+++ b/ui/src/components/PropertyDisplay.vue
@@ -16,14 +16,14 @@ const projectNS = namespace('project')
 @Component
 export default class PropertyDisplay extends Vue {
   @Prop() term!: Term
-  @projectNS.State((state) => state.createIdentifier) createIdentifier!: CreateIdentifier
+  @projectNS.State('createIdentifier') createIdentifier!: CreateIdentifier
 
   get value (): string {
     return this.term.value || ''
   }
 
   get expanded (): string {
-    return this.createIdentifier(this.value)
+    return this.createIdentifier(this.term)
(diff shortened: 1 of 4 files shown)


ui/src/components/PropertyDisplay.vue
```diff
@@ -23,7 +23,9 @@ export default class PropertyDisplay extends Vue {
   }
 
   get expanded (): string {
-    return this.createIdentifier(this.value)
+    return this.value.includes(':')
+      ? expand(this.value)
+      : this.createIdentifier(this.value)
```
tpluscode, line 28: The underlying `Organization#createIdentifier` already does that but it treats strings a term template.

I would also pass the `Term` instead. (needs also update a type declarations)

```suggestion
    return this.createIdentifier(this.term)
```

martinmaillard: I don't see any call to `expand` in `createIdentifier`

tpluscode: True, it does not expand. The input is either a Term (returned as-is) or a template, such as `cc:propertyTemplate` which gets concatenated. Prefixed URI is not an expected input

martinmaillard: Changed

martinmaillard: Crap, I need to change again then :D I wonder why it worked

PR #427, opened 2021-01-19
fix dimension metadata

l00mi:


l00mi:
@tpluscode the opendata.swiss definitions are obsolte correct?

tpluscode (approved):

tpluscode:
Yes. And they were wrong too :)

Issue #401, opened 2021-01-12
META: Technical Debt


l00mi:
* Transform can not be used if GitHub Actions is used for pipelines. (Missing field to add token.)
* Remove override for cube-test https://github.com/zazuko/cube-creator/pull/379/files/9e64b35552168041054288113a5f397bb69dde42..ca4255631c6a1eb643daceba224d3c93fd1ba425

l00mi:
* Remove, refactor admin.ch related parts
  * opendata.swiss related or named fields 
  * https://github.com/zazuko/cube-creator/pull/853/files#diff-416486e053a13581d750733185e8b6755430a8aaaa4b97f38ebac14335848abbR13

tpluscode:
Re. cube metadata, currently category (`dcat:theme`). Queried directly from the UI from the published cubes database

Until we decide on a better solution, they can be populated from Lindas like

```sparql
prefix schema: <http://schema.org/>
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>

INSERT {
  graph <https://lindas.admin.ch/sfa/opendataswiss> {
    ?theme a schema:DefinedTerm ;
      schema:name ?name ;
      schema:inDefinedTermSet <https://register.ld.admin.ch/opendataswiss/category> ;
      ?p ?o .
  }
} WHERE {
    service <https://int.lindas.admin.ch/query> {
          graph <https://lindas.admin.ch/sfa/opendataswiss> {
                  ?theme a schema:DefinedTerm ;
                         schema:name ?name ;
                         schema:inDefinedTermSet <https://register.ld.admin.ch/opendataswiss/category> ;
                         ?p ?o .
          }
    }
}
```

PR #428, opened 2021-01-19
fix: use organization URI as cube:observedBy

tpluscode:
Removed the SSZ observer from the transform pipeline. The cube pipeline step does create a default value for `cube:observedBy` anyway, which gets replaced with the organization URI

@l00mi I think we need an issue for the future to track this and find the long term correct solution


l00mi:
tracked here https://github.com/zazuko/cube-creator/issues/401

l00mi (approved):

PR #430, opened 2021-01-19
fix: dimension metadata not published

tpluscode:
The problem was that the pipeline looked up metadata by wrong identifiers, because of cube revision being injected into the URIs

This went unnoticed because the shape check returned false positive when there was nothing to validate in the input pointer.

### DIFF
--- a/cli/lib/cube.ts
+++ b/cli/lib/cube.ts
@@ -26,6 +26,8 @@ export function injectRevision(this: Pick<Context, 'variables' | 'log'>) {
   const revision = this.variables.get('revision')
   const previousCubes = new TermMap<Term, QuadSubject>()
   this.variables.set('previousCubes', previousCubes)
+  const mappedTerms = new TermMap<Term, Term>()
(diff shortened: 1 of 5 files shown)


cli/lib/cube.ts
```diff
@@ -35,7 +37,9 @@ export function injectRevision(this: Pick<Context, 'variables' | 'log'>) {
 
   function rebase<T extends Term>(term: T, rev = revision): T {
     if (term.termType === 'NamedNode' && term.value.startsWith(cubeNamespace)) {
-      return $rdf.namedNode(term.value.replace(new RegExp(`^${cubeNamespace}/?`), `${cubeNamespace}/${rev}/`)) as any
+      const newTerm = $rdf.namedNode(term.value.replace(new RegExp(`^${cubeNamespace}/?`), `${cubeNamespace}/${rev}/`)) as any
+      mappedTerms.set(newTerm, term)
+      return newTerm
```
tpluscode, line 42: Keep track of old and new terms when applying the revision

l00mi (approved):

PR #435, opened 2021-01-20
Feat/overwrite csv

### LINKED ISSUE #275: Feature Request: Overwriting the CSV data should be possible
... if (and that will have to be checked) the data is compatible with the current mapping.

Currently the entire mapping must be remade if new data arrives or some updates have been made on the data.

lucafurrer:
Replaces the current csv with the new one, if all old columns are present in the new csv.

When the file is compatible, the file on S3 gets replaced and and the sample date gets reloaded (same as when editing the dialect)

fixes #275

### DIFF
--- a/apis/core/lib/domain/csv-source/update.ts
+++ b/apis/core/lib/domain/csv-source/update.ts
@@ -27,29 +27,37 @@ export async function update({
 
   csvSource.name = changed.name
   if (csvSource.setDialect(changed.dialect) && csvSource.associatedMedia.identifierLiteral) {
-    try {
-      csvSource.pointer.deleteOut(schema.error)
-      csvSource.columns = []
-
-      const fileStream = await fileStorage.loadFile(csvSource.associatedMedia.identifierLiteral) as Readable
-      const head = await loadFileHeadString(fileStream, 500)
-      const { header, rows } = await parse(head, {
-        delimiter: csvSource.dialect.delimiter,
-        quote: csvSource.dialect.quoteChar,
-      })
-
-      const sampleCol = sampleValues(header, rows)
-
-      for (let index = 0; index < header.length; index++) {
-        const name = header[index]
-        const column = csvSource.appendColumn({ name })
-        column.samples = sampleCol[index]
-      }
-    } catch (err) {
-      error(err)
-      csvSource.pointer.addOut(schema.error, err.message)
-    }
+    await updateColumns(csvSource, fileStorage)
   }
 
   return csvSource.pointer
+}
+
+export async function updateColumns(csvSource: CsvSource, fileStorage: s3.FileStorage): Promise<void> {
+  try {
+    if (!csvSource.associatedMedia.identifierLiteral) {
+      throw new Error('Key to file missing')
+    }
+
+    csvSource.pointer.deleteOut(schema.error)
+    csvSource.columns = []
+
+    const fileStream = await fileStorage.loadFile(csvSource.associatedMedia.identifierLiteral) as Readable
+    const head = await loadFileHeadString(fileStream, 500)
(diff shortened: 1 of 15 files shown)


tpluscode (changes requested):
I would propose to have the csv as a separate resource under the `CsvSource`. Given for example `/source/ubd2` that would be `/source/ubd2/raw`, `/source/ubd2/data` or something like that. The actual URL does not matter and could be added dynamically

We already have the `schema:associatedMedia` so maybe add it there. Probably would have to replace it with a named node.

```turtle
<source/ubd2> 
  a cc:CsvSource ;
  schema:associatedMedia </source/ubd2/data> .
```

This way the upload would become `PUT /source/ubd2/data` and we would trivially add a `GET` handler to download the raw csv.

apis/core/lib/domain/csv-source/update.ts
```diff
@@ -27,29 +27,38 @@ export async function update({
...
+      bom: true,
+      delimiter: csvSource.dialect.delimiter,
+      quote: csvSource.dialect.quoteChar,
+    })
+
+    const sampleCol = sampleValues(header, rows)
+
+    for (let index = 0; index < header.length; index++) {
+      const name = header[index]
+      const column = csvSource.appendColumn({ name })
```
tpluscode, line 57: Do you rely on the `CsvColumn` identifiers to be stable when called multiple times?

I think this is not a good idea. For one, if the column order changes you will get multiple `dtype.order` triples, and also quite weird. Looking at the code of `appendColumn`, it will be like this:

Before (2 columns):

```turtle
<source>
  csvw:column <source/A> ; <source/B> .

<source/A> dtype:order 0 .
<source/B> dtype:order 1 .
```

After (updated with a file like `X;A;B`)

```turtle
<source>
  csvw:column <source/A> ; <source/B> ; <source/X> .

<source/A> dtype:order 0, 3 .
<source/B> dtype:order 1, 3 .
<source/X> dtype:order 2 .
```

Please add a test for updating with a CSV with new columns and/or columns in changed order

lucafurrer: Not exactly, we actually delete all columns beforehand. Hence the order is not preserved. But I can add a unit test for this.

This code is existing from the dialect change function, I did a refactoring to avoid duplication of the same thing.

lucafurrer:
> I would propose to have the csv as a separate resource under the `CsvSource`. Given for example `/source/ubd2` that would be `/source/ubd2/raw`, `/source/ubd2/data` or something like that. The actual URL does not matter and could be added dynamically
> 
> We already have the `schema:associatedMedia` so maybe add it there. Probably would have to replace it with a named node.
> 
> ```turtle
> <source/ubd2> 
>   a cc:CsvSource ;
>   schema:associatedMedia </source/ubd2/data> .
> ```
> 
> This way the upload would become `PUT /source/ubd2/data` and we would trivially add a `GET` handler to download the raw csv.

I was thinking about that a little bit, definitly for the download this could be helpful. 

For the put, how would we get the corresponding source, because we are mainly changing this object (new samples etc)

apis/core/lib/domain/csv-source/update.ts
tpluscode: Where are the columns deleted, I think I missed it?

tpluscode:
> For the put, how would we get the corresponding source, because we are mainly changing this object (new samples etc)

The operation would be declared indirectly, on the `schema:associatedMedia` property. hydra-box would actually give you the parent resource. Just like https://github.com/zazuko/cube-creator/blob/master/apis/core/hydra/dimension-metadata.ttl#L16-L19

tpluscode (approved):

PR #436, opened 2021-01-20
feat(ui): autocomplete for identifier template

### LINKED ISSUE #421: Add the table name to the prefix in the case of an internal identifier (no http(s):// in the uri).
- [ ] In a table if the {Identifier template} is not absolute (no http(s):// in the uri) add the table name between the cube uri and the identifier itself.

- [ ] Show the composed URI below the field. Similar to the one below found in **Edit column mapping**

Screenshot 2021-01-18 at 17 00 58

martinmaillard:
Auto-fill with table name, auto-complete column names and validation for identifier template input.

Fix #421


apis/core/bootstrap/shapes/table.ts
```diff
@@ -27,8 +27,8 @@ const commonProperties = turtle`
     ${sh.path} ${cc.identifierTemplate} ;
     ${sh.minCount} 1 ;
     ${sh.maxCount} 1 ;
-    ${sh.defaultValue} "" ;
```
martinmaillard, line 30: For some reason, `dash:editor` was not taken into account when `sh:defaultValue` was specified. Now I have a feeling we'll run into issues without this default value?

tpluscode: Yes, without `sh:default` we would get validation errors unless the user interacts with the field. Not good.

The problem is that the preferred editor would get removed if there is no metadata about it.

Fixed soon by https://github.com/hypermedia-app/shaperone/pull/84.

martinmaillard: 👍 Can wait on the shaperone PR

martinmaillard: I upgraded shaperone but it seems like it still doesn't select the proper editor when creating a new table

tpluscode: Yes, my mistake

tpluscode (approved):

PR #439, opened 2021-01-21
fix: do not inject version to property shape paths

tpluscode:


l00mi (approved):
Can we test if its working on test?

tpluscode:
You could have merged yourself. Will deploy to test when CI passes

l00mi:
I want to ask, can you please make a small project and test if the problems are fixed. I am no blocked until 15:00 with the presentation.

PR #444, opened 2021-01-22
refactor: stream dataset resource

### LINKED ISSUE #402: In some cases, the /dataset endpoint crashes the container
I don't know what causes that, but sometimes the API container crashes when visiting the Cube Designer page. The `/dataset` endpoint seems to be responsible.

Example https://pipelines-test.lindas.admin.ch/app/cube-projects/cube-project!!case-study-udb0028-rabmzrajjla/metadata

tpluscode:
Refactor `/dataset` resource to stream without going through a clownface object to prevent a mysterious memory uptake.

This should fix #402


apis/core/lib/handlers/dataset.ts
```diff
@@ -26,26 +32,44 @@ export const put = protectedResource(
...
+  const shapeStream = await loadCubeShapes(req.hydra.resource.term, streamClient)
+  const outStream = new PassThrough({
+    objectMode: true,
+  })
+
+  const types = clownface({
+    dataset: req.hydra.api.dataset,
+    term: dataset.out(rdf.type).terms,
+  })
+  const linkedResources = await loadLinkedResources(dataset, types.out(query.include).toArray(), req.app.sparql)
```
tpluscode, line 46: This could be improved by making `loadLinkedResources` a stream and having it called automatically behind the scene. That way `req.hydra.resource.clownface()` would also be replaced with `req.hydra.resource.quadStream()` to avoid unnecessarily materialising all quads

martinmaillard (approved):
It seems to work but the endpoint still loads all the non-observation resources. Do we have a way to fix that?

tpluscode:
I'm pretty sure it's not. Maybe you have to run the transformation?

martinmaillard (approved):

PR #458, opened 2021-01-26
chore: managed dimension sample

tpluscode:
Proposed API resources to hold dimension mappings

The CI will fail, to be fixed in subsequent PRs

### DIFF
--- a/fuseki/sample-ubd.trig
+++ b/fuseki/sample-ubd.trig
@@ -111,7 +111,7 @@ graph <cube-project/ubd/dimension-mapping/pollutant> {
     prov:hadDictionaryMember [
       a prov:KeyEntityPair ;
       prov:pairKey "co" ;
-      prov:pairEntity <https://environment.ld.admin.ch/foen/ubd/28/pollutant/co> ;
+      prov:pairEntity <http://www.wikidata.org/entity/Q2025> ;
     ] ;
   .
 }
@@ -159,7 +159,7 @@ graph <cube-project/ubd/cube-data> {
     <https://environment.ld.admin.ch/foen/ubd/28/aggregation>     <https://environment.ld.admin.ch/foen/ubd/28/aggregation/annualmean> ;
     <https://environment.ld.admin.ch/foen/ubd/28/dimension/value> "5.4"^^xsd:float ;
     <https://environment.ld.admin.ch/foen/ubd/28/dimension/year>  "1999"^^xsd:gYear ;
-    <https://environment.ld.admin.ch/foen/ubd/28/pollutant>       <https://environment.ld.admin.ch/foen/ubd/28/pollutant/co> ;
+    <https://environment.ld.admin.ch/foen/ubd/28/pollutant>       <http://www.wikidata.org/entity/Q2025> ;
     <https://environment.ld.admin.ch/foen/ubd/28/station>         <https://environment.ld.admin.ch/foen/ubd/28/station/blBAS> ;
     <https://environment.ld.admin.ch/foen/ubd/28/unit-id>         "µg/m3" ;
   .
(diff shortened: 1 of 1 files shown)


fuseki/sample-ubd.trig
```diff
@@ -82,13 +83,39 @@ graph <cube-project/ubd/dimensions-metadata> {
 
   <cube-project/ubd/dimensions-metadata/pollutant>
     schema:about <https://environment.ld.admin.ch/foen/ubd/28/pollutant> ;
+    scale:scaleOfMeasure scale:Nominal ;
+    cc:dimensionMapping <cube-project/ubd/dimension-mapping/pollutant> ;
```
tpluscode, line 87: This would be created when a dimension is set to `Nominal`

fuseki/sample-ubd.trig
```diff
@@ -110,7 +137,7 @@ graph <cube-project/ubd/cube-data> {
     <https://environment.ld.admin.ch/foen/ubd/28/aggregation>     <https://environment.ld.admin.ch/foen/ubd/28/aggregation/annualmean> ;
     <https://environment.ld.admin.ch/foen/ubd/28/dimension/value> "6.2"^^xsd:float ;
     <https://environment.ld.admin.ch/foen/ubd/28/dimension/year>  "2003"^^xsd:gYear ;
-    <https://environment.ld.admin.ch/foen/ubd/28/pollutant>       <https://environment.ld.admin.ch/foen/ubd/28/pollutant/so2> ;
+    <https://environment.ld.admin.ch/foen/ubd/28/pollutant>       "so2" ;
```
tpluscode, line 140: Being an unmapped value, `"so2"` would appear as string coming form the CSV

fuseki/sample-ubd.trig
```diff
@@ -132,7 +159,7 @@ graph <cube-project/ubd/cube-data> {
     <https://environment.ld.admin.ch/foen/ubd/28/aggregation>     <https://environment.ld.admin.ch/foen/ubd/28/aggregation/annualmean> ;
     <https://environment.ld.admin.ch/foen/ubd/28/dimension/value> "5.4"^^xsd:float ;
     <https://environment.ld.admin.ch/foen/ubd/28/dimension/year>  "1999"^^xsd:gYear ;
-    <https://environment.ld.admin.ch/foen/ubd/28/pollutant>       <https://environment.ld.admin.ch/foen/ubd/28/pollutant/so2> ;
+    <https://environment.ld.admin.ch/foen/ubd/28/pollutant>       <https://environment.ld.admin.ch/foen/ubd/28/pollutant/co> ;
```
tpluscode, line 162: Changed this to `CO` to show that a value already mapped would be replaced by the correct URI

fuseki/sample-ubd.trig
```diff
@@ -164,7 +191,10 @@ graph <cube-project/ubd/cube-data> {
     sh:path      cube:observedBy .
 
   _:pollutant
-    sh:in        ( <https://environment.ld.admin.ch/foen/ubd/28/pollutant/so2> ) ;
+    sh:in        (
+                   "so2" # <https://environment.ld.admin.ch/foen/ubd/28/pollutant/so2>
+                   <https://environment.ld.admin.ch/foen/ubd/28/pollutant/co>
+                 ) ;
```
tpluscode, line 197: Finally, the Cube Shape would be a mix of mapped and unmapped value until all mappings are created.

fuseki/sample-ubd.trig
martinmaillard: You mean when a mapped managed dimension is selected right? Not all Nominal dimensions will be mapped

tpluscode: > You mean when a mapped managed dimension is selected right?

Actually no. See that the `cc:managedDimension` is property of the mapping resource and not the metadata resource. But you should have the option to unselect it to clear the dictionary and "disable" the mapping

martinmaillard: 👍

Issue #442, opened 2021-01-22
Add fields to attach a value unit (like "meters", "liters") for the dimension.


l00mi:
On the dimension metadata we need the user to add a unit.

The Unit will be attached to the shape of the Cube. See https://github.com/zazuko/rdf-cube-schema-viz/blob/7a70503b50c2ada37a9f1b24957680a97482a78d/example/ubd28-static.ttl#L5-L6

The UNIT is externally defined and the use can choose from of these http://www.qudt.org/doc/DOC_VOCAB-UNITS.html

(If a unit is missing it needs to be created manually inside another namespace. Also we need to check how to translate the Descriptions of the Unit.)

The QUDT will be available in lindas.

To discuss:

- [ ] Where to add specific new units if not available in QUDT.
- [ ] How to translate QUDT?

l00mi:
Find units with https://s.zazuko.com/aMWjm

l00mi:
Cross reference https://github.com/zazuko/rdf-cube-schema-viz/issues/4

l00mi:
Fixed with https://github.com/zazuko/cube-creator/pull/460

PR #460, opened 2021-01-26
feat: add dimension metadata unit

martinmaillard:
My SPARQL query has some issues and I'm going to need help to fix it.

Addresses #442

### DIFF
--- a/apis/core/bootstrap/lib/query.ts
+++ b/apis/core/bootstrap/lib/query.ts
@@ -3,5 +3,6 @@ import $rdf from 'rdf-ext'
 import { SparqlTemplateResult } from '@tpluscode/rdf-string'
 
 export function lindasQuery(query: SparqlTemplateResult): NamedNode {
-  return $rdf.namedNode('https://register.ld.admin.ch/query?query=' + encodeURIComponent(query.toString()))
+  // TODO: Switch to production system once stable.
+  return $rdf.namedNode('https://test.lindas.admin.ch/query?query=' + encodeURIComponent(query.toString()))
 }
--- a/apis/core/bootstrap/shapes/dimension.ts
+++ b/apis/core/bootstrap/shapes/dimension.ts
@@ -1,4 +1,4 @@
-import { shape, scale } from '@cube-creator/core/namespace'
+import { shape } from '@cube-creator/core/namespace'
 import { sparql, turtle } from '@tpluscode/rdf-string'
 import { lindasQuery } from '../lib/query'
 import { dash, hydra, rdf, rdfs, schema, sh, qudt } from '@tpluscode/rdf-ns-builders'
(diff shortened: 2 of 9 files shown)


ui/src/forms/editors/AutoCompleteEditor.vue
```diff
@@ -0,0 +1,48 @@
...
+  get _options (): Option[] {
+    return this.options.map(([pointer, label]) => ({ value: pointer.term.value, label, term: pointer.term }))
+  }
+
+  get _value (): Option | null {
+    return this._options.find(({ value }) => value === this.value?.value) ?? null
+  }
+
+  onInput (value: Option | null): void {
+    this.update(value?.term ?? null)
```
tpluscode, line 45: Is it multiselect but only one value would be selected?

tpluscode (commented):
> My SPARQL query has some issues and I'm going to need help to fix it.

What is the desired output?

It also returns a whole lot of results. Do we want to filter it by the entered text? Or maybe add a filter in general to remove some unit altogether?

ui/src/forms/editors/AutoCompleteEditor.vue
martinmaillard: It is not multiselect, even though the library I used allows that ;)

tpluscode: Kk. I only judged by the package name indeed.

l00mi:
@martinmaillard please check https://s.zazuko.com/F1K56

l00mi (approved):

apis/core/bootstrap/lib/query.ts
```diff
@@ -3,5 +3,6 @@ import $rdf from 'rdf-ext'
 import { SparqlTemplateResult } from '@tpluscode/rdf-string'
 
 export function lindasQuery(query: SparqlTemplateResult): NamedNode {
-  return $rdf.namedNode('https://register.ld.admin.ch/query?query=' + encodeURIComponent(query.toString()))
+  // TODO: Switch to production system once stable.
+  return $rdf.namedNode('https://test.lindas.admin.ch/query?query=' + encodeURIComponent(query.toString()))
```
tpluscode, line 7: What's this about?

apis/core/bootstrap/shapes/dimension.ts
```diff
@@ -1,6 +1,44 @@
...
+              FILTER NOT EXISTS {?unit ${qudt.unitOfSystem} ${sou.SOU_IMPERIAL} . }
+              FILTER NOT EXISTS {?unit ${qudt.unitOfSystem} ${sou.SOU_USCS} . }
+              FILTER NOT EXISTS {?unit ${qudt.unitOfSystem} ${sou.SOU_CGS} . }
+              FILTER NOT EXISTS {?unit ${qudt.unitOfSystem} ${sou.SOU_PLANCK} . }
+              FILTER NOT EXISTS {?unit a ${qudt.CurrencyUnit} . }
+            }
+        }
+      ORDER BY ASC(?ucumCount) ASC(?name)
+    }
+  }
```
tpluscode, line 40: This query still returns over 1300 results, much of which are derived units?

I don't see a good to way to filter more. Should we dynamically query by the typed input?

apis/core/bootstrap/lib/query.ts
l00mi: The current production endpoint gave me randomly 3 different results on the same query. I expect it to be solved with the migration to V7. So I don't want to spend more time here.

apis/core/bootstrap/shapes/dimension.ts
l00mi: It works reasonably fast as is. But yes, this might be useful, please propose in an  issue how to optimize this.

apis/core/bootstrap/lib/query.ts
martinmaillard: Note that the `lindasQuery` function is also used in other places (themes?). I hope the test endpoint also contains this data.

l00mi: yes, thank you, currently all "metadata" is in sync on all environments. so its save as long we use it not for other data. shall we add a comment in the code?

PR #464, opened 2021-01-27
feat: download csv

lucafurrer:
Allow the download of uploaded CSVs

### DIFF
--- a/apis/core/hydra/csv-source.ttl
+++ b/apis/core/hydra/csv-source.ttl
@@ -43,7 +43,7 @@ cc:CSVSource
         hydra:Operation,
         cc:GetCSVSourceAction ;
       hydra:title
-        "Get CSV Source" ;
+        "Get CSV" ;
       hydra:method
         "GET" ;
       code:implementedBy
--- a/apis/core/hydra/csv-source.ttl
+++ b/apis/core/hydra/csv-source.ttl
@@ -41,9 +41,9 @@ cc:CSVSource
     [
       a
         hydra:Operation,
-        cc:GetCSVSourceAction ;
+        schema:DownloadAction ;
       hydra:title
-        "Get CSV" ;
+        "Download CSV" ;
       hydra:method
         "GET" ;
       code:implementedBy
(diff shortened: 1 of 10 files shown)


apis/core/lib/handlers/csv-source.ts
```diff
@@ -75,7 +75,8 @@ const getCSVSource: express.RequestHandler = asyncMiddleware(async (req, res, ne
     return next(new Error('s3 key not found'))
   }
 
-  res.redirect(directDownload)
+  res.header('Location', directDownload)
```
lucafurrer, line 78: @tpluscode I run into CORS-Problems using redirect and still wondering if there is a better way than this.

ui/src/api/mixins/CSVSource.ts
```diff
@@ -15,6 +15,7 @@ export default function mixin<Base extends Constructor> (base: Base): Mixin {
     get _additionalActions () {
       return {
         replace: ns.cc.ReplaceCSVAction,
+        download: ns.cc.GetCSVSourceAction
```
tpluscode, line 18:
```suggestion
        download: schema.DownloadAction
```

packages/core/namespace.ts
```diff
@@ -28,7 +28,8 @@ type CubeCreatorClass =
   'Observations' |
   'TransformJob' |
   'PublishJob' |
-  'ReplaceCSVAction'
+  'ReplaceCSVAction' |
+  'GetCSVSourceAction'
```
tpluscode, line 32: Won't need a custom type. Can use `schema` (see other comment)

ui/src/components/CsvSourceMapping.vue
```diff
@@ -186,6 +191,11 @@ export default class CsvSourceMapping extends Vue {
     })
   }
 
+  async downloadSource (source: CsvSource): Promise<void> {
+    const downloadLink = await this.$store.dispatch('project/downloadCSV', source)
+    window.open(downloadLink)
```
tpluscode, line 196: I don't think this belongs to the component. Please move `window.open` to `store/modules/project.ts`

apis/core/hydra/csv-source.ttl
```diff
@@ -43,7 +43,7 @@ cc:CSVSource
         hydra:Operation,
         cc:GetCSVSourceAction ;
       hydra:title
-        "Get CSV Source" ;
+        "Get CSV" ;
```
tpluscode, line 46:
```suggestion
        "Download CSV" ;
```

maybe?

tpluscode (changes requested):

PR #466, opened 2021-01-27
Display lines between source and table on hover

### LINKED ISSUE #454: Show source of mapping on CSV Mapper
On the CSV Mapper make it more obvious on the overview from what column the "table" is mapped.

Ideas (from more fancy to less):

- Show line from source to target if hovered.
- Show line from source to target.
- Add in-front of mapped Predicate.

martinmaillard:
Fixes #454 (fancy)


l00mi (approved):
Genius

PR #468, opened 2021-01-28
fix: radio buttons always set literal

tpluscode:


apis/core/bootstrap/shapes/dimension.ts
```diff
@@ -46,9 +46,9 @@ ${shape('dimension/metadata')} {
...
   .
 
-  ${qudt.NominalScale} ${rdfs.label} "Nominal Scale"@en .
-  ${qudt.OrdinalScale} ${rdfs.label} "Ordinal Scale"@en .
-  ${qudt.IntervalScale} ${rdfs.label} "Interval Scale"@en .
-  ${qudt.RatioScale} ${rdfs.label} "Ratio Scale"@en .
+  ${qudt.NominalScale} ${rdfs.label} "Nominal"@en .
+  ${qudt.OrdinalScale} ${rdfs.label} "Ordinal"@en .
+  ${qudt.IntervalScale} ${rdfs.label} "Interval"@en .
+  ${qudt.RatioScale} ${rdfs.label} "Ratio"@en .
```
tpluscode, line 52: @l00mi with only four, the choices are displayed as radio buttons. the repeated `Scale` makes it a little long

ui/src/forms/editors/RadioButtons.vue
```diff
@@ -45,7 +45,10 @@ export default class extends Vue {
   }
 
   emit (value: string): void {
-    this.update(value)
+    const choice = this.choices.find(ptr => ptr.value === value)
+    if (choice) {
+      this.update(choice)
+    }
```
tpluscode, line 51: Set the actual `Term` and not its string value

l00mi (approved):

PR #474, opened 2021-01-29
feat: when deleting table delete all mappings referencing this table

### LINKED ISSUE #425: Deleting table linked by another table should delete the link
When deleting a table that is linked by another table, the column-mapping that links to the deleted table should probably also be deleted.

lucafurrer:
When a table to delete has any referencing column mappings, this mapping should also be deleted to avoid orphaned column mappings.

fixes #425


tpluscode (approved):
LGTM

PR #475, opened 2021-01-29
feat: getting dimension mappings ensure all values

tpluscode:
This PR ensures that all unmapped values found in `cube:Observation` and `cube:Constraint` are dynamically added to the mappings dictionary when GETing the resource.

The individual entries repeat the link to the dimension so that is can be queried from the editor using `hydra:search` annotation.


lucafurrer (approved):
Looks fine to me

martinmaillard (approved):

PR #482, opened 2021-02-01
Move CSV source buttons in dropdown

martinmaillard:
These actions are starting to pile up and are not used everyday. Let's move them in a dropdown menu.

I also refactored the "download CSV" code a bit.

### DIFF
--- a/ui/src/components/CsvSourceMapping.vue
+++ b/ui/src/components/CsvSourceMapping.vue
@@ -2,52 +2,46 @@
   <div class="source-mapping columns">
     <div class="column">
       <div class="source panel">
-        <div class="panel-heading">
-          <div class="level">
-            <div class="level-left">
-              <div class="level-item">
-                {{ source.name }}
-              </div>
-            </div>
-            <div class="level-right">
-              <b-button
-                v-if="tableCollection.actions.create"
-                :disabled="selectedColumns.length === 0"
-                tag="router-link"
-                :to="{ name: 'TableCreate', query: createTableQueryParams }"
-                size="is-small"
-                icon-left="plus"
-              >
-                Create table from selected columns
-              </b-button>
-              <div class="level-item">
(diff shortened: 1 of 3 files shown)


ui/src/components/CsvSourceMapping.vue
```diff
@@ -192,7 +194,13 @@ export default class CsvSourceMapping extends Vue {
...
 
   async downloadSource (source: CsvSource): Promise<void> {
-    const downloadLink = await this.$store.dispatch('project/downloadCSV', source)
+    const headers = { accept: 'text/csv' }
+    const response = await api.invokeDownloadOperation(source.actions.download, headers)
+    const downloadLink = response.xhr.headers.get('Location')
+
+    if (downloadLink) {
+      window.open(downloadLink)
+    }
```
lucafurrer, line 203: @tpluscode asked me to move the window.open down, but I am fine with this change

lucafurrer (approved):
Good idea!

ui/src/components/CsvSourceMapping.vue
martinmaillard: What do you mean by "move down"?

lucafurrer: That it lives in the function in the project file (which you deleted completely)

martinmaillard: 👍 

Yeah, I don't agree ;)

tpluscode: It is not called `getDownloadLink`, is it? I expected an action which "invokes download" to do exactly that rather than having the UI component look for a HTTP header and call `window.open`

I think it's at least two ways in which this violates boundaries of the involved components

martinmaillard: I didn't try to change the implementation. But I agree that the flow is a bit weird.

PR #484, opened 2021-02-01
feat(ui): show dimension property when missing name

martinmaillard:
Instead of just saying "Missing dimension name", say "Missing name for year".


l00mi (changes requested):
can you please make it "danger" wherever something is missing, people should understand here is a problem

martinmaillard:
done

l00mi (approved):

l00mi:
Thank you!

PR #498, opened 2021-02-03
feat: pipeline applies dimension mappings

### LINKED ISSUE #455: Replace mapped terms during transformation
Part of #376 

A `map` step must be added to the transformation pipeline to automatically replace textual values from the CSV to the previously mapped dimension URIs

lucafurrer:
Apply dimension mapping in the transform pipeline

fix #455

### DIFF
--- a/cli/lib/cube.ts
+++ b/cli/lib/cube.ts
@@ -41,7 +41,7 @@ export function injectRevision(this: Pick<Context, 'variables' | 'log'>) {
     return term
   }
 
-  return map(({ subject, predicate, object, graph }: Quad) => {
+  return map(async ({ subject, predicate, object, graph }: Quad) => {
     const rebasedSub = rebase(subject)
 
     if (revision > 1 && predicate.equals(rdf.type) && object.equals(cube.Cube)) {
--- a/cli/lib/cube.ts
+++ b/cli/lib/cube.ts
@@ -41,7 +41,7 @@ export function injectRevision(this: Pick<Context, 'variables' | 'log'>) {
     return term
   }
 
-  return map(async ({ subject, predicate, object, graph }: Quad) => {
+  return map(({ subject, predicate, object, graph }: Quad) => {
     const rebasedSub = rebase(subject)
 
     if (revision > 1 && predicate.equals(rdf.type) && object.equals(cube.Cube)) {
--- a/cli/lib/output-mapper.ts
+++ b/cli/lib/output-mapper.ts
@@ -1,0 +1,67 @@
+import $rdf from 'rdf-ext'
+import { TransformJob } from '@cube-creator/model'
+import { Hydra } from 'alcaeus/node'
+import type { Context } from 'barnard59-core/lib/Pipeline'
+import { Quad } from 'rdf-js'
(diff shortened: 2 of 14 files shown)


cli/lib/output-mapper.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  return mappingResource.representation.root?.hadDictionaryMember
+}
+
+export async function mapDimensions(this: Pick<Context, 'variables'>) {
+  Hydra.cacheStrategy.shouldLoad = (previous) => {
+    return !previous.representation.root?.types.has(prov.Dictionary) ||
+    !previous.representation.root?.types.has(hydra.ApiDocumentation)
+  }
```
lucafurrer, line 47: @tpluscode is this correct like this?

cli/pipelines/main.ttl
```diff
@@ -68,6 +68,7 @@
                        <#loadCsvStep>
                        <#parse>
                        <#filterObservationTable>
+                       <#applyMapping>
```
tpluscode, line 71: The name is a little vague. Perhaps
```suggestion
                       <#mapDimensions>
```

cli/lib/output-mapper.ts
tpluscode: Yes. I would in fact place it somewhere else though, in a more central place so that it will apply to all steps

cli/test/lib/output-mapper.test.ts
```diff
@@ -0,0 +1,115 @@
...
+import { expect } from 'chai'
+import $rdf from 'rdf-ext'
+import { cube } from '@cube-creator/core/namespace'
+import { mapDimensions } from '../../lib/output-mapper'
+import { insertTestData } from '@cube-creator/testing/lib'
+import { setupEnv } from '../support/env'
+import { env } from 'process'
+
+describe('lib/output-mapper', function () {
+  this.timeout(360 * 1000)
```
tpluscode, line 11: A tad too long maybe? :D

cli/lib/cube.ts
```diff
@@ -41,7 +41,7 @@ export function injectRevision(this: Pick<Context, 'variables' | 'log'>) {
     return term
   }
 
-  return map(({ subject, predicate, object, graph }: Quad) => {
+  return map(async ({ subject, predicate, object, graph }: Quad) => {
```
tpluscode, line 44: Won't have to be async if you also apply the change in the `.d.ts`
```suggestion
  return map(({ subject, predicate, object, graph }: Quad) => {
```

tpluscode (changes requested):

cli/lib/output-mapper.ts
lucafurrer: Suggestions?

cli/test/lib/output-mapper.test.ts
lucafurrer: maybe :wink:

cli/lib/output-mapper.ts
tpluscode: Like, in a new module in `lib` and import in both `lib/commands/*.ts`

PR #503, opened 2021-02-05
feat: meta:dataKind

### LINKED ISSUE #490: Add dimension property. cube:dataKind
Add a UI element to add the cube:dataKind as described in https://github.com/zazuko/rdf-cube-schema-viz/tree/master

Ideally there are two elements. The first to choose either "Spatial - Coordinates", "Spatial - Shape" and "Temporal". The second to choose only if "Temporal" is active between the "year", "month" .... as described.

Also okay to have a big list with "Spatial - Coordinates", "Spatial - Shape", "Temporal - year", "Temporal - month" ...

tpluscode:
fixes #490

Adds `cube:dataKind` property to dimension metadata

While the form does not support alternatives for `sh:node`, I added a field wrapper to disallow the `time:unitType` when the geo kinds are selected.

Feb-05-2021 09-10-18

At the same time I kept `sh:node sh:xone ( )` to actually do accurate request validation (the shapes to drive the form would be insufficient)


apis/core/bootstrap/shapes/dimension.ts
```diff
@@ -39,6 +39,46 @@ WHERE
...
+        ${sh.minCount} 1 ;
+        ${sh.maxCount} 1 ;
+      ] , [
+        ${sh.path} ${time.unitType} ;
+        ${sh.minCount} 1 ;
+        ${sh.maxCount} 1 ;
+      ] ;
+    ] ) ;
+  ] ;
+]`
```
tpluscode, line 80: Hidden specification of `cube:dataKind` for validation only

apis/core/bootstrap/shapes/dimension.ts
```diff
@@ -83,18 +123,92 @@ ${shape('dimension/metadata')} {
...
   ${qudt.RatioScale} ${rdfs.label} "Ratio"@en .
+
+  ${schema.GeoCoordinates}
+    ${rdfs.label} "Geographic coordinates"@en ;
+  .
+  ${schema.GeoShape}
+    ${rdfs.label} "Geographic shape"@en ;
+  .
+  ${time.GeneralDateTimeDescription}
+    ${rdfs.label} "Time description"@en ;
```
tpluscode, line 190: Label suggestions welcome

apis/core/bootstrap/shapes/dimension.ts
```diff
@@ -83,18 +123,92 @@ ${shape('dimension/metadata')} {
...
+          ${sh.order} 10 ;
+        ] , [
+          ${sh.name} "Time precision" ;
+          ${sh.path} ${time.unitType} ;
+          ${sh.maxCount} 1 ;
+          ${sh.nodeKind} ${sh.IRI} ;
+          ${sh1.if} [
+            ${sh.path} ${rdf.type} ;
+            ${sh.hasValue} ${time.GeneralDateTimeDescription} ;
+          ] ;
```
tpluscode, line 153: Custom annotation to allow `time.unitType` only when `?focusNode rdf:type time:GeneralDateTimeDescription`

martinmaillard: Not a fan of the prefix, but the feature looks great! I think it will allow me to do some stuff that I avoided until now :)

apis/core/bootstrap/shapes/dimension.ts
l00mi: I will ask the customer on the labels. Good for now.

apis/core/bootstrap/shapes/dimension.ts
l00mi: Just curious, could http://datashapes.org/constraints.html#HasValueWithClassConstraintComponent be used for a similar purpose?

tpluscode: Did not know that. Close but I think it's a little different

l00mi (approved):

PR #509, opened 2021-02-08
feat: redefine supported languages

### LINKED ISSUE #393: Order of languages (in editor) should be DE FR IT RM EN (instead of DE EN FR RM IT)
DE FR IT RM EN is the standard in federal offices as far as I remember.

martinmaillard:
Fixes #393


.changeset/serious-cougars-film.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cube-creator/core-api": patch
+"@cube-creator/model": patch
+"@cube-creator/ui": patch
+---
+
+Redefine supported langauges and their order
```
tpluscode, line 7: Typo and propose to add issue numbers where relevant so that they appear in the changelogs

```suggestion
Redefine supported languages and their order (#393)
```

apis/core/bootstrap/shapes/dimension.ts
```diff
@@ -117,7 +118,7 @@ ${shape('dimension/metadata')} {
       ${sh.name} "Description" ;
       ${sh.path} ${schema.description} ;
       ${sh.datatype} ${rdf.langString} ;
-      ${sh.languageIn} ( "en" "de" "fr" "it" ) ;
+      ${sh.languageIn} ${supportedLanguagesRdf} ;
```
tpluscode, line 121: The additional export is unnecessary. You can [interpolate the array](https://runkit.com/embed/td1qgbq2rpju) directly. Just keep the parens

```suggestion
      ${sh.languageIn} ( ${supportedLanguages} ) ;
```

tpluscode (changes requested):

.changeset/serious-cougars-film.md
martinmaillard: 👍

apis/core/bootstrap/shapes/dimension.ts
martinmaillard: Isn't it missing the quotes around each value?

tpluscode: Ah, my bad 😞

packages/model/languages.ts
```diff
@@ -0,0 +1,3 @@
+
+export const supportedLanguages = ['de', 'fr', 'it', 'rm', 'en']
+export const supportedLanguagesRdf = `(${supportedLanguages.map(lang => `"${lang}"`).join(' ')})`
```
tpluscode, line 3: It appears that `supportedLanguagesRdf` is used more actually So how about the opposite? Only export RDF/JS literals and map them to their `.value` in the vue component?

```suggestion
import { literal } from '@rdf-esm/data-model'

export const supportedLanguages = [ literal('de'), literal('fr'), literal('it'), literal('rm'), literal('en') ]
```

And if anything, I would not have the export what appears to be hardcoded RDF List in turtle ;)

martinmaillard: Alright, I don't mind...

tpluscode (approved):

l00mi (approved):

PR #516, opened 2021-02-10
fix: add placeholder value to mapped form, ignore on save

tpluscode:
Added a default value (had to be URI) to the mapping pairs. It will be ignored when saving unchanged


martinmaillard (approved):
Can we merge that in your branch instead, so that we don't have 10 places to track changes?

tpluscode:
Ok, pushed there. Closing this

PR #529, opened 2021-02-15
UI to map dimension to managed dimension

martinmaillard:
It's only a first draft of the feature, but it should already be mostly functional.


.changeset/ten-carpets-think.md
```diff
@@ -1,7 +1,5 @@
 ---
-"@cube-creator/core-api": patch
```
martinmaillard: Doesn't that mean that it won't create a new version for the API?

tpluscode: From this PR alone yes, but the APIs got their own changesets already

tpluscode (approved):

PR #530, opened 2021-02-15
docs: managed dimensions readme

tpluscode:


apis/managed-dimensions/README.md
```diff
@@ -0,0 +1,54 @@
+# Cube Curator Managed Dimensions API
+
+This Hydra-driven API provides functionality for creating
```
martinmaillard, line 3: missing some words ;)

apis/managed-dimensions/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## RDF namespaces
+
+| `md` | `https://cube.link/meta/` |
+
+## Resource organization
+
+The API provides two distinct of functionalities:
+
+1. Retrieving Shared Dimensions from the store
```
martinmaillard, line 17: We seem to be using both "Shared Dimension" and "Managed Dimension". Is there a difference? If not, we should probably settle on one. I have a slight preference for "Shared Dimension".

tpluscode: That's a good observation. @l00mi proposed to use the `meta:SharedDimension` as a more general term for anything that we can map to from cube creator. Shared Dimensions can be created by third parties and the API will find them in Lindas, in any graph

That leaves Managed Dimensions as those which will be created and managed directly from the app.

We could call all of them Shared Dimensions...

martinmaillard: I see, so "Managed Dimensions" are "Shared Dimensions managed in cube-creator". To me it would make sense to only use "Shared Dimension", but it would be good to wait for @l00mi's opinion before we change everything.

tpluscode: ping @l00mi

l00mi: As discussed, lets change everything to "Shared Dimensions".

tpluscode:
Ok, so I obliterated the term "Managed Dimension" every where in code except the ENV variables...

l00mi:
impressive, now i need to do it on our board i guess (-:

PR #536, opened 2021-02-17
API for creating Shared Dimensions (term set)

tpluscode:


apis/managed-dimensions/hydra/index.ttl
```diff
@@ -34,7 +36,33 @@ md:ManagedDimensions
...
+         hydra:expects <shape/managed-dimension> ;
+         code:implementedBy
+           [
+             a code:EcmaScript ;
+             code:link <file:handlers/managed-dimensions#post> ;
+           ]
+       ]
+.
+
+<shape/managed-dimension> a sh:NodeShape .
```
martinmaillard, line 52:
```suggestion
<shape/managed-dimension> a sh:Shape .
```

apis/managed-dimensions/lib/shapes/managed-dimension.ts
```diff
@@ -0,0 +1,87 @@
...
+import { supportedLanguages } from '@cube-creator/core/languages'
+import type { Initializer } from '@tpluscode/rdfine/RdfResource'
+import type { NodeShape } from '@rdfine/shacl'
+import $rdf from 'rdf-ext'
+import { meta, sh1 } from '@cube-creator/core/namespace'
+
+export default (): Initializer<NodeShape> => ({
+  closed: true,
+  property: [{
+    path: schema.name,
```
martinmaillard, line 11:
```suggestion
    name: 'Name',
    path: schema.name,
```

apis/managed-dimensions/lib/shapes/managed-dimension.ts
```diff
@@ -0,0 +1,87 @@
...
+    uniqueLang: true,
+    minCount: 1,
+    maxCount: 4,
+    languageIn: supportedLanguages,
+    order: 10,
+  }, {
+    path: schema.validFrom,
```
martinmaillard, line 18:
```suggestion
    name: 'Valid from',
    path: schema.validFrom,
```

apis/managed-dimensions/lib/shapes/managed-dimension.ts
```diff
@@ -0,0 +1,87 @@
...
+  }, {
+    path: sh.property,
+    minCount: 1,
+    maxCount: 1,
+    nodeKind: sh.BlankNode,
+    defaultValue: $rdf.blankNode(),
+    order: 30,
+    node: {
+      property: [{
+        path: schema.name,
```
martinmaillard, line 32:
```suggestion
        name: 'Name',
        path: schema.name,
```

apis/managed-dimensions/lib/shapes/managed-dimension.ts
```diff
@@ -0,0 +1,87 @@
...
+        uniqueLang: true,
+        maxCount: 4,
+        languageIn: supportedLanguages,
+        order: 10,
+      }, {
+        path: qudt.scaleType,
```
martinmaillard, line 38:
```suggestion
        name: 'Scale type',
        path: qudt.scaleType,
```

apis/managed-dimensions/lib/shapes/managed-dimension.ts
```diff
@@ -0,0 +1,87 @@
+import { dash, qudt, rdf, schema, sh, time, xsd } from '@tpluscode/rdf-ns-builders'
+import { supportedLanguages } from '@cube-creator/core/languages'
+import type { Initializer } from '@tpluscode/rdfine/RdfResource'
+import type { NodeShape } from '@rdfine/shacl'
+import $rdf from 'rdf-ext'
+import { meta, sh1 } from '@cube-creator/core/namespace'
+
+export default (): Initializer<NodeShape> => ({
```
martinmaillard, line 8: May I ask why you used a different way to declare the shape than in the main API?

tpluscode: I decided not to put the shapes in the store, because we only have that single graph and that would be rather difficult to maintain.

So since they are initialised in code, it's easier to create them using JS rather turtle and parse

apis/managed-dimensions/lib/shapes/managed-dimension.ts
```diff
@@ -0,0 +1,91 @@
...
+    order: 10,
+  }, {
+    name: 'Valid from',
+    path: schema.validFrom,
+    maxCount: 1,
+    datatype: xsd.dateTime,
+    defaultValue: $rdf.literal(new Date().toISOString(), xsd.dateTime),
+    order: 20,
+  }, {
+    path: sh.property,
```
martinmaillard, line 26: Any way we can avoid a nested form here? If not, I would add something like `name: ' '` to force it not to display `sh:property` as label.

martinmaillard (approved):

apis/managed-dimensions/lib/shapes/managed-dimension.ts
tpluscode: I actually added a label/description which this deserves.

Cannot avoid nesting unless we try to support complex `sh:path`

PR #537, opened 2021-02-17
fix: allow unmapping dimension

martinmaillard:
Removing the mapped dimension currently causes an error. This PR fixes that.

### DIFF
--- a/apis/core/lib/domain/dimension-mapping/update.ts
+++ b/apis/core/lib/domain/dimension-mapping/update.ts
@@ -29,7 +29,7 @@ export async function update({
     throw new error.BadRequest('Unexpected value of schema:about')
   }
 
-  if (!managedDimension.equals(dimensionMappings.managedDimension)) {
+  if (!managedDimension || !managedDimension.equals(dimensionMappings.managedDimension)) {
     dimensionMappings.changeManageDimension(managedDimension)
   }
 
--- a/apis/core/test/domain/dimension-mapping/update.test.ts
+++ b/apis/core/test/domain/dimension-mapping/update.test.ts
@@ -411,7 +411,7 @@ describe('domain/dimension-mapping/update', () => {
   })
 
   describe('changing managed dimension', () => {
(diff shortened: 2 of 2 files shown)


.changeset/fast-cars-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cube-creator/core-api": patch
+---
+
+Fix removing link between dimension and managed dimension
```
tpluscode, line 5: I would remove this time. The whole feature has been unreleased so far, so there is technically no bug to fix

apis/core/lib/domain/dimension-mapping/update.ts
```diff
@@ -29,7 +29,7 @@ export async function update({
     throw new error.BadRequest('Unexpected value of schema:about')
   }
 
-  if (!managedDimension.equals(dimensionMappings.managedDimension)) {
+  if (!managedDimension || !managedDimension.equals(dimensionMappings.managedDimension)) {
```
tpluscode, line 32: Worthy of adding a unit test?

.changeset/fast-cars-talk.md
martinmaillard: Yeah... that's why I'm not a fan of semi-automated changelogs ;)
Will remove

apis/core/lib/domain/dimension-mapping/update.ts
martinmaillard: Probably yes. I tried and then got lazy. Is there a reason why the "when" of these unit tests is in the `beforeEach`? That seems really wrong and makes it hard to add new test cases in the same `describe`.

tpluscode: Yea, that doesn't make sense. Especially that there is only one test case in that `describe`.

Definitely move the `await update()` inside the `it()`

tpluscode (approved):

PR #542, opened 2021-02-23
Disable OIDC session monitoring

### LINKED ISSUE #480: "User is not defined" when creating a project
I got this Error Message this morning when trying to creat a new project

GitHubIssue_480

- Happened after 1st logon to Cube Creator (INTE)
- Gets fixed by logging off and on again
- Had the same issue already last week but wasn't sure if that was just random

sandhose:
The session monitoring sometime triggered a race condition where the auth server notified a session change, which made the OIDC client silently re-check the session, but with the wrong scopes.

The hardcoded scope can be seen here: https://github.com/IdentityModel/oidc-client-js/blob/e7093cf17cfa006036066bcae27f7728fafe8419/src/UserManager.js#L334

This disables the session monitoring. It was there to trigger app logouts when the user logs out from the auth server but not from within the app. I think we can live without that :)

This should fix #480

### DIFF
--- a/ui/src/store/modules/auth.ts
+++ b/ui/src/store/modules/auth.ts
@@ -14,14 +14,15 @@ const cookieStorage = new CookieStorage({
   sameSite: 'Strict',
 })
 
-const oidcSettings: VuexOidcClientSettings = {
+const oidcSettings = {
   redirectUri: new URL('oidc-callback', base).toString(),
   postLogoutRedirectUri: new URL('logout', base).toString(),
   responseType: 'code',
   automaticSilentRenew: true,
   stateStore: new WebStorageStateStore({ store: cookieStorage }),
+  monitorSession: false,
   ...oidc,
-}
+} as VuexOidcClientSettings
 
 export function auth (): Module<VuexOidcState, RootState> {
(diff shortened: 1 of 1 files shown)


ui/src/store/modules/auth.ts
```diff
@@ -14,14 +14,15 @@ const cookieStorage = new CookieStorage({
...
+const oidcSettings = {
   redirectUri: new URL('oidc-callback', base).toString(),
   postLogoutRedirectUri: new URL('logout', base).toString(),
   responseType: 'code',
   automaticSilentRenew: true,
   stateStore: new WebStorageStateStore({ store: cookieStorage }),
+  monitorSession: false,
   ...oidc,
-}
+} as VuexOidcClientSettings
```
sandhose, line 25: This is necessary because of bad typing in `vuex-oidc`. `VuexOidcClientSettings` extends from `OidcClientSettings`, while it should extend from `UserManagerSettings`.

martinmaillard (approved):
I don't know how you figured that out but... good job 😅

tpluscode:
I think this deserves changelog entry....

PR #546, opened 2021-02-25
Create term

tpluscode:


martinmaillard:
I think we are missing some code to expose the new shapes through the API. I got a 404 and solved it by adding stuff in `bootstrap/shapes`.

tpluscode:
You think we're good to merge here? @martinmaillard

martinmaillard (approved):

PR #555, opened 2021-03-01
Shared dimensions manager UI

martinmaillard:


martinmaillard:
@tpluscode One feature I'm missing in the API is the collection of terms of a dimenison

ui/src/store/modules/app.ts
```diff
@@ -9,12 +9,14 @@ export interface Message {
...
 
 export interface AppState {
+  language: string
   messages: Message[]
   loading: boolean
   commonRDFProperties: string[]
 }
 
 const initialState = {
+  language: 'en',
```
tpluscode, line 19: Might try [navigator.language](https://caniuse.com/mdn-api_navigatorlanguage_language) to hope for user's agent language

ui/src/store/types.ts
```diff
@@ -1,13 +1,40 @@
...
+}
+
+export interface SharedDimension extends Resource {
+  name?: string
+  terms?: Term
+}
+
+export interface SharedDimensionTerm extends Resource {
+  name: Term[]
+  identifiers: string[]
```
tpluscode, line 39: Does the `Resource` interface essentially duplicate `import { RdfResource } from '@tpluscode/rdfine'` and could be substituted?

ui/src/store/modules/app.ts
martinmaillard: I don't think that's worth it, since the whole app is in english anyway

ui/src/store/types.ts
martinmaillard: Kinda. It's technically a subset of `RdfResource`. My types are a bit broken because I'm using the types from `@cube-creator/model` when they exist, but my serializers don't really respect them. This `Resource` is actually more correct.

tpluscode (approved):

PR #558, opened 2021-03-02
feat(shared-dimensions): add link to dimension's terms

tpluscode:


martinmaillard:
The `md:terms` link is added when requesting all the dimensions, but not when requesting a single dimension... and that's when I need it the most 😄

tpluscode:
Added too @martinmaillard

martinmaillard (approved):

PR #571, opened 2021-03-04
fix(cli): do not generated sh:in for xsd:int dimension

### LINKED ISSUE #570: Dimensions with "int" datatype are missing min/maxInclusive in shape
Instead they have a list of values in the `in` constraint.

See https://s.zazuko.com/uNnLu

I would expect them to behave the same as the "decimal" dimensions.

Sidenote: The dimension has datatype "http://www.w3.org/2001/XMLSchema#int" but the individual values have "http://www.w3.org/2001/XMLSchema#integer" when queried with https://github.com/zazuko/rdf-cube-view-query

tpluscode:
closes #570

### DIFF
--- a/cli/lib/toCubeShape/Dimension.js
+++ b/cli/lib/toCubeShape/Dimension.js
@@ -14,7 +14,7 @@ const datatypeParsers = (datatype) => {
     case ns.xsd.integer.value:
     case ns.xsd.int.value:
     case ns.xsd.gYear.value:
-      return term => parseInt(term.value)
+      return term => parseInt(term.value, 10)
   }
 }
 
--- a/cli/lib/toCubeShape/Dimension.js
+++ b/cli/lib/toCubeShape/Dimension.js
@@ -1,17 +1,22 @@
 const clownface = require('clownface')
 const rdf = require('rdf-ext')
-const TermMap = require('@rdfjs/term-map')
 const TermSet = require('@rdfjs/term-set')
 const ns = require('@tpluscode/rdf-ns-builders')
 
-const datatypeParsers = new TermMap([
-  [ns.xsd.date, term => new Date(term.value)],
-  [ns.xsd.double, term => parseFloat(term.value)],
-  [ns.xsd.float, term => parseFloat(term.value)],
-  [ns.xsd.integer, term => parseInt(term.value)],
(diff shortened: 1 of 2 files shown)


cli/lib/toCubeShape/Dimension.js
```diff
@@ -1,26 +1,31 @@
...
+    case ns.xsd.double.value:
+    case ns.xsd.float.value:
+    case ns.xsd.decimal.value:
+      return term => parseFloat(term.value)
+    case ns.xsd.integer.value:
+    case ns.xsd.int.value:
+    case ns.xsd.gYear.value:
+      return term => parseInt(term.value)
+  }
+}
```
tpluscode, line 19: I suppose implemented like this is a little easier on the eye and similar in performance. Could improve by memorising the individual cases or changing them into contents.

By the way, do we want to include other `xsd:integer` subtypes and `xsd:dateTime` here too? @bergos

cli/lib/toCubeShape/Dimension.js
```diff
@@ -1,26 +1,31 @@
...
+    case ns.xsd.date.value:
+      return term => new Date(term.value)
+    case ns.xsd.double.value:
+    case ns.xsd.float.value:
+    case ns.xsd.decimal.value:
+      return term => parseFloat(term.value)
+    case ns.xsd.integer.value:
+    case ns.xsd.int.value:
+    case ns.xsd.gYear.value:
+      return term => parseInt(term.value)
```
herrstucki, line 17: @tpluscode Not sure if RDF integers/years can start with 0 but just in case they can, it's usually safer to specify the radix explicitly:

```suggestion
      return term => parseInt(term.value, 10)
```

tpluscode: When will it make a difference?

herrstucki: It used to be a problem because JS would interpret something like `parseInt("010")` as octal instead of decimal. But actually I'm not sure if this is still the case.

https://stackoverflow.com/questions/6611824/why-do-we-need-to-use-radix-parameter-when-calling-parseint

tpluscode: I see. MDN mentions that this should be the case when `radix = 0` but it won't hurt to make it explicit an be sure 👌

PR #577, opened 2021-03-09
fix: shared dimension idnetifier

tpluscode:
Add `dcterms:identifier` to be used to generate term (set) URIs. (only lowercase, number and hyphen)

Will respond 409 if another term (Set) with the generated URI already exists.

Is it a little weird to have both `dcterms:identfier` and `schema:identifier`? 🤷


martinmaillard:
Having 2 fields called "Identifier" will be *very* confusing for users.

tpluscode:
The `dcterms:identifier` is only used for creating so it might be a labeling matter only...

PR #584, opened 2021-03-09
Improve cube preview values display

martinmaillard:
- Only display numerical terms as monospace
- Underline temporal terms in yellow. That's a bit arbitrary but we talked about this a long time ago and I couldn't come up with a better option.


tpluscode (approved):
LGTM. @l00mi ?

l00mi (approved):

PR #607, opened 2021-03-16
chore: add more prefixes when editing raw RDF

martinmaillard:


ui/src/components/HydraRawRdfForm.vue
```diff
@@ -5,6 +5,7 @@
       v-if="shape"
       format="text/turtle"
       :quads.prop="initQuads"
+      :prefixes="editorPrefixes"
```
tpluscode, line 8: Does this vue stringify the array?

ui/src/components/HydraRawRdfForm.vue
```diff
@@ -52,6 +53,7 @@ export default class HydraRawRdfForm extends Vue {
 
   parseError: string | null = null
   editorQuads: Quad[] | null = null
+  editorPrefixes = ['hydra', 'rdf', 'rdfs', 'schema', 'xsd']
```
tpluscode, line 56: FYI `rdf(s)` and `xsd` are not necessary. Always added anyway

tpluscode (approved):

tpluscode:
On the other hand, I thought we could add a textbook/dropdown where more prefixes could be added as necessary

l00mi (approved):

l00mi:
@tpluscode lets KISS for now.

ui/src/components/HydraRawRdfForm.vue
martinmaillard: Probably not, why?

ui/src/components/HydraRawRdfForm.vue
martinmaillard: "Explicit is better than implicit"™ 😉

ui/src/components/HydraRawRdfForm.vue
tpluscode: https://zazuko.github.io/rdfjs-elements/rdf-editor/

> `prefixes`: a comma-separated list of prefixes to use for serializing.

martinmaillard: Funny, it seemed to work 🤔

tpluscode: Good, so maybe that is what vue does. Native `Array#toString()` simply comma-joins its elements 👌

PR #611, opened 2021-03-16
feat(ui): show "read-only" and "deprecated" tags on shared dimensions

martinmaillard:


l00mi:
Good one to think about Deprecated too! Kudos.

PR #612, opened 2021-03-16
fix(ui): stop polling jobs on API error

martinmaillard:
I don't really understand why the error never reaches `App.errorCaptured`, but it probably has to do with the fact that the code is async.

Anyway, this change will just stop polling jobs if any error occurs. This should solve our Sentry overflow issue.


sandhose (approved):
Looks alright, thanks!

PR #616, opened 2021-03-17
Ability to edit shared dimension or cube metadata as raw RDF

martinmaillard:
One issue for the cube metadata: the data contains all sorts of stuff that we shouldn't probably edit (resource labels, generated shape, etc.). Not sure what to do about that.

### DIFF
--- a/apis/shared-dimensions/lib/shapes/shared-dimension.ts
+++ b/apis/shared-dimensions/lib/shapes/shared-dimension.ts
@@ -160,7 +160,6 @@ export const create = (): Initializer<NodeShape> => ({
 })
 
 export const update = (): Initializer<NodeShape> => ({
-  closed: false,
   ignoredProperties: [
     dcterms.identifier,
     md.terms,
--- a/apis/shared-dimensions/lib/shapes/shared-dimension.ts
+++ b/apis/shared-dimensions/lib/shapes/shared-dimension.ts
@@ -160,7 +160,7 @@ export const create = (): Initializer<NodeShape> => ({
 })
 
 export const update = (): Initializer<NodeShape> => ({
-  closed: true,
+  closed: false,
   ignoredProperties: [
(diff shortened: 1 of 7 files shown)


apis/shared-dimensions/lib/shapes/shared-dimension.ts
```diff
@@ -160,7 +160,7 @@ export const create = (): Initializer<NodeShape> => ({
 })
 
 export const update = (): Initializer<NodeShape> => ({
-  closed: true,
+  closed: false,
```
tpluscode, line 163:
```suggestion
```

ui/src/components/HydraOperationFormWithRaw.vue
```diff
@@ -0,0 +1,93 @@
...
+      const rdfEditor: HydraRawRdfForm = this.$refs.rdfEditor as any
+      const resource = Object.freeze(clownface({
+        dataset: $rdf.dataset(rdfEditor.editorQuads || []),
+        term: this.resource.term,
+      }))
+      this.$emit('sync-resource', resource)
+    } else {
+      const form: HydraOperationForm = this.$refs.form as any
+      const resource = Object.freeze(form.clone)
+      this.$emit('sync-resource', resource)
```
tpluscode, line 77: Does this have to be emitted and handled by the parent?

Maybe it could be synced internally inside `HydraOperationFormWithRaw`

tpluscode (changes requested):

ui/src/components/HydraOperationFormWithRaw.vue
martinmaillard: A component should never mutate its props.

There may be a way to improve this, but didn't seem worth it

l00mi (approved):

tpluscode (approved):

PR #627, opened 2021-03-22
build: pin versions to have them bumped on upstream changes

tpluscode:
Re: https://github.com/atlassian/changesets/issues/544

This ensures that version of dependent packages are bumped when pinned dependencies are updated

### DIFF
--- a/apis/core/package.json
+++ b/apis/core/package.json
@@ -7,9 +7,9 @@
   },
   "dependencies": {
     "@cube-creator/api-errors": "0.0.1",
-    "@cube-creator/core": "^0.3.0",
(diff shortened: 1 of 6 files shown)


tpluscode:
Specifically it will see that `core-api` is bumped (and deployed) when `shared-dimensions-api` gets a new version

martinmaillard:
Does that also mean we have to manually upgrade all the dependant packages when a package gets a new version?

martinmaillard (approved):

tpluscode:
No. The thing is that with a `^x.y.z` range it does not get bumped if the updated dependency still matches that range. Fixed version will always be updated

PR #668, opened 2021-04-08
feat(cli): add schema:sameAs to versioned concepts

### LINKED ISSUE #666: Add a non-versioned IRI for every non-observation table
For the "concept" tables we need to have a version independent way to filter. To achieve this add aside the versioned IRI also a IRI without the version.

Further add a triple to connect the versioned to it with `https://schema.org/sameAs`.

Finally add the unique identifier part as a literal to the non versioned IRI per `https://schema.org/identifier`.

tpluscode:
fixes #666

### DIFF
--- a/cli/lib/cube.ts
+++ b/cli/lib/cube.ts
@@ -3,9 +3,10 @@ import { DatasetCore, Quad, Term, Quad_Subject as QuadSubject } from 'rdf-js'
 import { cc, cube } from '@cube-creator/core/namespace'
 import { GraphPointer } from 'clownface'
 import type { Context } from 'barnard59-core/lib/Pipeline'
-import map from 'barnard59-base/lib/map'
+import { obj } from 'through2'
 import TermMap from '@rdfjs/term-map'
-import { rdf, sh } from '@tpluscode/rdf-ns-builders'
+import { rdf, schema, sh } from '@tpluscode/rdf-ns-builders'
+import { loadDataset } from './metadata'
 
 export const getObservationSetId = ({ dataset }: { dataset: DatasetCore }) => {
   const cubeId = [...dataset.match(null, cc.cube)][0].object.value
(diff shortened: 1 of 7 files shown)


fuseki/sample-ubd.trig
```diff
@@ -197,6 +198,11 @@ graph <cube-project/ubd/cube-data> {
     <https://environment.ld.admin.ch/foen/ubd/28/unit>            <http://qudt.org/vocab/unit/MicroGM-PER-M3> ;
   .
 
+  <https://environment.ld.admin.ch/foen/ubd/28/station/blBAS>
+    schema:identifier "blBAS" ;
+    vaem:owner <https://environment.ld.admin.ch/foen/ubd/28/maintainer/blBAS> ;
```
tpluscode, line 203: This is not mapped in transformation but I added so that we can test that only objects of observation dimensions are considered for the `schema:sameAs` treatment

cli/lib/cube.ts
```diff
@@ -41,7 +43,11 @@ export function injectRevision(this: Pick<Context, 'variables' | 'log'>) {
     return term
   }
 
-  return map(({ subject, predicate, object, graph }: Quad) => {
+  function isDimension(predicate: Term) {
+    return dataset.pointer.any().has(schema.about, predicate).terms.length
+  }
```
tpluscode, line 48: This way only predicates which are dimensions are considered for `schema:sameAs`

@l00mi do we want any additional constraint here?

martinmaillard (approved):

l00mi:
please wait before merging

l00mi (approved):

cli/lib/cube.ts
```diff
@@ -41,7 +47,15 @@ export function injectRevision(this: Pick<Context, 'variables' | 'log'>) {
     return term
   }
 
-  return map(({ subject, predicate, object, graph }: Quad) => {
+  function shouldAddSameAs(predicate: Term) {
+    if (!dataset) {
+      return false
+    }
+
+    return dataset.pointer.any().has(schema.about, predicate).terms.length > 0
```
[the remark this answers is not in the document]
tpluscode: Meant to only check whether the `predicate` is a dimension.

Issue #672, opened 2021-04-12
Add the cube:KeyDimension cube:MeasurementDimension to the rdf-cube-schema spec.


l00mi:

l00mi:
ready https://github.com/zazuko/rdf-cube-schema/pull/18

PR #673, opened 2021-04-13
Dimension type

martinmaillard:
#672 

Adds dimension type on dimension metadata shape + changes cube preview based on that.


tpluscode (approved):
LGTM

PR #676, opened 2021-04-13
feat(api): dimension kind selected on mapping

### LINKED ISSUE #670: Give an option to specify Key and Measurent Attribute already on column mapping.

tpluscode:
Adds checkboxes for key/measure dimensions and addition `sh:xone` to prevent both being set to `true`

closes #670, but first it would need to also update the reference mapping form to include said checkboxes


ui/src/components/ReferenceColumnMappingForm.vue
```diff
@@ -109,12 +120,27 @@ export default class extends Vue {
...
+
+    this.shape = await api.fetchOperationShape(this.operation)
+  }
+
+  get dimensionTypes (): MultiPointer {
+    const emptyPointer = clownface({ dataset: $rdf.dataset(), term: [] })
+
+    if (!this.shape) return emptyPointer
+
+    const nodes = this.shape.pointer.out(sh.property).has(sh.path, [cc.dimensionType]).out(sh.in).list()
```
tpluscode, line 139:
```suggestion
  get dimensionTypes (): AnyPointer {
    const emptyPointer = clownface({ dataset: $rdf.dataset() })

    if (!this.shape) return emptyPointer

    const nodes: any = this.shape.pointer.out(sh.property).has(sh.path, [cc.dimensionType]).out(sh.in).list()
```

PR #684, opened 2021-04-20
fix(cli): wrong type cube:SharedDimension

tpluscode:


cli/lib/metadata.ts
```diff
@@ -87,21 +88,24 @@ export async function injectMetadata(this: Context, jobUri: string) {
...
 
-      [...datasetTriples.match(null, schema.about, quad.object)].forEach(dim => {
-        [...datasetTriples.match(dim.subject)]
-          .filter(c => !c.predicate.equals(schema.about))
-          .forEach(meta => {
-            this.push($rdf.triple(propertyShape, meta.predicate, meta.object))
-            visited.add(propertyShape)
-            copyChildren(meta.object)
+      for (const dim of [...datasetTriples.match(null, schema.about, quad.object)]) {
+        for (const meta of [...datasetTriples.match(dim.subject)]
```
martinmaillard, line 94: I like the switch to `for` in this case, but I would have extracted the `[...` part in a variable to make it more readable.

That's just a nitpick though, do what you want :)

martinmaillard (approved):

l00mi (approved):

Issue #651, opened 2021-03-29
Additional information to show in the Publication List


l00mi:
- Last upload of a CSV (date and file name)
- Last modification of mapping (general and for each table)
- Last modification of Cube design

l00mi:
@ortnever please add the description here.

ortnever:
Here the information that should be displayed about the publication, in addition to the date
- cube version (will be taken over by visualize)
- status Publish / Draft
- where the cube is published. For the moment: "opendata.swiss" and/or "visualize".

Information about other stages of the work (not for the Publication List). 
- Last upload of a CSV (date and file name)
- Last modification of mapping (general and for each table)
- Last modification of Cube design

l00mi:
@tpluscode can you please extend the publication list with the upper part. That we see in the list of publications what was the metadata of the publication run. I guess you do not have any structure to describe that yet?

Can we add as a metadata triple the JobId from gitlab? So we can query directly on Lindas to get this infos?

l00mi:
I move this ticket back to back-log for the still open parts, which are less prio:

- Last upload of a CSV (date and file name)
- Last modification of mapping (general and for each table)
- Last modification of Cube design

ortnever:
Parfait. Merci.

ortnever:
@Michael, @pierre : We should discuss next Wednesday if the second part of this issue can be done in the next sprint.

ortnever:
oups, I guess i didn't use the good Email-Adresse, Sorry Michael and Pierre !

l00mi:
@ortnever nono all good lets discuss wednesday.

PR #687, opened 2021-04-22
feat: more details on job list

tpluscode:
Extends the details of publish jobs:

- version (will show for existing jobs)
- status
- "publish to"
- link to a Lindas query (only when job succeeds)

"status" and "publish to" actually dereference to show `schema:name`

The Lindas query is indirect, searching by the parent cube and version rather than concrete URI. Like below:

```sparql
describe ?cube
from <https://lindas.admin.ch/foen/cube>
where {
  <https://environment.ld.admin.ch/foen/faostat-prod_test001> schema:hasPart ?cube .
  ?cube schema:version 4 .
}
```

Partially addresses #651

### DIFF
--- a/apis/core/lib/domain/job/update.ts
+++ b/apis/core/lib/domain/job/update.ts
@@ -3,12 +3,32 @@ import { GraphPointer } from 'clownface'
 import { Job, JobMixin, Project, Dataset } from '@cube-creator/model'
 import { isPublishJob } from '@cube-creator/model/Job'
 import RdfResource from '@tpluscode/rdfine'
+import env from '@cube-creator/core/env'
+import { DESCRIBE } from '@tpluscode/sparql-builder'
+import type { Organization } from '@rdfine/schema'
 import { ResourceStore } from '../../ResourceStore'
 import { schema } from '@tpluscode/rdf-ns-builders'
 
 interface JobUpdateParams {
   resource: GraphPointer<NamedNode>
   store: ResourceStore
+}
+
+function lindasWebQueryLink(cube: NamedNode, version: number, cubeGraph: NamedNode) {
+  const describe = DESCRIBE`?cube`
+    .FROM(cubeGraph)
+    .WHERE`
+      ${cube} ${schema.hasPart} ?cube .
+      ?cube ${schema.version} ${version} .
+    `.build()
+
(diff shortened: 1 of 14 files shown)


apis/core/lib/domain/job/update.ts
```diff
@@ -11,6 +14,23 @@ interface JobUpdateParams {
...
 
+function lindasWebQueryLink(cube: NamedNode, version: number, cubeGraph: NamedNode) {
+  const describe = DESCRIBE`?cube`
+    .FROM(cubeGraph)
+    .WHERE`
+      ${cube} ${schema.hasPart} ?cube .
+      ?cube ${schema.version} ${version} .
+    `.build()
+
+  const queryLink = new URL('https://lindas.admin.ch/sparql')
```
tpluscode, line 25: Put the URL in config variable?

ui/src/components/ExternalResource.vue
```diff
@@ -0,0 +1,35 @@
...
+  @Prop({ required: true }) resourceId?: Term | undefined
+
+  name = ''
+  title = ''
+
+  async mounted () {
+    if (this.resourceId?.termType === 'NamedNode') {
+      const { representation } = await Hydra.loadResource(this.resourceId.value)
+      if (representation?.root) {
+        [this.name] = representation.root.pointer.out(schema.name, { language: ['en', '*'] }).values
```
tpluscode, line 24: @martinmaillard the UI language is only English right now?

apis/core/lib/domain/job/update.ts
l00mi: We should discuss this once in more detail. I also see that we should migrate some of these fields into the "Organisation Profile" (currently only BAFU). Can we open an issue and just make a pointer here for now?

tpluscode: This entire `DESCRIBE` query is very specific. Maybe the most Lindas-specific piece of code we have

l00mi (approved):

apis/core/lib/domain/job/update.ts
tpluscode: Extracted into a configuration variable `TRIFID_UI`

Issue #689, opened 2021-04-23
Designer: Show labels for shared terms


tpluscode:
Right now the cube preview only shows `schema:name` of terms generated by the transformation pipeline. A shared dimension term will always show as the URI and needs a click to load the preview panel

A proposed improvement is to try dereferencing URIs if no label is found in the cube query response.

If dereferenced but on `schema:name` in the selected language is found - keep the URL.

l00mi:
Its perfect, check with @martinmaillard who is better to take a look at it.

PR #698, opened 2021-05-03
feat: more ways to get resource details

tpluscode:
I merged the `ExternalResource` component I added earlier to the publish job list with `TermDisplay`. It will now make multiple attempts at fetching more triples about resources (in that order):

1. Display `schema:name` if already found
1. Display a prefixed name for vocals known to `@zazuko/rdf-vocabularies`
2. Use cube-creator API to look up cube data
4. Run `DESCRIBE` on Lindas endpoint provided by API
3. Dereference (only if HTTPS)
5. Try to prefix using cube URI as base 
6. Last resort, display full value


l00mi:
That is a first step for #689 ?

tpluscode:
What is the second step then?

l00mi:
> What is the second step then?

At least here: https://pipelines-test.lindas.admin.ch/app/cube-projects/cube-project!!martin-test-yrl7gb4q64b/metadata I expect to see the Label "Aluminium" if German is selected, and not "https://ld.admin.ch/cube/dimension/metals/al"

l00mi:
Ah its not yet merged.. okay the lets see if merged.

l00mi (approved):

l00mi:
@martinmaillard can you have also a look, especially in regards of the "label" strategy in the cube-viewer.

martinmaillard (commented):
I'm not a fan of this change. If I understand correctly, it will do a separate HTTP request for all the resources displayed on screen (which can be a lot in case of the Cube Preview).

I also don't like the idea of making `TermDisplay` (supposed to "display a term") transparently doing multiple HTTP requests anywhere in the UI. I would at least keep this behavior in a separate component that we can explicitly use where we really need it.

ui/src/views/CubeProject.vue
```diff
@@ -86,7 +86,7 @@ export default class CubeProjectView extends Vue {
 
   async pollJobs (): Promise<any> {
     try {
-      await this.$store.dispatch('project/fetchJobCollection')
+      // await this.$store.dispatch('project/fetchJobCollection')
```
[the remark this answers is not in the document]
tpluscode: Uh, this was bound to happen at some point 🙄 . I don't like how the polling clutters the browser network tab and comment this out to turn it off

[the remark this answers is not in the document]
martinmaillard: Yeah, same here. I don't think we're going to setup websockets now but that would be an option at some point.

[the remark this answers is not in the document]
tpluscode: I was also thinking of a dev-time ENV switch to simply have it disabled locally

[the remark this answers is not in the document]
martinmaillard: Also an option :)

l00mi:
Is it possible to use an ordered list of endpoints for 5. ? Or is this complex? If yes, I would propose to add `https://query.wikidata.org/sparql` here too for the time being?

ui/src/components/TermDisplay.vue
```diff
@@ -8,30 +8,68 @@
...
+    return shrunk !== this.node.value ? shrunk : null
+  }
+
+  get __schemaName (): string | null {
+    const pointer = '_context' in this.term
+      ? this.term
+      : null
+
+    if (pointer) {
+      return pointer.out([schema.name, rdfs.label], { language: ['en', '*'] }).values[0]
```
martinmaillard, line 61: I still have the same issue: `TermDisplay` was never supposed to display a label for a resource. It's supposed to display a term. I think it would be better to not touch it and implement what you need in `ExternalTerm`.

ui/src/components/CubePreviewValue.vue
```diff
@@ -4,7 +4,7 @@
   </span>
   <router-link v-else-if="isResource" :to="{ name: 'ResourcePreview', params: { resourceId: value.id.value, sharedTerm: isSharedTerm } }" class="tag is-rounded is-small">
     <term-with-language :values="label" :selected-language="selectedLanguage">
-      <term-display :term="value.id" :base="cubeUri" />
+      <external-term :resource="value" :base="cubeUri" />
```
martinmaillard, line 7: This is a bit weird.

Here `term-label` is used as a fallback for when `label` has no value for the provided language. `external-term` should probably replace `term-with-language` in this case.

tpluscode: Ok, I think I gotcha. Moved the label handling to `<external-term>` and also compose the `term-with-language`

martinmaillard (approved):

ui/src/views/ResourcePreview.vue
```diff
@@ -27,7 +27,7 @@
 <script lang="ts">
 import { Vue, Component } from 'vue-property-decorator'
 import { namespace } from 'vuex-class'
-import $rdf from '@rdfjs/data-model'
+import { namedNode } from '@rdfjs/data-model'
```
martinmaillard, line 30: I'm curious, does this change have an effect on the amount of code that is imported or is it just a cosmetic preference of yours?

tpluscode: Neither really. I followed the warning which said that the module does not have a default export.

Curiously, I actually checked now and it turns out that the typings inside this package are wrong. Bottom line this makes no difference 🙄

PR #706, opened 2021-05-07
Import cube SHACL shapes and models

tpluscode:
* [x] Use `sh:xone` to have one shape describe both kinds of project
* [ ] Make the forms dynamically hide fields for CSV/Import

This requires a data update spa that we can actually tell the two kinds of projects apart. I propose to keep the "project kind" but could try with `rdf:type` too


ui/src/forms/FormProperty.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <label v-if="property && !isHidden" class="form-property">
+  <label class="form-property">
```
tpluscode, line 2: Not necessary, since Shaperone now supports the `dash:hidden` as well as explicit hide/show feature, used fox alternating `xone` visibility

packages/model/Project.ts
```diff
@@ -22,19 +22,33 @@ export interface Project extends RdfResource {
...
+}
+
+export interface CsvProject extends Project {
+  cubeIdentifier: string
+}
+
+export interface ImportProject extends Project {
+  importCube: NamedNode
+  importFromGraph: NamedNode | undefined
+  importFromEndpoint: NamedNode
```
tpluscode, line 37: Don't particularly like the names. Maybe `sourceCube` etc?

packages/model/Project.ts
```diff
@@ -22,19 +22,33 @@ export interface Project extends RdfResource {
...
 }
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface ProjectsCollection extends Collection<Project> {
+export interface ProjectsCollection extends Collection<CsvProject | ImportProject> {
 
 }
 
+export const isCsvProject = (project: CsvProject | ImportProject): project is CsvProject => {
+  return project.pointer.out(dcterms.identifier).terms.length > 0
```
tpluscode, line 46: Maybe this should instead check the `projectSourceKind`?

packages/core/namespace.ts
```diff
@@ -65,7 +65,10 @@ type CubeCreatorProperty =
   'sharedDimension' |
   'sharedDimensions' |
   'applyMappings' |
-  'dimensionType'
+  'dimensionType' |
+  'CubeProject/importCube' |
+  'CubeProject/importFromEndpoint' |
+  'CubeProject/importFromGraph'
```
tpluscode, line 71: Ditto

ui/src/store/serializers.ts
```diff
@@ -29,36 +27,10 @@ const displayLanguage = ['en', 'de', 'fr', '']
 export function serializeProjectsCollection (collection: ProjectsCollection): ProjectsCollection {
   return Object.freeze({
     ...serializeResource(collection),
-    member: collection.member.map(serializeProject),
+    member: collection.member,
```
martinmaillard, line 30: Any reason for this change? It breaks the projects list

packages/model/Project.ts
martinmaillard: I like `sourceCube`, `sourceGraph` and `sourceEndpoint`

packages/model/Project.ts
martinmaillard: That would make sense

ui/src/store/serializers.ts
tpluscode: That's a bit of an overstatement. You mean the organisation name does not show?

martinmaillard: Well... yeah, it breaks it ;)

packages/model/Project.ts
martinmaillard: Project doesn't seem to have a `cc:projectSourceKind` in the frontend. I guess we could also check if the project has a `csvMapping`?

tpluscode: Is this about existing project? We did not persist that property until now. Make sure you run the SPARQL from the PR description

martinmaillard: Ah that's probably it then. Will do that, thanks

Issue #79, opened 2020-10-20
Import an existing cube


martinmaillard:
Start a new project from an existing cube

A lot of things need to be clarified. There are some infos in #11

lucafurrer:
I think #135 solves this as well, no?

martinmaillard:
What's missing to get this feature is the ability to import an existing cube. I'll rename the issue.

l00mi:
Pre-requirement: Cube produced by our pipeline.

Steps todo:
* Pipeline to copy the external cube: currently data.zazuko.com -> cube-creator on lindas
  * Make sure to not have cubes reimported in the same namespaces in a new project.
* We assume final URIs which are published.
* Metadata:
  * Copy already present metadata to the graph which is editable?
  * Publish is mostly the same.

l00mi:
Input RAW Cubes from PX for testing: https://github.com/zazuko/rdf-parser-px

l00mi:
@tpluscode and @martinmaillard please keep the overall effort below 8 days of work.

l00mi:
please update status every second week at https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/280

PR #707, opened 2021-05-11
feat: ImportJob class and API

tpluscode:
This adds another operation to the API, which creates a job of a new type `ImportJob` and triggers it to run

Part of #79


martinmaillard (approved):
Will the API conditionally provide `ImportAction` or `TransformAction` based on the type of project or is this the responsibility of the UI?

tpluscode:
Good point. I think we don't have a good way to improve that immediately

PR #714, opened 2021-05-12
Import pipeline

tpluscode:
* [x] import cube data
* [x] initialise `cc:DimensionMetadataCollection`
* [x] import metadata
* [x] ensure metadata is not overwritten on subsequent runs

**ad 1.** 

The cube observations and constraint shape are copied into `?project cc:cubeData ?graph`. This will always replace the entire contents of that graph

**ad 4.**

The last point is especially interesting. The import, of cube and dimension metadata alike, will not overwrite any existing values. New literals will be added if they did not previously exist for a given language. Otherwise, the existing values win, and new data from the source will be discarded.

**Limitations**

Blank nodes will be ignored in imported cube metadata. Only values which are direct properties of cube and dimensions will be imported


packages/testing/lib/seedData.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const oldCubeNs = /^http:\/\/ns\.bergnet\.org\/cube\//
+
+function rebaseCubeNs<T extends Term>(term: T): T {
+  if (term.termType === 'NamedNode') {
+    return $rdf.namedNode(term.value.replace(oldCubeNs, cube().value)) as any
+  }
+
+  return term
+}
```
tpluscode, line 51: To be removed when PX parser uses new cube schema

martinmaillard:
The import works well, but some metadata seems to be overriden on subsequent imports.
I added a name in english for a given dimension and it disappeared when re-importing. Same when I change an existing dimension name.

martinmaillard (approved):

Issue #731, opened 2021-05-21
Only english label of Shared Dimension is "resolved" in cube viewer


l00mi:
In https://pipelines-integ.lindas.admin.ch/app/cube-projects/cube-project!!ubd006-t3-k7orldamyq1/metadata/resource/https:%2F%2Fld.admin.ch%2Fcube%2Fdimension%2Fel01%2Fcr we see multiple labels also for FR and DE in the shared dimension, but the listing only works in EN.

l00mi:
<img width="2837" alt="Screenshot 2021-05-22 at 00 21 57" src="https://user-images.githubusercontent.com/502518/119203715-04991700-ba94-11eb-80a7-42e4fbfda08e.png">

tpluscode:
I think I noticed that and problem is a little more subtle. The language dropdown only changes the headers but not the data

However, if you change the page, the correct language will be displayed

PR #736, opened 2021-05-25
fix(ui): switch language for resource labels

martinmaillard:
Should fix https://github.com/zazuko/cube-creator/issues/731

### DIFF
--- a/ui/src/components/ExternalTerm.vue
+++ b/ui/src/components/ExternalTerm.vue
@@ -49,7 +49,7 @@ export default class extends Vue {
 
   getLabels (pointer: GraphPointer | null): Literal[] {
     if (pointer) {
-      return pointer.out(schema.name, { language: [this.selectedLanguage, '*'] }).terms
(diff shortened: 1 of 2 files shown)


ui/src/components/ExternalTerm.vue
```diff
@@ -49,7 +49,7 @@ export default class extends Vue {
 
   getLabels (pointer: GraphPointer | null): Literal[] {
     if (pointer) {
-      return pointer.out(schema.name, { language: [this.selectedLanguage, '*'] }).terms
+      return pointer.out(schema.name).terms.filter(isLiteral) as Literal[]
```
martinmaillard, line 52: I had to add `as Literal[]` because for some reason, Typescript was ignoring my filter with guard.

tpluscode: Apparently typescript cannot figure out the common type for `.terms` return type

> `Term[] | [BlankNode] | ...etc`

Issue #740, opened 2021-05-25
Cube Designer - Data kind do not show up


FabianCretton:
I did set a data kind for year, the icon is displayed in the column header, but when editing the metadata the data kind and time precision are not displayed.

I do not really understand those values, just reporting this "edition" behavior

![image](https://user-images.githubusercontent.com/6167669/119525844-24278c80-bd7f-11eb-8a64-8a72ade502fb.png)

tpluscode:
Thank you @FabianCretton 

Small issue, we can already fix in next release

PR #741, opened 2021-05-25
fix(ui): load CBD of dimension meta into form when editing

tpluscode:
re #740 

The form needs to be populated with deep graph including blank nodes


martinmaillard (approved):
CBD? 🌿

l00mi:
Oh how I had the same thought.

Issue #742, opened 2021-05-26
SharedDimensions Linking: Filter in list of resources to link to.


martinmaillard:

martinmaillard:
I think that's done? I don't remember the difference with #743. Any idea @l00mi ?

tpluscode:
This one is to filter the entire table of KV pairs

Last time @l00mi and I discussed this, we concluded to keep as-is for now. If the list of values to map grows too much, we might first look into automations using `schema:identifier`

martinmaillard:
👍 Should we close this issue then?

tpluscode:
I would keep open. It is still something valid to implement potentially

PR #744, opened 2021-05-26
feat: filtering shared dimension terms

tpluscode:
First step towards #742 

This PR adds a search parameter to the terms URL template so that we can filter them by their names

### DIFF
--- a/apis/shared-dimensions/lib/domain/shared-dimensions.ts
+++ b/apis/shared-dimensions/lib/domain/shared-dimensions.ts
@@ -1,8 +1,9 @@
-import { CONSTRUCT } from '@tpluscode/sparql-builder'
+import { CONSTRUCT, SELECT } from '@tpluscode/sparql-builder'
 import { schema } from '@tpluscode/rdf-ns-builders'
 import { md, meta } from '@cube-creator/core/namespace'
 import { Term } from 'rdf-js'
 import env from '../env'
+import $rdf from 'rdf-ext'
 
 export function getSharedDimensions() {
   return CONSTRUCT`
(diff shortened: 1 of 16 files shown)


apis/shared-dimensions/lib/domain/shared-dimensions.ts
```diff
@@ -17,12 +18,31 @@ export function getSharedDimensions() {
...
+    .WHERE`
+      ${sharedDimension} a ${meta.SharedDimension} .
+      ${term} ${schema.inDefinedTermSet} ${sharedDimension} .
+      ${term} ${schema.name} ${name} .
+    `
+
+  if (freetextQuery) {
+    select = select.WHERE`FILTER (
+      REGEX(${name}, "^${freetextQuery}", "i")
+    )`
```
tpluscode, line 35: I quickly realised that this may give unexpected results with multilingual labels.

Should we take `Accept-Language` request header and filter by the requested languages, possibly with a fallback on a preconfigured lang (`de`/`en`)?

apis/shared-dimensions/lib/domain/shared-dimensions.ts
```diff
@@ -17,12 +18,31 @@ export function getSharedDimensions() {
     `
 }
 
-export function getSharedTerms(sharedDimension: Term) {
+export function getSharedTerms(sharedDimension: Term, freetextQuery: string | undefined, limit = 100) {
```
tpluscode, line 21: I did not want to completely hardcode the limit be we need it somewhere...

apis/shared-dimensions/lib/domain/shared-dimensions.ts
martinmaillard: We don't currently have a multi-lingual app, would you use the language selected for the cube preview?

martinmaillard (approved):
Looks good, apart from the language issue

apis/shared-dimensions/lib/domain/shared-dimensions.ts
l00mi: Can we simply stick to the label currently shown in the list? How is it done right now?

tpluscode: > How is it done right now?

The form indeed defaults to labels matching browser language (`window.navigator.languages`)

Filtering is not the same. If you filter on wrong language, or all combined, you might get an unexpected result back

PR #748, opened 2021-05-27
fix: set datatype of default csvw values

tpluscode:
Not convinced an integration test is necessary


martinmaillard (approved):
The commit message and PR title are a bit confusing, but I guess it's about CSVW default value and datatype?

PR #752, opened 2021-05-31
fix(shared dim): improve store queries

tpluscode:


apis/shared-dimensions/lib/store.ts
```diff
@@ -16,16 +16,21 @@ export interface SharedDimensionsStore {
   exists(id: NamedNode, type: NamedNode): Promise<boolean>
 }
 
-function resourceQueryPatterns(id: NamedNode) {
+function resourceQueryPatterns(id: NamedNode, strict: boolean) {
+  let rootPropPatterns = sparql`${id} ?rootProp ?rootObject .`
+  if (!strict) {
+    rootPropPatterns = sparql`OPTIONAL { ${rootPropPatterns} }`
+  }
```
tpluscode, line 23: Set `strict` only when reading.

When deleting, add `OPTIONAL` so that the query is less likely to leave any leftovers

martinmaillard (approved):

PR #774, opened 2021-06-17
feat(ui): search with freetext input

tpluscode:
This extends the implementation of `dash.AutoCompleteEditor` to use server-side search when the search URI Template has mapping for a `hydra:freetextSearch` property


ui/src/forms/editors/index.ts
```diff
@@ -73,16 +74,94 @@ export const enumSelect: Lazy<EnumSelectEditor> = {
...
+  hasFreeTextQueryVariable (params) {
+    return !!this.searchTemplate?.(params)?.mapping
+      .some(({ property }) => property?.equals(hydra.freetextQuery))
+  },
+  init (params) {
+    if (!this.hasFreeTextQueryVariable(params)) {
+      return instancesSelectCore.init?.call(this, params) || true
+    }
+
+    const { form, property, value, updateComponentState } = params
```
tpluscode, line 95: The "overloaded" `init` function is used when collection is filtered on input so that the initial value gets dereferenced and its label used as the placeholder.

ui/src/forms/editors/index.ts
```diff
@@ -73,16 +74,94 @@ export const enumSelect: Lazy<EnumSelectEditor> = {
...
+        }
+
+        loading(true)
+        const pointers = await this.loadChoices(params, filter)
+        const instances = pointers.map<Item>(p => [p, this.label(p, params.form)])
+        params.updateComponentState({
+          instances,
+        })
+        loading(false)
+      }
```
tpluscode, line 153: Instances are loaded when the component signals user's search interaction

martinmaillard:
UX-wise, I find it annoying that no options are shown until the user types something.

tpluscode:
Gotcha. It appears to be [by-design of vue-select](https://vue-select.org/guide/ajax.html#example). I think it does not load. Maybe it can be forced on first focus?

tpluscode:
Ok, I think it's working nicely now. I added `sh:minLength 0` so that empty search string also triggers a request and the component stats searching when the dropdown gets focus.

martinmaillard (approved):

PR #777, opened 2021-06-24
fix(publish): emit dimension versions at end of stream

tpluscode:
I don't understand why, but somehow the timing as off here, and the code which would emit the `schema:version` came too early, when the dimensions to version were not yet determined.

By moving it to the flush callback, we can wait for the entire stream to pass and only then push these quads

### DIFF
--- a/.changeset/quick-melons-lick.md
+++ b/.changeset/quick-melons-lick.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/cli": patch
+---
+
+In some cubes the `schema:version` was not applied to dimensions (re. #776)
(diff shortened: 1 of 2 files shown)


martinmaillard (approved):
Not sure I understand this one, but if it works 🤷

tpluscode:
See the `const versionedDimensions = this.variables.get('versionedDimensions')`

It's a set populated in one step and used in a subsequent one. I assumed that individual quad is processed synchronously via the callbacks but somehow the set did not have the expected content at the moment it processed its chunk.

By waiting until the flush callback we make sure that all versioned properties have been identified from the data and only then push the version triples

PR #778, opened 2021-06-25
fix: missing import step to map shared dimensions

tpluscode:


martinmaillard (approved):
Does that mean the data must be re-imported from source for the mappings to be applied?
We should probably adjust the help text that says: "If checked, the Cube will be immediately updated with new mappings. Otherwise, running the transformation will be necessary"

tpluscode:
`s/transformation/pipeline`?

martinmaillard:
I would say "transformation or import"

PR #779, opened 2021-06-30
fix: improve queries for /dataset resource

tpluscode:
I completely rewrote the query which loads the cube/shape/properties for the `/dataset` resource

Now it is done in 2 separate queries which are much faster combined than the previous, especially for large cubes.

### DIFF
--- a/apis/core/lib/domain/queries/cube.ts
+++ b/apis/core/lib/domain/queries/cube.ts
@@ -1,39 +1,55 @@
-import { CONSTRUCT } from '@tpluscode/sparql-builder'
+import { CONSTRUCT, DESCRIBE, SELECT } from '@tpluscode/sparql-builder'
 import StreamClient from 'sparql-http-client/StreamClient'
-import { cc, cube } from '@cube-creator/core/namespace'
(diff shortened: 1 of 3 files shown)


apis/core/lib/domain/queries/cube.ts
```diff
@@ -1,39 +1,55 @@
...
+function constructPropertyShapes(shape: NamedNode, cubeData: NamedNode, streamClient: StreamClient) {
+  return CONSTRUCT`?s ?p ?o. ${shape} ?sp ?so .`
+    .FROM(cubeData)
+    .WHERE`
+      ${shape} ?sp ?so .
+      ${shape} ${sh.property} ?property .
+
+      ?property ${sh.path} ?path .
+      ?property (<>|!<>)* ?s .
+      ?s ?p ?o
```
tpluscode, line 40: This is the core change. Rather than excluding triples from the result, the property shapes and their subgraphs are traversed

This is the crucial part is the restriction on `sh:path` property. Without it the query is two orders of magnitude slower

martinmaillard (approved):

PR #785, opened 2021-07-05
Unhandled rejections should fail job

tpluscode:
In the end, handling the `unhandledRejection` event serves the purpose of actually ending the process gracefully and marking the job resource as failed

Thus, the `--unhandled-rejections=strict` flag and newer version of `sparql-http-client` are unnecessary but won't hurt either. I can revert...


bergos (approved):
looks good to me

martinmaillard (approved):

PR #786, opened 2021-07-06
build: update lando

tpluscode:
At long last, an update to lando finally allows the proxied hosts to be used for all inter-service communication nada without using ports

This potentially breaking change came in a patch release, because why not.


martinmaillard (approved):
Tested, seems to work :)

PR #792, opened 2021-07-16
fix(ui): showing correct term URI

tpluscode:
I am not stoked about this.

Because `camouflage-rewrite` replaces all occurrences of the `ld.admin` URIs from the responses, I had to add a maimed URI as the "canonical" and fix it back on the UI so that it does not get replaced with local API identifiers like the rest of it...

### DIFF
--- a/.changeset/hip-jobs-switch.md
+++ b/.changeset/hip-jobs-switch.md
@@ -1,0 +1,6 @@
+---
+"@cube-creator/shared-dimensions-api": patch
+"@cube-creator/ui": patch
+---
+
+Displaying the canonical term URI on Shared Dimension screen (closes #787)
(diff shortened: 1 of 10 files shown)


martinmaillard:
I have no idea what this is about 🙈

l00mi:
@tpluscode can you point me the camouflage setting of this? It might be that this is a wrong config right now, and we better fix it there.

tpluscode:
https://github.com/zazuko/cube-creator/blob/master/apis/shared-dimensions/index.ts#L19-L23

What do you mean by "wrong config"? It's all or nothing string replacement.

tpluscode:
An alternative could be replace the camouflage middleware with a custom stream transform to rewrite everything except `oa:canonical` quad

tpluscode:
Ok, done that. I replace camouflage's content rewrite with custom middleware so that the tricks are no longer necessary

PR #794, opened 2021-07-20
Big files

### LINKED ISSUE #663: Big Files
Files surpassing the 100mb.

martinmaillard:
Will close #663 

I used the [Uppy](https://uppy.io/) library to setup chunked file upload directly to S3. Uppy has a plugin system that would allow us to support other storage backends.

The implementation is made of a UI component and a server middleware that adds all the routes necessary to generate pre-signed URIs for each chunk. Once a file is completely uploaded to S3, the UI uses the POST operation to create a CSVSource, providing the filename/key/location of the uploaded file to the API.

The integration with the existing API is not great. There may be some things to improve there.

TODO:
- [x] Check if file already exists in project
- [ ] The "replace CSV" feature won't currently work with very large files
- [x] Tests need to be fixed


martinmaillard:
The "replace file" feature still uses the old upload way because it is quite complicated (temp files moved around in a non-efficient way). But I would like to merge and deploy this version anyway to make sure it works well and fits the use case.

l00mi:
@martinmaillard that is ok for me as long the replace CSV works still for the "small files"?

martinmaillard:
Yeah it still works as it does now

l00mi (approved):
blind approval

ui/src/components/FileUpload.vue
```diff
@@ -0,0 +1,94 @@
...
+        maxNumberOfFiles: this.allowMultiple ? null : 1,
+      },
+      meta: this.fileMeta,
+    })
+
+    uppy.use(AwsS3Multipart, { companionUrl: uploadURL })
+    uppy.addPreProcessor(async () => {
+      // Hack to set fresh auth token before each upload
+      const headers = prepareHeaders(uploadURL, this.$store)
+      const headersObj = Object.fromEntries(headers.entries());
```
tpluscode, line 57: You can simplify a little by returning `Record<string, string>` from `prepareHeaders`

ui/src/components/FileUpload.vue
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+@Component({
+  components: { LoadingBlock, UppyDashboard },
+})
+export default class extends Vue {
+  @Prop() afterUpload: any
+  @Prop() fileMeta: any
+  @Prop({ default: true }) allowMultiple!: boolean
```
tpluscode, line 34: Do we ever change the default value of this prop?

e2e-tests/cube-projects/delete.hydra
```diff
@@ -38,7 +38,7 @@ With Class api:EntryPoint {
 With Class cc:CubeProject {
     Expect Link cc:csvMapping {
         Expect Property cc:csvSourceCollection {
-            Expect Operation cc:UploadCSVAction {
+            Expect Operation schema:CreateAction {
```
tpluscode, line 41: I notice that the body of these requests are not what we now submit but the tests pass.

We should add a `hydra:expects` to that operation and a simple shape to describe the file description.

tpluscode (changes requested):

martinmaillard:
Thanks for the review, I'll address all of these

ui/src/components/FileUpload.vue
martinmaillard: Not yet, we're going to need it if we implement the "replace file" feature.

e2e-tests/cube-projects/delete.hydra
martinmaillard: I added a shape to validate the input instead.

apis/core/bootstrap/shapes/csv-source.ts
```diff
@@ -3,12 +3,65 @@ import { csvw, dash, hydra, rdfs, schema, sh, xsd } from '@tpluscode/rdf-ns-buil
...
+    ${sh.property} [
+      ${sh.name} "Quote char" ;
+      ${sh.path} ${csvw.quoteChar} ;
+      ${sh.datatype} ${xsd.string} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${sh.minLength} 1 ;
+      ${sh.order} 20 ;
+    ] ;
+  .
```
tpluscode, line 58: Copy-pasted? I think this should be removed

martinmaillard: yup

tpluscode (approved):

PR #798, opened 2021-08-02
Reduce the pushily pipeline data flow and update barnard59

tpluscode:


l00mi (commented):
That is a huge PR. Any plan how to test basic functionality once deployed?

tpluscode:
It's not that big TBH. Half the changes is the lockfile and otel service configuration.

Changes are pretty much only in the cli project which is well tested in CI and in TEST it would be only the matter of ensuring  the pipelines run. Might switch GitLab to run from this branch before we even merge

Finally, it would be nice to actually set up open telemetry in our infrastructure so that the logging does not go to /dev/null @ludovicm67

martinmaillard (approved):

PR #800, opened 2021-08-03
feat: handle large files when replacing CSV file

martinmaillard:
Implements the same mechanism as #794 for the "replace CSV file" feature. I didn't add a changeset because the main feature hasn't been released yet.

I think it won't cause a memory issue, but we'll see when we deploy...


tpluscode (approved):
LGTM

PR #801, opened 2021-08-04
Add Open Telemetry to CLI

tpluscode:
I also added some spans to public steps


cli/index.ts
```diff
@@ -61,7 +66,9 @@ async function main() {
     .option('--auth-param <name=value>', 'Additional variables to pass to the token endpoint', parseVariables, new Map())
     .action(capture('Import', ({ job }) => ({ job }), importCube))
 
-  return program.parseAsync(process.argv)
+  return trace.getTracer('cube-creator-cli').startActiveSpan('run', span => {
```
sandhose, line 69: I think the better semantic would be one tracer per package, and name it with the name of the package, e.g. `@cube-creator/cli`

cli/lib/otel.ts
```diff
@@ -0,0 +1,56 @@
+import { trace } from '@opentelemetry/api'
+import { CollectorTraceExporter } from '@opentelemetry/exporter-collector'
+import { HttpInstrumentation } from '@opentelemetry/instrumentation-http'
+import { WinstonInstrumentation } from '@opentelemetry/instrumentation-winston'
+import { Resource, envDetector, processDetector } from '@opentelemetry/resources'
+import { NodeSDK } from '@opentelemetry/sdk-node'
+import { ResourceAttributes } from '@opentelemetry/semantic-conventions'
+import { BatchSpanProcessor } from '@opentelemetry/tracing'
+
+export const publishTracer = trace.getTracer('cube-publish')
```
sandhose, line 10: I'm not 100% sure if creating the tracer before the SDK was initialized works as expected.
What I've been doing in barnard59 is have a separate file with just
```ts
import { trace } from '@opentelemetry/api'

export const tracer = trace.getTracer('@cube-creator/cli')
```

cli/lib/otel.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+  await sdk.shutdown()
+  process.exit(1)
+}
+
+export async function opentelemetry() {
+  try {
+    await sdk.detectResources({ detectors: [envDetector, processDetector] })
+
+    const exporter = new CollectorTraceExporter()
```
sandhose, line 42: Since we're not doing that conditionally, it can be simplified by just having this in the SDK creation:
```
const sdk = new NodeSDK({
  traceExporter: new CollectorTraceExporter(),
  // ...
})
```

The `BatchSpanProcessor` is automatically setup by the SDK if you do this.

cli/lib/metadata.ts
```diff
@@ -117,11 +138,15 @@ export async function injectMetadata(this: Context, jobUri: string) {
 
     callback()
   }, function (callback) {
-    for (const [propertyShape, dimension] of propertyShapes) {
-      if (versionedDimensions.has(dimension)) {
-        this.push($rdf.quad(propertyShape, schema.version, revision))
+    publishTracer.startActiveSpan('injectMetadata#finished', span => {
```
sandhose, line 141: it usually does not make sense to create span for synchronous code like this, since it will only last a very short amount of time.
Maybe you want to have a bigger span that wraps all this and log additional events with [`span.addEvent()`](https://open-telemetry.github.io/opentelemetry-js-api/interfaces/span.html#addevent)? Also note that you can dynamically add attributes to spans with [`span.setAttribute()`](https://open-telemetry.github.io/opentelemetry-js-api/interfaces/span.html#setattribute)

cli/index.ts
```diff
@@ -61,7 +66,9 @@ async function main() {
...
+    return program.parseAsync(process.argv).finally(span.end.bind(span))
```
sandhose, line 70: I'd suggest you add some error handling there to record the exception if there is one. You could also do that at other stages to make it clearer where the error comes from, but at least one at the root will help.

```suggestion
  return trace.getTracer('cube-creator-cli').startActiveSpan('run', async span => {
    try {
      return await program.parseAsync(process.argv)
    } catch (err) {
      span.recordException(err)
      span.setStatus({ code: SpanStatusCode.ERROR, message: err.message })
      throw err
    } finally {
      span.end()
    }
```

sandhose (changes requested):
Left a few comments, otherwise LGTM

cli/lib/bufferDebug.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  finished(pipeline.stream, () => {
+    done = true
+  })
+
+  const pid = process.pid.toString()
+  const boundMeters = new Map<string, BoundBaseObserver>()
+  function getMeter(step: string) {
+    const bound = boundMeters.get(step) || bufferObserver.bind({
+      jobUri,
```
sandhose, line 60: same here, `job.uri` would be better

sandhose (changes requested):

PR #807, opened 2021-08-16
Upgrade deps

martinmaillard:
Will squash when merging


tpluscode (commented):
I would also update `@hyfrofoil/shaperone*` packages to their latest patch.

Minor might not break but it changes `lit-html` to `lit@2`

tpluscode (approved):

PR #817, opened 2021-08-31
Exporting project

tpluscode:
* [x] Project link to download Trig with project data
* [x] Show export link on UI
* [x] Tests

Existing data will require migration to add the `cc:export` link:

```sparql
INSERT {
  GRAPH ?project {
    ?project cc:export ?export
  }
} WHERE {
  GRAPH ?project {
    ?project a cc:CubeProject .

    BIND ( IRI(CONCAT(STR(?project), '/export')) as ?export )
  }
}
```

### DIFF
--- a/.changeset/red-pots-fry.md
+++ b/.changeset/red-pots-fry.md
@@ -1,0 +1,8 @@
+---
+"@cube-creator/core": patch
+"@cube-creator/model": patch
+"@cube-creator/ui": minor
+"@cube-creator/core-api": minor
+---
+
+Exporting project
--- a/apis/core/hydra/projects.ttl
+++ b/apis/core/hydra/projects.ttl
@@ -17,7 +17,9 @@ cc:CubeProject
     [ hydra:property
         cc:dataset ],
     [ hydra:property
-        cc:csvMapping ] ;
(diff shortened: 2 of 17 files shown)


martinmaillard (approved):
Works perfectly 👍 

The only thing I find weird is to put a property like `export` in the database even though it has no meaning outside of the API... but I guess that's the eternal debate.

The behavior that surprised me is that the `export` button was already shown before I ran the migration query, but when clicking it it returned 404. The project had an `export` property, but it was pointing to a resource that did not exist?

tpluscode:
> The behavior that surprised me is that the export button was already shown before I ran the migration query, but when clicking it it returned 404. The project had an export property, but it was pointing to a resource that did not exist?

That is odd... I will check that

tpluscode:
Ay yes, so it is like that because the Project mixin automatically populates that property, and so it appears in the response even if it's not in the data

https://github.com/zazuko/cube-creator/blob/e3f4bf4e9064c0ce723cbb2fd66426bfa694ca94/packages/model/Project.ts#L89

Might keep like that since we don't really plan project which would not be exportable?

martinmaillard:
Yeah not critical at all

PR #824, opened 2021-09-07
Add endpoint to retrieve organization datasets

### LINKED ISSUE #823: Create a REST endpoint, per Application Profile, that delivers all Metadata in a RDF/XML format.
To feed opendata.swiss we need to export the metadata in a specific format. The ideal solution is a SPARQL query which response in RDF/XML which is DCAT and it works.

Potentially we need to tweak the RDF/XML to make it actually work.

martinmaillard:
Closes #823 

TODO:
- [ ] Query the org's `sparqlEndpoint` instead of cube-creator DB
        Or maybe Lindas directly so that we can query the relevant endpoint.
- [ ] Remove authentication for endpoint?
- [ ] Only provide "published" datasets

### DIFF
--- a/apis/core/lib/domain/organization/datasets.ts
+++ b/apis/core/lib/domain/organization/datasets.ts
@@ -1,0 +1,129 @@
+import { dcat, dcterms, rdf } from '@tpluscode/rdf-ns-builders'
+import clownface, { GraphPointer, MultiPointer } from 'clownface'
+import { BlankNode, Literal, NamedNode } from 'rdf-js'
+import $rdf from 'rdf-ext'
+import { create as createXml } from 'xmlbuilder2'
+import { fetchOrganizationDatasets } from './query'
+import { shrink } from '@zazuko/rdf-vocabularies'
+import TermSet from '@rdfjs/term-set'
+
+export async function getOrganizationDatasets(organization: NamedNode): Promise<string> {
+  const quads = await fetchOrganizationDatasets(organization)
+  const pointer = clownface({ dataset: $rdf.dataset(quads) })
+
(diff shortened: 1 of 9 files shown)


apis/core/lib/domain/organization/datasets.ts
```diff
@@ -0,0 +1,129 @@
...
+            'dcterms:spacial': serializeTerm(dataset.out(dcterms.spacial)),
+            'dcterms:coverage': serializeTerm(dataset.out(dcterms.coverage)),
+            'dcterms:temporal': serializeTerm(dataset.out(dcterms.temporal)),
+            'dcat:distribution': serializeTerm(dataset.out(dcterms.distribution)),
+            'dcterms:accrualPeriodicity': serializeTerm(dataset.out(dcterms.accrualPeriodicity)),
+          },
+        })),
+      },
+    },
+  }).doc()
```
martinmaillard, line 60: There's probably a better way to do this. But it should be enough to see if the harvester is happy.

apis/core/lib/domain/organization/query.ts
```diff
@@ -58,3 +58,16 @@ export function cubeNamespaceAllowed(cube: NamedNode, organization: Term, client
...
+  return CONSTRUCT`?s ?p ?o`
+    .WHERE`
+      GRAPH ?project {
+        ?project ${schema.maintainer} ${organization} .
+        ?project ${cc.dataset} ?dataset .
+      }
+      GRAPH ?dataset {
+        ?s ?p ?o .
+      }
+    `.execute(client.query)
```
martinmaillard, line 72: Just realized we will need to filter-in only datasets that are "published to opendata.swiss" and not expired.

martinmaillard:
Closing, as it's going to be a separate app

PR #826, opened 2021-09-09
Project import

tpluscode:
Importing projects:

* [x] Add third option on project creation form
* [x] Create a separate handler for `multipart` request with project backup file
* [ ] Create project and restore its data+related resources from uploaded dataset

---

Here's my proposal for building and handling of multipart requests. A multipart request requires a different body which combines the RDF part for the project for itself and separate part(s) for the file(s). To make that work nicely with shaperone, I created a `FileUpload` editor which will set a custom `FileLiteral` to the resource graph. That literal is extended with the actual `File` selected by the user and will be used to construct the right request payload.

From the UI's perspective this looks like this:

1. The operation has a annotation property `hex:multiPartPath cc:export` which instructs it which property (path) will lead to the `FileLiteral`
2. When file literals are found, their associated files will be set as parts of the request. 
3. The actual RDF coming from shaperone becomes a part of its own (it serialises nicely because a literal is still just a literal). Currently on the part's name is hardcoded as "representation" but it could be optional; the server might search for parts with RDF media type, ignoring the name...
4. Finally, the server needs to conditionally call a different processing path for `multipart/form-data` so that it runs SHACL on the `representation` part.
5. The other parts are found from the request by matching the literals with the part names

Here's such an example. `cc:export "ubd28.trig"` matches `filename="ubd28.trig"` in second part

```
------WebKitFormBoundaryBWo9kLiqzhUbkpmg
Content-Disposition: form-data; name="representation"; filename="blob"
Content-Type: application/ld+json

{
  "id": "",
  "https://cube-creator.zazuko.com/vocab#projectSourceKind": {
    "id": "https://cube-creator.lndo.site/shape/cube-project/create#ExportedProject"
  },
  "http://www.w3.org/2000/01/rdf-schema#label": "UBD 28 (imported)",
  "https://cube-creator.zazuko.com/vocab#export": "ubd28.trig",
  "http://schema.org/maintainer": {
    "id": "https://cube-creator.lndo.site/organization/bafu"
  },
  "@context": {
    "id": "@id",
    "type": "@type"
  }
}
------WebKitFormBoundaryBWo9kLiqzhUbkpmg
Content-Disposition: form-data; name="ubd28.trig"; filename="ubd28.trig"
Content-Type: application/trig

# Exported project data
------WebKitFormBoundaryBWo9kLiqzhUbkpmg--
```


ui/src/forms/FileLiteral.ts
```diff
@@ -0,0 +1,13 @@
+import { Literal } from 'rdf-js'
+import LiteralExt from 'rdf-ext/lib/Literal'
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface FileLiteral extends Literal {}
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
tpluscode, line 8: These shenanigans are necessary until DefinitelyTyped/DefinitelyTyped#55671 ice merged

ui/src/api/index.ts
```diff
@@ -79,12 +84,57 @@ export const api = {
     return null
   },
 
+  prepareOperationBody (data: RdfResource | File, operation: RuntimeOperation): { body: File | FormData | string; contentHeaders: HeadersInit } {
+    if (data instanceof File) {
+      return {
+        body: data,
+        contentHeaders: { 'content-type': data.type },
+      }
+    }
```
tpluscode, line 93: We don't do direct file uploads any more so maybe I'd remove that path?

ui/src/forms/FileLiteral.ts
martinmaillard: Seems merged

apis/core/lib/domain/cube-projects/import.ts
```diff
@@ -0,0 +1,121 @@
...
+  const importedProject = clownface({ dataset: importedDataset, graph: project.id }).node(project.id)
+  const cubeIdentifier = importedProject.out(dcterms.identifier).value
+  if (!cubeIdentifier) {
+    throw new BadRequest('Missing cube identifier name in imported data')
+  }
+  if (await exists(cubeIdentifier, maintainer)) {
+    throw new DomainError('Another project is already using same identifier')
+  }
+  const sourceKind = importedProject.out(cc.projectSourceKind).term
+  if (!sourceKind || sourceKind.termType !== 'NamedNode' || !sourceKinds.has(sourceKind)) {
```
martinmaillard, line 104: I think that won't work because `sourceKinds` will have a different prefix depending on where the tool is deployed.

tpluscode: Oh, you're right. It might be best to migrate the kinds to the shared `cc:` namespace ?

martinmaillard: Fine by me. Makes sense if they have a meaning outside of a single instance.

tpluscode: They do now 😃

martinmaillard:
I had to manually change `projectSourceKind`, but appart from that it's working well.

One thing that needs improvement is the fact that there is zero indication that a CSV file is missing in the UI after the import. But I think we can address that later (unless we already have a way to do that?).

tpluscode:
> unless we already have an easy way to do that?

How do we signal the user that a file is broken? A red warning is displayed on sources when parsing failed

martinmaillard:
Yup, do you want to use the same `error` property? That may be the easiest way

martinmaillard (approved):

Issue #821, opened 2021-09-06
Unlist a published Cube


martinmaillard:
https://gitlab.ldbar.ch/bafu/umweltdatenkiosk-planning/-/issues/285

- [ ] Use `schema:expires` instead of `schema:validThrough` (migration path?)
- [ ] Unlisting a cube sets `schema:expires` on all previous versions of the cube
- The cube library should ignore all expired cubes by default

To be able to unlist we further need to tweak how we `expire` current Cubes. The rule set should be:
* Publishing a "Published" cube does expire all former cubes.
* Publishing a "Draft" does `expire` all former "Draft" cubes and does not touch the "Published" cube.
* Unlisting will add `expire` to all cubes.

l00mi:
Also documented in https://github.com/zazuko/rdf-cube-schema-viz/blob/master/README.md#version-history-of-cubes and https://github.com/zazuko/rdf-cube-schema-viz/issues/1. 

@herrstucki our proposition is to add `schema:expires` additionally and stop doing it once you have added the change on your side. Can you provide us an issue to track this?

l00mi:
After reflecting how it works on your side, you do not take `schema:validThrough` in consideration at all correct?

PR #831, opened 2021-09-14
Switch schema:validThrough -> schema:expires for cubes

martinmaillard:
`schema:expires` is semantically more correct.

Part of #821

### DIFF
--- a/cli/lib/cube.ts
+++ b/cli/lib/cube.ts
@@ -34,14 +34,14 @@ export function expirePreviousVersions(this: Pick<Context, 'variables' | 'log'>)
   })
 
   return CONSTRUCT`
-    ?cube ${schema.validThrough} ${timestamp}
+    ?cube ${schema.expires} ${timestamp} .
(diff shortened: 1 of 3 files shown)


cli/lib/cube.ts
```diff
@@ -33,15 +33,20 @@ export function expirePreviousVersions(this: Pick<Context, 'variables' | 'log'>)
...
+    ?cube ${schema.validThrough} ${timestamp} .
   `.WHERE`
     ${baseCube} ${schema.hasPart} ?cube .
 
+    OPTIONAL { ?cube ${schema.expires} ?expires }
     OPTIONAL { ?cube ${schema.validThrough} ?validThrough }
 
     filter (
-      !bound(?validThrough)
+      !(bound(?validThrough) or bound(?expires))
```
martinmaillard, line 49: Not 100% that's correct

martinmaillard: Well... definitely not correct

PR #853, opened 2021-09-22
Provide dataset workExamples

martinmaillard:
- [x] Display link to visualize.admin.ch on publication job when cube is published to visualize.
- [x] Add `workExample` for visualize and sparql endpoint on the dataset when publishing

### DIFF
--- a/.local.env
+++ b/.local.env
@@ -10,6 +10,7 @@ STORE_UPDATE_ENDPOINT=http://db.cube-creator.lndo.site/cube-creator/update
 STORE_GRAPH_ENDPOINT=http://db.cube-creator.lndo.site/cube-creator/data
 PUBLIC_QUERY_ENDPOINT=https://test.lindas.admin.ch/query
 TRIFID_UI=https://lindas.admin.ch/sparql
+VISUALIZE_UI=https://visualize.admin.ch
 
 # AuthN config
 AUTH_AUDIENCE=cube-creator
--- a/apis/core/lib/domain/job/create.ts
+++ b/apis/core/lib/domain/job/create.ts
@@ -97,7 +97,7 @@ export function prepareWorkExamples(cubeId: NamedNode, job: Job.PublishJob, data
   if (env.has('TRIFID_UI')) {
     workExamples.push({
       url: $rdf.namedNode(lindasWebQueryLink(cubeId, job.revision, job.publishGraph)),
-      encodingFormat: $rdf.literal('Application/Sparql-query'),
+      encodingFormat: $rdf.literal('application/sparql-query'),
       name: [
         $rdf.literal('SPARQL Endpoint mit Vorauswahl des Graph', 'de'),
         $rdf.literal('SPARQL Endpoint with graph preselection', 'en'),
--- a/apis/core/lib/domain/job/create.ts
+++ b/apis/core/lib/domain/job/create.ts
@@ -1,13 +1,17 @@
+import $rdf from 'rdf-ext'
 import { cc } from '@cube-creator/core/namespace'
 import { GraphPointer } from 'clownface'
 import { NamedNode, Term } from 'rdf-js'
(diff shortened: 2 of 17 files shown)


.local.env
```diff
@@ -10,6 +10,7 @@ STORE_UPDATE_ENDPOINT=http://db.cube-creator.lndo.site/cube-creator/update
 STORE_GRAPH_ENDPOINT=http://db.cube-creator.lndo.site/cube-creator/data
 PUBLIC_QUERY_ENDPOINT=https://test.lindas.admin.ch/query
 TRIFID_UI=https://lindas.admin.ch/sparql
+VISUALIZE_UI=https://visualize.admin.ch
```
l00mi, line 13: added an entry to https://github.com/zazuko/cube-creator/issues/401

l00mi (approved):

apis/core/lib/domain/job/create.ts
```diff
@@ -74,9 +79,75 @@ export async function createPublishJob({
...
 
+  const cubeId = isCsvProject(project)
+    ? organization.createIdentifier({
+      cubeIdentifier: project.cubeIdentifier,
+    })
+    : project.sourceCube
+
+  const dataset = await store.getResource<Dataset>(project.dataset.id)
+  const workExamples = prepareWorkExamples(cubeId, job, dataset)
+  workExamples.forEach(workExample => job.addWorkExample(workExample))
```
martinmaillard, line 90: Now adding the links to lindas and visualize when creating the job instead of after the job is finished. It's slightly less good on the UI side, but it allows us to re-use these "work examples" in the publish pipeline.

cli/test/lib/commands/publish.test.ts
```diff
@@ -235,6 +235,40 @@ describe('@cube-creator/cli/lib/commands/publish', function () {
...
+              property.node($rdf.blankNode()).addOut(sh.node, nodeSparql => {
+                nodeSparql
+                  .addOut(sh.property, prop => {
+                    prop
+                      .addOut(sh.path, schema.encodingFormat)
+                      .addOut(sh.hasValue, $rdf.literal('Application/Sparql-query'))
+                  })
+              }),
+            ])
+        })
```
martinmaillard, line 268: Used clownface instead of the JSON DLS because I don't know how to create an RDF list with the DSL.

apis/core/lib/domain/job/create.ts
```diff
@@ -74,9 +79,75 @@ export async function createPublishJob({
...
   return jobPointer
 }
 
+export function prepareWorkExamples(cubeId: NamedNode, job: Job.PublishJob, dataset: Dataset): Job.WorkExampleInput[] {
+  const workExamples = []
+
+  if (env.has('TRIFID_UI')) {
+    workExamples.push({
+      url: $rdf.namedNode(lindasWebQueryLink(cubeId, job.revision, job.publishGraph)),
+      encodingFormat: $rdf.literal('Application/Sparql-query'),
```
martinmaillard, line 101: @l00mi I copied this "format" from opendata.swiss but I remember you telling me that it was BS. What should I use for the link to LINDAS query endpoint instead?

tpluscode: `application/sparql-query` is the request type for SPARQL queries.

should this be the response format instead? a query endpoint can do content negotiation so maybe it multiple should be listed? `application/sparql-results+json` and `application/sparql-results+xml`?

tpluscode (approved):
Hardcoding the visualise UI is a bit icky but otherwise looks good.

apis/core/lib/domain/job/create.ts
martinmaillard: I'm open to anything. The only thing to know is that I use this value in the CKAN endpoint to determine the "format" property (SERVICE, HTML, etc.). So I need different values for visualize and the query endpoint.

PR #859, opened 2021-09-27
Shared dimension import

tpluscode:


martinmaillard:
I tried to export a dimension from test and import it locally and got a 400 "Imported dimension is invalid". Did the format change or is it another error. It's not easy to debug...

tpluscode:
Hm, there are a few problem here.

First, the export from TEST is not going into the `sh:property`. Those are the validation errors (missing `sh:path`). The duplication is also problematic but not a factor right now

I changed to send the validation report when validation fails. It does not help much when importing from TEST. What do you get when you try the local export?

martinmaillard:
```
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
Default metadata:  Value does not have shape Blank node b501
```

martinmaillard (approved):

PR #867, opened 2021-09-30
refactor: rename org profile

tpluscode:


l00mi (commented):
Lets go the full on and actually also call it in the UI Publishing Profile ?

tpluscode:
> Lets go the full on and actually also call it in the UI Publishing Profile ?

You mean [this](https://github.com/zazuko/cube-creator/pull/867/files#diff-3a15aec8e472f6d63c2b693d9735078b13263b2e35aabc70d15c1d5fcf8243e4R28-R29)

cli/lib/metadata.ts
```diff
@@ -54,6 +54,8 @@ export async function injectMetadata(this: Context, jobUri: string) {
 
   const propertyShapes = new TermMap<QuadSubject, QuadObject>()
 
+  const creatorTerms = dataset.pointer.out(dcterms.creator).terms
```
martinmaillard, line 57: Do if the "opendata.swiss organization" is not set in the cube metadata, observations won't have an `observedBy`. Is that ok? I thought `observedBy` was mandatory?

From the discussion last week, I thought the idea was to store the value for `observedBy` on the publishing profile?

tpluscode: Ah, maybe. @l00mi ?

l00mi (approved):

cli/lib/metadata.ts
```diff
@@ -159,6 +158,7 @@ export async function injectMetadata(this: Context, jobUri: string) {
     }
 
     if (quad.predicate.equals(cube.observedBy)) {
+      const creatorTerms = maintainer.pointer.out(cube.observedBy).terms
       for (const creator of creatorTerms) {
         if (creator.termType === 'NamedNode') {
```
martinmaillard, line 163: Is this check needed?

tpluscode: Other than satisfying the type system, technically a term can be multiple things and we do want an URI

tpluscode:
@martinmaillard why do cypress tests fail?

martinmaillard:
Should work now

PR #873, opened 2021-10-04
fix: expire cubes correctly based on cube status

tpluscode:
As mentioned in #821, the `schema:expires` should be set depending on the status of published cube

- Publishing a "Published" cube does expire all former cubes.
- Publishing a "Draft" does expire all former "Draft" cubes and does not touch the "Published" cube.

Until now all cube would have been expired regardless of the status. This PR fixes that

### DIFF
--- a/cli/lib/cube.ts
+++ b/cli/lib/cube.ts
@@ -4,10 +4,12 @@ import { cc } from '@cube-creator/core/namespace'
 import { GraphPointer } from 'clownface'
 import type { Context } from 'barnard59-core/lib/Pipeline'
 import { obj } from 'through2'
-import { CONSTRUCT } from '@tpluscode/sparql-builder'
+import { CONSTRUCT, sparql } from '@tpluscode/sparql-builder'
+import { IN } from '@tpluscode/sparql-builder/expressions'
 import { schema, sh } from '@tpluscode/rdf-ns-builders'
 import { Dataset } from '@cube-creator/model'
 import StreamClient from 'sparql-http-client/StreamClient'
+import { Draft } from '@cube-creator/model/Cube'
 import { loadDataset } from './metadata'
 
 export const getObservationSetId = ({ dataset }: { dataset: DatasetCore }) => {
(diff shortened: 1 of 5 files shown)


fuseki/sample-ubd.trig
```diff
@@ -41,9 +41,13 @@ graph <cube-project/ubd> {
...
   <https://environment.ld.admin.ch/foen/ubd/28/>
-    schema:hasPart <https://environment.ld.admin.ch/foen/ubd/28/1> .
+    schema:hasPart <https://environment.ld.admin.ch/foen/ubd/28/1> ;
+    schema:hasPart <https://environment.ld.admin.ch/foen/ubd/28/2> .
 
-  <https://environment.ld.admin.ch/foen/ubd/28/1> a cube:Cube .
+  <https://environment.ld.admin.ch/foen/ubd/28/1> a cube:Cube ;
+    schema:creativeWorkStatus <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Draft> .
+  <https://environment.ld.admin.ch/foen/ubd/28/2> a cube:Cube ;
+    schema:creativeWorkStatus <https://ld.admin.ch/vocabulary/CreativeWorkStatus/Published> .
```
tpluscode, line 50: `/foen/ubd/28/2` is published and must not be expired when publishing a newer draft (see updated tests)

cli/lib/cube.ts
```diff
@@ -33,11 +35,25 @@ export function expirePreviousVersions(this: Pick<Context, 'variables' | 'log'>)
...
+    `
+  }
+
   return CONSTRUCT`
     ?cube ${schema.expires} ${timestamp} .
   `.WHERE`
     ${baseCube} ${schema.hasPart} ?cube .
 
+    ${onlyDraftCubes}
+
```
martinmaillard, line 56: We are now also using this function to unlist cubes and I suspect this change would break that.
Should we create a separate function to unlist cubes?

tpluscode: No need, it does not break. The filter is only added for publish job with status `Draft`

martinmaillard: Ok, I got confused by `const job = this.variables.get('publish-job')`. Maybe worth adding a comment.

martinmaillard (approved):

martinmaillard (approved):

PR #918, opened 2021-10-28
feat: track last transformed row

tpluscode:
As the transformation job runs, this inspects the `csvw:rownum` quads and keeps track of where in the CSV the pipeline is at any given moment

In case of an error, this information is added to the job and displayed in the UI

### DIFF
--- a/ui/src/views/Job.vue
+++ b/ui/src/views/Job.vue
@@ -8,6 +8,9 @@
       <p class="has-text-grey-dark">
         Triggered at {{ job.created | format-date }}
       </p>
+      <p v-if="job.error && job.error.disambiguatingDescription" class="has-background-danger-light">
+        {{ job.error.disambiguatingDescription }}
+      </p>
     </header>
     <div class="is-flex gap-1">
       <b-tag v-if="job.revision">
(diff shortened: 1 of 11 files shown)


martinmaillard (approved):

ui/src/views/Job.vue
```diff
@@ -42,8 +45,8 @@
         <b-icon icon="book" />
         <span>View full log</span>
       </a>
-      <pre v-show="error" class="has-background-danger-light">
-        {{ error }}
+      <pre v-if="error" v-show="error" class="has-background-danger-light">
```
martinmaillard, line 48: I don't think you need to add this `v-if`

ui/src/views/Job.vue
```diff
@@ -42,8 +45,8 @@
...
+        {{ error.description }}
```
martinmaillard, line 49: What about old jobs? Will they have the same structure?

ui/src/views/Job.vue
martinmaillard: Nevermind, I did it on master

PR #929, opened 2021-11-02
chore: avoid duplicate test runs

martinmaillard:
This should avoid running all the workflows twice in each pull request.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 
 jobs:
   unit-test:


tpluscode:
Those are not "duplicate". They run of different code (PR branch vs merge result).

It has already saved us a couple of times from doing a bad merge where there were breaking changes on master which needs to be applied to feature branch

martinmaillard:
So it merges master into the branch and runs the tests? I had no idea. Is that documented somewhere (in the GH docs)?

tpluscode:
https://github.community/t/how-to-prevent-running-actions-on-push-when-pull-request-is-also-specified/17651/5

martinmaillard:
Thank you, good to know :)

Do we need the runs on push then? In which cases is that necessary?

tpluscode:
Mhm, I see what you're getting at now. The push is what you have locally. It also runs even before actually creating a PR

PR #940, opened 2021-11-03
fix: cube:Undefined for broken references

tpluscode:
This will replace incomplete referenced URIs with `cube:Undefined` if any referenced column is empty

To do that, a regular expression will be added to the CSVW columns. For example, a template `/resource/{NAME}/{ID}` will add a regex pattern `^/resource/.+/.+$` (will actually include the full domain)

This way, transformed values for the column's property which do not satisfy that regex will be identified as incomplete, and replaced with `cube:Undefined`

### DIFF
--- a/packages/model/lib/uriTemplateParser.js
+++ b/packages/model/lib/uriTemplateParser.js
@@ -40,6 +40,12 @@ class ParsedTemplateWrapper {
 
     return baseUri + this.toString()
   }
+
+  toRegex(baseUri) {
+    const absolute = this.toAbsoluteUrl(baseUri)
+
+    return '^' + absolute.replace(/{[^/]+}/g, '.+') + '$'
+  }
 }
 
 export function parse(template) {
--- a/.changeset/seven-moose-develop.md
+++ b/.changeset/seven-moose-develop.md
@@ -1,0 +1,6 @@
+---
+"@cube-creator/core-api": patch
+"@cube-creator/cli": patch
+---
+
+Reference columns would produce incorrect URLs when referenced columns were empty (fixes #789)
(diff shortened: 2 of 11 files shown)


packages/model/lib/uriTemplateParser.js
```diff
@@ -40,6 +40,12 @@ class ParsedTemplateWrapper {
 
     return baseUri + this.toString()
   }
+
+  toRegex(baseUri) {
+    const absolute = this.toAbsoluteUrl(baseUri)
+
+    return '^' + absolute.replace(/{[^/]+}/g, '.+') + '$'
```
martinmaillard, line 47: Why is `[^/]+` replaced with `.+`?

tpluscode: `/{[^/]+}/` matches template variables (excludes slash, otherwise it's greedy interpreting multiple variables as one)

Variable expressions are thus replaced with greedy patterns to match segments substituted with CSV values

tpluscode: Ah, you made me realise that the pattern must also be `[^/]` otherwise it is also too greedy

martinmaillard: Oh I get it. And don't slashes cause issues if there are multiple variables replaced? I guess not since your tests are passing ;)

martinmaillard (approved):

packages/model/lib/uriTemplateParser.js
tpluscode: Well, no, tests weren't good enough apparently :)

tpluscode: The pattern was good in the end so I added only the missing tests to check

martinmaillard (approved):

PR #944, opened 2021-11-08
chore: reduce polling frequency

martinmaillard:
The DB performance is really bad and reducing the amount of requests may
help?


tpluscode:
This polling is now orders of magnitude faster. May not need this change any more

PR #951, opened 2021-11-08
fix: insert timeouts no Stardog

tpluscode:


apis/core/lib/ResourceStore.ts
```diff
@@ -147,7 +147,7 @@ export class SparqlStoreFacade implements TripleStoreFacade {
         BIND ( ${now} as ?now )
         BIND ( ${pointer.term} as ?object )
```
martinmaillard, line 148: Is the BIND still needed?

tpluscode: In the constructed triples.

I thought it was neat not to repeat the `<URL>` everywhere. Maybe I misunderstand BIND anyway

martinmaillard (approved):

Issue #949, opened 2021-11-08
Warning duplicate use of Key Dimensions (creating a clashing IRI)


l00mi:
Generated or Defined Observation IRIs with the key dimensions are not unique.

For key dimensions, if the same set for 2 observations are used, hence the unique key becomes not unique anymore, a warning is issued.

Most probably that is a step which should be done on the store per SPARQL after transformation.

l00mi:
Finally the error is shown on top of the cube-designer, with the possibility to "filter" for the duplicates.

l00mi:
Check if there are observations which have all dimensions the same value defined.

Issue #948, opened 2021-11-08
Warning should be issued if data is missing (cube:Undefined) at least for key dimensions


l00mi:
There should be a warning, or a list created, for any observation where we have no value. Especially for key dimensions which will create wrong cubes.

tpluscode:
I'd propose to handle this similarly to #949, calculating the problems when the transformation completes and then showing on the Cube Designer screen.

Now, if we'd want to actually display observations with missing key dimensions, maybe this could be an optional filter when retrieving the observations on said screen

PR #952, opened 2021-11-08
feat: calculate "no measure dimension" error on the API

tpluscode:
The moves the piece of code to determine that there is no measure dimension to the code which saves dimension metadata and updates the error accordingly

### DIFF
--- a/fuseki/sample-ubd.trig
+++ b/fuseki/sample-ubd.trig
@@ -75,17 +75,28 @@ graph <cube-project/ubd/dataset> {
 <cube-project/ubd/dimensions-metadata> void:inDataset <ubd-example> .
 graph <cube-project/ubd/dimensions-metadata> {
   <cube-project/ubd/dimensions-metadata> a cc:DimensionMetadataCollection , hydra:Resource ;
+    schema:hasPart <cube-project/ubd/dimensions-metadata/dimension-value> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/dimension-year> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/station> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/aggregation> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/pollutant> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/unit> ;
+    schema:error
+      [
+        schema:identifier "MissingMeasureDimension" ;
+        schema:description "No Measure dimension defined" ;
+      ] ;
   .
 
   <cube-project/ubd/dimensions-metadata/dimension-year>
     schema:about <https://environment.ld.admin.ch/foen/ubd/28/dimension/year> ;
     schema:name "Jahr"@de ;
     qudt:scaleType qudt:IntervalScale ;
+  .
+
+  <cube-project/ubd/dimensions-metadata/dimension-value>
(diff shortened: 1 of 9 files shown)


fuseki/sample-ubd.trig
```diff
@@ -75,11 +75,17 @@ graph <cube-project/ubd/dataset> {
...
     schema:hasPart <cube-project/ubd/dimensions-metadata/dimension-year> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/station> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/aggregation> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/pollutant> ;
     schema:hasPart <cube-project/ubd/dimensions-metadata/unit> ;
+    schema:error
+      [
+        schema:identifier "MissingMeasureDimension" ;
+        schema:description "No Measure dimension defined" ;
+      ] ;
```
tpluscode, line 88: The description to be displayed on the cube preview table
`identifier` is only used by the API so as not to create multiple same errors

ui/src/components/CubePreview.vue
```diff
@@ -181,12 +183,11 @@ export default class extends Vue {
...
-  get warningMessage (): string | null {
-    if (this.dimensions.length > 0 && !this.dimensions.some(({ isMeasureDimension }) => isMeasureDimension)) {
-      return 'No Measure dimension defined'
-    }
-
-    return null
+  get errors (): Schema.Thing[] {
+    return [
+      ...this.cubeMetadata.errors!,
+      ...this.dimensionMetadataCollection.errors!,
```
tpluscode, line 189: Combine all errors from cube dataset and metadata collection. This way we can place them freely depending on where they belong. The issues #949 and #948 will probably add to the former

martinmaillard:
One disadvantage of this is that the errors doesn't show up until the user edits the metadata of a dimension for the first time. Since it is supposed to be used as a guidance for the user, it's a bit problematic.

tpluscode:
Thanks @martinmaillard, very good remark

I think that on every transformation is not necessary in this case, since the metadata does not change when transforming.

I did however just commit so that the error is set initially when the cube metadata is initialized. Later I think it's only updating dimensions that should affect it?

packages/model/DimensionMetadata.ts
```diff
@@ -65,8 +67,18 @@ export function DimensionMetadataCollectionMixin<Base extends Constructor>(Resou
...
+} as const
+
+export const createNoMeasureDimensionError = Thing.fromPointer({
+  identifierLiteral: Error.MissingMeasureDimension,
+  description: 'No Measure dimension defined',
+})
+
 export const createCollection = initializer<DimensionMetadataCollection>(DimensionMetadataCollectionMixin, {
   types: [cc.DimensionMetadataCollection],
+  [schema.error.value]: [createNoMeasureDimensionError],
```
tpluscode, line 81: If you wonder about the lockfile updates, they are to bring in this feature where properties can be set using factory. `createNoMeasureDimensionError` is a `pointer => Resource` function so that resources can be initialized lazily, with the actual node being assigned only when the setter is used

martinmaillard:
Will it interact properly with the CSV mapping "key/measure dimension" inputs?

tpluscode:
What do you mean?

martinmaillard (approved):

PR #960, opened 2021-11-16
Undefined values warning

tpluscode:
This will create a warning about missing values for dimensions. There is only one warning with nested `schema:additionalProperty` for each affected dimension. Here's an example from a project rom INT (note that blank node will not be used in actual code)

```turtle
<https://int.cube-creator.lindas.admin.ch/cube-project/test-import-csv-from-url-w629nyz4dq1/dataset> 
    schema:error [
        schema:identifier "MissingObservationValues" ;
	schema:description "There are observation values missing for some dimensions" ;
	schema:additionalProperty [
		schema:name "Dimension" ;
		schema:identifier <https://culture.ld.admin.ch/sfa/test_import_csv_from_url/bemerkung> ;
		schema:value 4857 ;
	], [
		schema:name "Dimension" ;
		schema:identifier <https://culture.ld.admin.ch/sfa/test_import_csv_from_url/noga-code> ;
		schema:value 192 ;
	], [
		schema:name "Dimension" ;
		schema:identifier schema:longitude ;
		schema:value 192 ;
	], [
		schema:name "Dimension" ;
		schema:identifier <https://culture.ld.admin.ch/sfa/test_import_csv_from_url/bur> ;
		schema:value 1923 ;
	], [
		schema:name "Dimension" ;
		schema:identifier <https://culture.ld.admin.ch/sfa/test_import_csv_from_url/nace-code> ;
		schema:value 192 ;
	], [
		schema:name "Dimension" ;
		schema:identifier schema:latitude ;
		schema:value 192 ;
	] ;
] .
```

This will be recalculated after every time a transformation job completes successfully

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -44,6 +44,17 @@ services:
         interval: 5s
         timeout: 3s
         retries: 3
+  trifid:
+    type: compose
+    app_mount: false
+    ssl: true
+    services:
+      image: zazuko/trifid
+      command: tini -- /app/server.js --config=/config-sparql.json --sparql-endpoint-url=http://admin:password@db.cube-creator.lndo.site/cube-creator
+      ports:
+        - 8080:8080
+      environment:
+        DEBUG: "trifid*"
   s3:
     type: compose
     ssl: true
@@ -99,6 +110,8 @@ services:
 proxy:
   store:
     - db.cube-creator.lndo.site:3030
+  trifid:
+    - trifid.cube-creator.lndo.site:8080
   core:
     - cube-creator.lndo.site:45670
   ui:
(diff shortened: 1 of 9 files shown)


apis/core/lib/domain/job/transformJob.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+      }
+    }
+
+    {
+      ${countMissingObservationValues(job)}
+    }
+
+    BIND(IRI(CONCAT(STR(?dataset), "#missing-values-error")) as ?error)
+    BIND(IRI(CONCAT(STR(?dataset), "#missing-values-error/", md5(str(?dim)))) as ?errorDim)
```
tpluscode, line 65: Hash URIs under `/dataset` are bound to prevent using blank nodes

.lando.yml
```diff
@@ -44,6 +44,17 @@ services:
         interval: 5s
         timeout: 3s
         retries: 3
+  trifid:
```
martinmaillard, line 47: What do you use trifid for?

martinmaillard (approved):

.lando.yml
tpluscode: Just for YASGUI. It's quite annoying that Fuseki's UI only has a single query. Multiple tabs are useful

PR #961, opened 2021-11-16
fix: properly stream cube shape and sh:in only on demand

tpluscode:
This improves how we handle large `sh:in` lists in shapes:

1. Does not materialise Cube Shape in memory but stream the CONSTRUCT instead
2. Only include the `sh:in` RDF lists when requested with HTTP header `Prefer: include-in-lists`
3. Only set the preference from the publish pipeline


martinmaillard (approved):

martinmaillard:
Tested locally and it seems to do what's expected. Merging so that I can test with real data.

PR #970, opened 2021-11-18
Shared term dynamic props

tpluscode:
## Example

https://user-images.githubusercontent.com/588128/142428165-0e085f6c-bbf6-4b61-9abc-2ec56173792a.mp4

## Details

The dynamic properties are defined on the shared dimension, with the possibility to make it required and literal or as a reference to another shared dimension. Here's an example

```turtle
[]
  hydra:required false ;
  sh:datatype xsd:int ;
  rdf:predicate qb:order ;
  rdfs:label "Order" ;
  md:dynamicPropertyType "Literal" ;
.
```

They are then transformed and injected into the shared dimension term as SHACL Property

```turte
[]
  sh:path qb:order ;
  sh:name "Order" ;
  sh:minCount 0 ;
  sh:maxCount 1 ;
  sh:order 100 ;
  sh:datatype xsd:int ;
.
```

Because there is only one shape as a dereferancable resource, the dynamic properties are injected on demand when `Prefer` header is provided, referring to the shared dimension:

```
Prefer: target-class=http://shared/dimension/uri
```


l00mi:
Looks very promising! Lets wait until @martinmaillard has the time to have a more indepth look.

martinmaillard:
For some reason I don't get the dynamic properties in the form to create a term.

Does it have anything to do with?:
> Because there is only one shape as a dereferancable resource, the dynamic properties are injected on demand when Prefer header is provided, referring to the shared dimension:

Do I have to add this header in the UI?

tpluscode:
Should be good now @martinmaillard

martinmaillard (approved):

PR #973, opened 2021-11-22
Fix contactPoint

martinmaillard:
- Use `schema:contactPoint` instead of `lindas:contactPoint`
- Avoid storing a new `schema:contactPoint` every time the dataset is saved

### DIFF
--- a/apis/core/lib/domain/dataset/update.ts
+++ b/apis/core/lib/domain/dataset/update.ts
@@ -1,4 +1,4 @@
-import { cc, lindas } from '@cube-creator/core/namespace'
+import { cc } from '@cube-creator/core/namespace'
 import { dcat, hydra, rdf, schema, vcard, _void } from '@tpluscode/rdf-ns-builders'
 import { GraphPointer } from 'clownface'
 import { NamedNode } from 'rdf-js'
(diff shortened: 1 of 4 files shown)


l00mi (approved):
much cleaner now too

apis/core/lib/domain/dataset/update.ts
```diff
@@ -36,19 +36,20 @@ export async function update({
   datasetResource.deleteOut(schema.hasPart).addOut(schema.hasPart, hasPart.terms)
   datasetResource.deleteOut(cc.dimensionMetadata).addOut(cc.dimensionMetadata, dimensionMetadata.terms)
 
-  // Set LINDAS-specific contact point
+  // Set schema.org contact point
+  datasetResource.deleteOut(schema.contactPoint)
```
tpluscode, line 40: There will be leftovers. Let's also remove the second level

```suggestion
  datasetResource.out(schema.contactPoint).deleteOut()
  datasetResource.deleteOut(schema.contactPoint)
```

tpluscode (changes requested):

tpluscode (approved):

PR #977, opened 2021-11-22
build: run nodemon using npx should hopefully fix occasional CI failures

tpluscode:
Oftentimes we get CI failures because somehow lando's build step does not install `nodemon` or whatever happens.

By installing nodemon with our dependencies and starting from local we should mitigate the problem

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -66,9 +66,7 @@ services:
         - ./minio:/data/
   pipeline:
     type: node:custom
-    build:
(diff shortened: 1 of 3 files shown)


martinmaillard:
Do we need to add it to our dependencies if we use npx? I think it automatically installs it if missing

tpluscode:
This changes in newer versions of node but we should be good for now

martinmaillard (approved):

PR #990, opened 2021-11-29
fix: use xsd:integer for numbers in forms

tpluscode:


packages/core/datatypes.ts
```diff
@@ -5,7 +5,7 @@ export const datatypes: [NamedNode, string[]][] = [
   // Most used datatypes, in alphabetical order of the label
   [xsd.string, ['string']],
   [xsd.boolean, ['boolean']],
-  [xsd.int, ['int']],
+  [xsd.integer, ['int']],
```
martinmaillard, line 8: Maybe change the label as well

martinmaillard (approved):

PR #995, opened 2021-12-01
Remove `sh:targetClass` from nested shape

martinmaillard:
I noticed the issue while tested the new nested validation results. `sh:targetClass` would cause the nested shape to be applied twice: once because of `sh:node` and once because of `sh:targetClass`. The validation results would also appear twice: once at the root of the report because of `sh:targetClass`, and once nested because of `sh:node`.

### DIFF
--- a/apis/core/bootstrap/shapes/dataset.ts
+++ b/apis/core/bootstrap/shapes/dataset.ts
@@ -320,22 +320,22 @@ ${shapeId} {
   .
 
   ${vcardOrganization} a ${sh.NodeShape} ;
-  ${rdfs.label} "Organization" ;
-  ${sh.property} [
-    ${sh.name} "Name" ;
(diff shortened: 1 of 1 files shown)


tpluscode:
Wouldn't this be something to improve in the shacl library? To run each shape only once for any given focus node?

martinmaillard:
I don't know, the behavior makes sense to me. `sh:targetClass` means the shape should be applied to any node with this class. How would the library decide if the result should be nested or not if it runs it only once?

I don't know why we need `targetClass` here. Is there a reason?

tpluscode:
> How would the library decide if the result should be nested or not if it runs it only once?

Ah I see, this is tricky. Maybe there is no answer. Ok, let's refrain from mixing `sh:node` and `sh:target*` for now. Looks like it's superfluous here

tpluscode (approved):

PR #998, opened 2021-12-01
Nested validation results in forms

martinmaillard:


packages/testing/lib/chaiShapeMatcher.ts
```diff
@@ -34,7 +34,7 @@ function isDataset(shapeLike: any): shapeLike is DatasetCore {
 }
 
 function toJSON(result: Validate.ValidationResult) {
-  const { focusNode, resultMessage, resultPath, resultSeverity } = RdfResourceImpl.factory.createEntity<ValidationResult>(result.cf).toJSON()
+  const { focusNode, resultMessage, resultPath, resultSeverity } = RdfResourceImpl.factory.createEntity<ValidationResult>(result.pointer).toJSON()
```
martinmaillard, line 37: Will require upgrading @types/rdf-validate-shacl to 0.4 once released

martinmaillard:
I think tests will still fail because of https://github.com/zazuko/rdf-validate-shacl/pull/81

The validator now doesn't fail when the target node doesn't exist in the data dataset. The middleware returns 204 (empty) instead of 400. I'll have to fix that.

tpluscode:
I updated the types package

packages/shacl-middleware/index.ts
```diff
@@ -74,6 +76,15 @@ export const shaclMiddleware = ({ getTargetNode, loadResource, loadResourcesType
...
   }
 
+  if (resource.dataset.match(targetNode).size === 0 && shapes.match(null, sh.property).size > 0) {
+    const response = new ProblemDocument({
+      status: 400,
+      title: 'Request validation error',
+      detail: 'Missing target node in request body',
+    })
+    return res.status(400).send(response)
+  }
```
martinmaillard, line 86: Not too sure about this. I added this check to keep the same behavior.

tpluscode: Mhm, so I added this check previously to prevent false positives when no validation actually occurred.

Why did it stop working after your changes?

martinmaillard: Because of this change in the lib https://github.com/zazuko/rdf-validate-shacl/pull/81

tpluscode: Right. I wonder if SHACL should have some kind of "strict" mode so that both scenarios would natively supported

martinmaillard: 🤷

PR #1010, opened 2021-12-06
fix: "disappearing" CSVs

tpluscode:
Prior to this PR we would export the S3 location to sources as they appeared in the original project. If said project would be imported to same environment, these files would be deleted to "make room" for new project's CSVs

This removes the actual paths to uploaded CSVs so they will not be inadvertently deleted


l00mi (commented):
Good catch!

martinmaillard (approved):

PR #1018, opened 2021-12-08
feat: validate cube meta with SHACL

tpluscode:


cli/shapes.ttl
```diff
@@ -0,0 +1,21 @@
...
+      sh:path [ sh:inversePath schema:hasPart ] ;
+      sh:node
+        [
+          sh:property
+            [
+              sh:path rdf:type ;
+              sh:hasValue schema:CreativeWork ;
+            ]
+        ]
+    ] ;
```
tpluscode, line 20: To be replaced with the shape(s) developed for cube schema

cli/pipelines/publish.ttl
```diff
@@ -19,21 +26,76 @@
...
 .
 
+<#cubeMetadata>
+  a :Pipeline, :ReadableObjectMode ;
+  :steps
+    [
+      :stepList (
+                  <#loadCubeMetadata>
+                  <#shaclValidate>
+                  <#flatten>
```
tpluscode, line 42: The core change, separating the query for cube metadata from the observations themselves so that we can run SHACL validation on the metadata alone

martinmaillard:
I tried to run the job locally and got the following error. It's not happening on master:
Screenshot 2021-12-13 at 17 03 57

It looks like a timeout, but I don't know exactly where/when it happens.

tpluscode:
I had that initially too and the timeout was because the pipeline was not actually streaming. That was caused by misconfigured pipeline step but I thought I'd fixed that now. Works in the CI after all 😅

martinmaillard (approved):

PR #1019, opened 2021-12-08
feat: add fields to define relations between dimensions

martinmaillard:


ui/src/views/DimensionEdit.vue
```diff
@@ -45,7 +48,33 @@ export default class extends Vue {
...
+      this.dimensions.forEach((d, index) => {
+        position
+          .addOut(rdf.first, d.about, dimPtr => {
+            dimPtr.addOut(schema.name, d.name)
+          })
+          .addOut(rdf.rest, position.blankNode(`_dimensions-${index + 1}`))
+
+        position = position.out(rdf.rest)
+      })
+      position.in(rdf.rest).deleteOut(rdf.rest).addOut(rdf.rest, rdf.nil)
```
martinmaillard, line 68: I had to do this horror to avoid a weird blank node clash when using `addList`. We may want to fix that, but I wanted to get it working first.

tpluscode (approved):

PR #1023, opened 2021-12-09
feat: large dimension warnings

tpluscode:


martinmaillard:
It's working well, but I find the message a bit confusing.

- "truncated dimension" makes it sound like the data was truncated, which was hopefully not the case
- Can we make it a bit more actionnable? What can users do to fix this warning? Select a non-string datatype probably?

tpluscode:
Yes, I agree the phrase is not great. Maybe this, also linking to the schema doc now that we have it?

> Dimension <${this.predicate.value}> contains ${this.in.size} unique values. At this size they will not be listed in cube's [code list](https://zazuko.github.io/rdf-cube-schema/#usage-of-code-lists) constraint. This may be a indication of a numeric or temporal dimension not mapped to its correct data type

martinmaillard:
This is much better 👍 
Not sure how to handle the link though ;) Do we want to put markdown in messages? Or did you have another idea?

tpluscode:
Yea, I though markdown like we already support on forms?

tpluscode:
Alternatively make this an object node and add a "see also" link

martinmaillard:
I'm working on improvement the job details page. I will include markdown rendering for errors 👍

martinmaillard (approved):

PR #1030, opened 2021-12-14
Better dataset keywords input

martinmaillard:
Instead of having to create one entry per keyword, we have one field per language, with multiple keywords in the same field. It is much faster to edit.

### DIFF
--- a/.changeset/short-lions-beam.md
+++ b/.changeset/short-lions-beam.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/ui": patch
+---
+
+Improve dataset keywords input field (closes #924)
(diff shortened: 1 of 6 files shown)


tpluscode (changes requested):
I was confused fo a moment because it looks like the other language fields we have

I would propose changes as follows:

1. Always show all languages
2. Make the language a label or readonly
2. No (+) button
3. Add a placeholder text to the inputs

martinmaillard:
I made a compromise where, when hitting the "+" button, the language is auto-populated with the next language that is not used yet. I think that will improve usability quite a bit.

tpluscode (approved):
OK let's run with what we have

PR #1031, opened 2021-12-14
fix: make all fields optional in CSV dialect form

martinmaillard:
The upload used to store "empty string" as delimiter and quote char when
they could not be determined from the file. Store `undefined` instead
and make these 2 fields optional.

### DIFF
--- a/.changeset/tender-crabs-call.md
+++ b/.changeset/tender-crabs-call.md
@@ -1,0 +1,6 @@
+---
+"@cube-creator/core-api": patch
+"@cube-creator/ui": patch
+---
+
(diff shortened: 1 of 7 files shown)


tpluscode (approved):
Looks good, I think. My inly concern is leaving the dialect up to parser's default but the worst thing is going to happen is failed pipeline, right?

How does this actually fix the linked issue?

martinmaillard:
It's a valid concern and we may still decide to enforce a default. But I think it's better than storing `empty string` when the sniffer returns `undefined`.

It fixes the issue by not making mandatory a field that was not populated in the first place. That was very confusing.

PR #1033, opened 2021-12-15
feat(ui): add support for `sh:group` to group properties in separate tabs

martinmaillard:


tpluscode (commented):
👌 
Works well but I would consider adding some default label at least to the default group

martinmaillard:
Probably a good idea... but since we control the whole chain, why would we create a group without a label?

Issue #700, opened 2021-05-05
Add a field to define the shacl:order of a Dimension.


l00mi:
According to https://github.com/zazuko/rdf-cube-schema/commit/bc8fe2a5fb4151a9278a5e82601d917e5f2a6a0e add a field on the Dimension which can specify the order of the dimensions in between each other.

1. Find mechanism to set order without allowing same value in different dimensions. Ideally create an interface which only allows to set the order, and does not ask for integers to be set.
2. Apply the order directly on the table in the cube-viewer, and further give a visual hint that the order is explicitly set.

Issue #987, opened 2021-11-29
Order Dimensions by Type in Cube-Designer


l00mi:
In the designer view order the dimensions by Type: 

* 1 explicitly set #700
* 2 KeyDimension, MeasureDimension, other.
* 3 Alphabetic

l00mi:
Depends on https://github.com/zazuko/cube-creator/issues/700

PR #1041, opened 2022-01-04
feat: add dimension order property

martinmaillard:
- Add a `sh:order` property on dimension metadata (manually input in the form for now)
- Sort dimensions by
  1. Order
  2. Measure dimension
  3. Key dimension
  4. Alphabetically (property)

Part of #700, but only with the simplest UI possible: a field to enter the order value.

### DIFF
--- a/ui/src/store/modules/project.ts
+++ b/ui/src/store/modules/project.ts
@@ -84,7 +84,16 @@ const getters: GetterTree<ProjectState, RootState> = {
   },
 
   dimensions (state) {
-    return state.dimensionMetadataCollection?.hasPart || []
+    const dimensions = state.dimensionMetadataCollection?.hasPart || []
+
+    // Sort dimensions with the following priority:
+    //   sh:order / measure dimension / key dimension / property (alphabetically)
+    return dimensions.sort((dim1, dim2) => (
+      ((dim1.order ?? Infinity) - (dim2.order ?? Infinity))) ||
+      ((dim1.isMeasureDimension ? 1 : Infinity) - (dim2.isMeasureDimension ? 1 : Infinity)) ||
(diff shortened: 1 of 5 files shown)


tpluscode (approved):
Good in general though I wonder if SHACL is the rightly used here.

Might consider [dtype:order](https://prefix.zazuko.com/dtype:order) as a general-purpose. We use that for ordering csvw columns

ui/src/store/modules/project.ts
```diff
@@ -84,7 +84,16 @@ const getters: GetterTree<ProjectState, RootState> = {
   },
 
   dimensions (state) {
-    return state.dimensionMetadataCollection?.hasPart || []
+    const dimensions = state.dimensionMetadataCollection?.hasPart || []
+
+    // Sort dimensions with the following priority:
+    //   sh:order / measure dimension / key dimension / property (alphabetically)
```
tpluscode, line 90: Ah, wasn't the proposal to put measure/key first?

martinmaillard:
Interesting point. Aren't dimensions described in a SHACL shape in the end? I took `sh:order` from https://zazuko.github.io/rdf-cube-schema/#sh-order

ui/src/store/modules/project.ts
martinmaillard: If we put measure/key first, users won't have full control over the ordering. But I may be wrong, I did what made sense to me. Did someone write the proposal somewhere?

martinmaillard: Found it: https://github.com/zazuko/cube-creator/issues/987
I think my solution follows the proposal.

tpluscode: 👌

tpluscode:
Your call, I don't mind either really. `sh:order` does not define any semantics so it's never going to be an issue. FWIW, I'm not sure I would use it to annotate cube dimensions either 😉

PR #1044, opened 2022-01-05
Dynamic properties improvements

tpluscode:


tpluscode:
Yes, thanks for checking. It was missing `schema:multipleValues` in ignore props

martinmaillard:
This is nice. The only thing that could really use some improvement is the languages selection:
Screenshot 2022-01-10 at 09 30 36

Ideally, this would be a list of checkboxes (or at least a multi-select).

tpluscode:
> Ideally, this would be a list of checkboxes (or at least a multi-select).

I like that idea. For now #1045 will change the above into a series of dropdowns but it should totally be a `dash:MultiEditor`.  Buefy already has what we need: https://buefy.org/documentation/checkbox#checkbox-button

tpluscode:
app cube-creator lndo site_shared-dimensions_dimension!!technologies_edit

tpluscode:
I improved the selector. Please have another look @martinmaillard

martinmaillard (approved):
Much better!

PR #1045, opened 2022-01-05
fix(ui): smarter radio selection

tpluscode:


ui/src/forms/matchers.ts
```diff
@@ -9,7 +9,8 @@ export const radioButtons: SingleEditor = {
     if (choices.length === 0) {
       return 0
     }
-    if (choices.length > 5) {
+    const valuesLength = choices.flatMap(term => [...term.value]).length
+    if (choices.length > 5 || valuesLength > 30) {
```
tpluscode, line 13: Obviously arbitrary, we could in the future make this more adaptive to actual width of the form itself

martinmaillard (approved):

PR #1046, opened 2022-01-11
feat(cli): write published cube to file

tpluscode:
This adds a `publish --to (filesystem|graph-store)` option to CLI so that we will be able to materialise the cube on GitLab and curl to Stardog instead of streaming directly from the pipeline
That file will be uploaded when the pipeline is done, in a new `after` hook

Importantly, it was also necessary to break up the query to retrieve the cube itself into smaller portions. This is implemented by running a sequence of smaller queries with a `LIMIT` and increasing `OFFSET`


martinmaillard (approved):

martinmaillard (approved):
🚀

PR #1053, opened 2022-01-17
ci: no need to build images on PR hook

tpluscode:
I figure that we should only run the docker packaging on pull. Running also on PR hooks is a little pointless because the `branch` tags will be the same


ludovicm67 (approved):
LGTM

PR #1056, opened 2022-01-19
feat: add Dataset EU Theme Category

martinmaillard:


apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -26,6 +26,21 @@ const themesQuery = sparql`construct {
...
+  ?c ${hydra.member} ?theme .
+  ?theme ${rdfs.label} ?name ; ?p ?o .
+} WHERE {
+  BIND ( iri('${env.API_CORE_BASE}eu-themes') as ?c )
+
+  graph <https://lindas.admin.ch/ontologies> {
+    ?theme a <http://publications.europa.eu/ontology/euvoc#DataTheme> ;
+      ${skos.prefLabel} ?name ;
+      ${skos.inScheme} <http://publications.europa.eu/resource/authority/data-theme> ;
+      ?p ?o .
```
tpluscode, line 40: Oof, this query returns 25k results.

I would at least add a filter on `?name` to only include "our" 5 languages

```sparql
filter (lang(?name) in ("en", "de", "rm", "fr", "it" ))
```

This reduces the dataset to 3800 triples, but which of `?p ?o` do we need? If we keep all that, let's also exclude the unwanted EU languages. 

```
    filter (!isLiteral(?o) || lang(?o) in ("en", "de", "rm", "fr", "it" ))
```

Now we're down to 2100 results but for 13 themes it still is still seems a bit excessive

tpluscode (changes requested):

apis/core/bootstrap/shapes/dataset.ts
martinmaillard: Thanks for checking this :)

I stupidly copy-pasted the query. But actually I don't think we need anything else than the label. I'll check and try to improve this.

tpluscode (approved):
Perfect 👌

PR #1061, opened 2022-01-25
feat(ui): show labels of IRI values in form

tpluscode:


ui/src/forms/editors/URIInput.vue
```diff
@@ -1,25 +1,65 @@
 <template>
-  <b-input :value="textValue" @input="onUpdate" />
+  <b-field :message="message">
+    <b-input :value="textValue" @blur="onUpdate" />
```
tpluscode, line 3: Changed to blur as an alternative for `debounce`

martinmaillard (approved):

PR #1067, opened 2022-02-01
fix: leftover shapes in shared dimension DB and very slow queries

tpluscode:
Every resource in shared dimensions API is stored alongside a shape which describe its properties. See [the example data](https://github.com/zazuko/cube-creator/blob/8b6a59ab0c1713ccb0338beb8e78881b9048e442/fuseki/shared-dimensions.trig#L77-L102)

Until now to retrieve a resource from that API we ran a single query which used property paths to query those shapes and build patterns to get the precise triples. This same implementation was used to delete and update resources, and export a dimension

This proved quite ineffective on Stardog. I assume that SPARQL property paths are to blame. 

This PR changes the implementation so that any needed shape(s) are retrieved first with a fast `CONSTRUCT` and the SHACL properties are recessed in memory to avoid deep SPARQL Paths in `OPTIONAL` clauses. This looks somewhat like

1 Get shape for a resource 

```SPARQL
PREFIX sh: <http://www.w3.org/ns/shacl#> 

CONSTRUCT { ?s ?p ?o }  
FROM <https://lindas.admin.ch/cube/dimension> WHERE {     
  {     
    SELECT ?rootShape  WHERE {     
      BIND( <https://ld.admin.ch/cube/dimension/technologies/shacl> as ?term )    
      ?rootShape sh:targetNode ?term .  
      MINUS {      
        # Exclude shapes which are children of property shapes    
        ?propertyShape sh:node ?rootShape .
      }     
    }  
    LIMIT 1  
  }    
  ?rootShape (!sh:targetNode)* ?s .   
  ?s ?p ?o .  
} 
```

2 Get its exact data

```SPARQL
CONSTRUCT { 
  <https://ld.admin.ch/cube/dimension/technologies/shacl> schema:name ?value1 .
  <https://ld.admin.ch/cube/dimension/technologies/shacl> rdf:type ?value2 .
  # etc
}
FROM <https://lindas.admin.ch/cube/dimension> 
WHERE { 
  <https://ld.admin.ch/cube/dimension/technologies/shacl> schema:name ?value1 .
  <https://ld.admin.ch/cube/dimension/technologies/shacl> rdf:type ?value2 .
  # etc
} 
```

### DIFF
--- a/.changeset/bright-bikes-talk.md
+++ b/.changeset/bright-bikes-talk.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/shared-dimensions-api": patch
+---
+
+The shared dimension API would query the store ineffectively, ultimately preventing the server from starting
--- a/apis/shared-dimensions/lib/domain/shared-dimension.ts
+++ b/apis/shared-dimensions/lib/domain/shared-dimension.ts
@@ -13,7 +13,7 @@ import { CONSTRUCT, sparql } from '@tpluscode/sparql-builder'
 import { streamClient } from '../sparql'
 import { StreamClient } from 'sparql-http-client/StreamClient'
 import TermMap from '@rdfjs/term-map'
-import { resourceQueryPatterns } from '../resource'
+import { getPatternsFromShape, resourceShapePatterns } from '../resource'
 
 export { importDimension } from './shared-dimension/import'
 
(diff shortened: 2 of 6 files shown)


l00mi (approved):
Lets try on test and see if works as expected.

martinmaillard (approved):

Issue #1052, opened 2022-01-17
Changing of Target Property follows in error of transformation.


l00mi:
https://int.cube-creator.lindas.admin.ch/app/cube-projects/cube-project!!fab-ofpp-poc-carte-couverture-communes-test-9q7azzbrlkk/csv-mapping/column-mappings/cube-project!!fab-ofpp-poc-carte-couverture-communes-test-9q7azzbrlkk!!csv-mapping!!table!!couverture-communes-onjdajnnjwr!!column-mapping!!codecommune-pybbo8rdp6b/edit

Changing the Target Property from `schema:identifier` to `commune` followed in error.

martinmaillard:
It seems like the dimension mapping's `schema:about` is not updated when changing the property.

PR #1068, opened 2022-02-01
fix: transformation failing after changing mapped property

martinmaillard:
#1052

### DIFF
--- a/.changeset/smooth-badgers-beg.md
+++ b/.changeset/smooth-badgers-beg.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/core-api": patch
+---
+
+Fix transformation failing after changing the property of a CSV column mapping (fixes #1052)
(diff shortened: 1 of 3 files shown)


tpluscode (commented):
Can we have a test for this is `apis/core/test/domain/column-mapping/update.test.ts`?

martinmaillard:
Yeah I will add the test.

But I was wondering: why do we need the dimension mapping to have a `schema:about` to the property of the dimension?
Wouldn't it be safer to access it through `( [ sh:inversePath cc.dimensionMapping ] schema:about )`?

tpluscode:
Maybe but it would probably be a larger refactoring. I expect issues where you don't have all data in memory. The changes would be to access properties but also potentially queries. The current way keeps a mapping dictionary resource self-describing

Issue #1051, opened 2022-01-17
Rework Mapping


l00mi:
- [x] Keep original strings in mapped dimensions (to be able to change later on) and distinguish them from mapped concepts. (By using a new temporary predicate.)
- [x] On publish only publish the temporary predicate attached values with the original predicate.
- [x] Allow multiple Shared Dimensions to be connect to one dimension.

martinmaillard:
We considered 3 solutions to store the original value:
1. RDF*:
    - Example: `<<observation1> <country> <switzerland>> cc:originalValue "CH"`
    - May cause support issue in the stores and the different libs
2. Reification: 
    - Example: 
       ```
        [ 
           rdf:subject <observation1> ;
           rdf:predicate <country> ; 
           <rdf:value> <switzerland> ; 
           cc:originalValue "CH" ; 
        ]
       ```
    - Will result in a lot more data
3. A new property: 
    - Example:
      ```
      <observation> <country#original> "CH" . 
      <country#original> a cc:OriginalValueProperty .
       ```
    - We'll need to filter out these values when publishing the cube

No decision yet, but leaning towards 3. Do you have an opinion @l00mi ?

l00mi:
Yes, I think solution 3 is the current best way to go.

PR #1094, opened 2022-02-15
Keep original value when mapping shared dimension

martinmaillard:
#1051

### DIFF
--- a/cli/lib/output-filter.ts
+++ b/cli/lib/output-filter.ts
@@ -58,6 +58,14 @@ export function removeHydraTriples({ subject, predicate, object }: Quad): boolea
   return !(isHydraTerm(subject) || isHydraTerm(predicate) || isHydraTerm(object))
 }
 
+function isOriginalValueTerm(term: Term): boolean {
+  return term.termType === 'NamedNode' && term.value.endsWith('#_original')
+}
+
+export function removeMappingsOriginalValues({ subject, predicate, object }: Quad): boolean {
+  return !(isOriginalValueTerm(subject) || isOriginalValueTerm(predicate) || isOriginalValueTerm(object))
+}
+
 export function ensureUniqueBnodes(this: Context, quad: Quad): Quad {
   const uuid = this.variables.get('bnodeUuid') ||
     this.variables.set('bnodeUuid', nanoid()).get('bnodeUuid')
--- a/cli/lib/output-filter.ts
+++ b/cli/lib/output-filter.ts
@@ -58,14 +58,6 @@ export function removeHydraTriples({ subject, predicate, object }: Quad): boolea
   return !(isHydraTerm(subject) || isHydraTerm(predicate) || isHydraTerm(object))
 }
 
-function isOriginalValueTerm(term: Term): boolean {
-  return term.termType === 'NamedNode' && term.value.endsWith('#_original')
-}
-
-export function removeMappingsOriginalValues({ subject, predicate, object }: Quad): boolean {
-  return !(isOriginalValueTerm(subject) || isOriginalValueTerm(predicate) || isOriginalValueTerm(object))
-}
-
 export function ensureUniqueBnodes(this: Context, quad: Quad): Quad {
   const uuid = this.variables.get('bnodeUuid') ||
     this.variables.set('bnodeUuid', nanoid()).get('bnodeUuid')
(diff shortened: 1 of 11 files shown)


cli/lib/output-filter.ts
```diff
@@ -58,6 +58,14 @@ export function removeHydraTriples({ subject, predicate, object }: Quad): boolea
   return !(isHydraTerm(subject) || isHydraTerm(predicate) || isHydraTerm(object))
 }
 
+function isOriginalValueTerm(term: Term): boolean {
+  return term.termType === 'NamedNode' && term.value.endsWith('#_original')
```
martinmaillard, line 62: I don't like this check, but I don't really know if it's possible to check which predicate has type `cc:OriginalValue` in a streaming way.

martinmaillard:
@tpluscode ping

cli/lib/output-filter.ts
```diff
@@ -58,6 +58,14 @@ export function removeHydraTriples({ subject, predicate, object }: Quad): boolea
...
+}
+
+export function removeMappingsOriginalValues({ subject, predicate, object }: Quad): boolean {
+  return !(isOriginalValueTerm(subject) || isOriginalValueTerm(predicate) || isOriginalValueTerm(object))
```
tpluscode, line 66: Why not just check predicates? When will there be a `#_original` as subject or object?

packages/core/namespace.ts
```diff
@@ -33,7 +33,8 @@ type CubeCreatorClass =
   'UnlistJob' |
   'ReplaceCSVAction' |
   'MediaLocal' |
-  'MediaURL'
+  'MediaURL' |
+  'OriginalValuePredicate'
```
tpluscode, line 37: Should be `OriginalValue` ?

cli/lib/output-filter.ts
tpluscode: Ah, this is to remove the values when publishing?

You could remove the `removeMappingsOriginalValues` step and filter out the originals when querying the observations by adding in `loadCube.ts`

```spraql
filter(not exists{    
  ?p a cc:OriginalValue    
})
```

cli/lib/output-mapper.ts
```diff
@@ -104,23 +120,51 @@ export async function mapDimensions(this: Pick<Context, 'variables'>) {
...
+          $rdf.quad(subject, predicateOriginal, object, graph),
+        ]
       }
     }
 
-    return quad
+    return [quad]
+  }
+
+  return flatMap(handleQuad)
```
tpluscode, line 153: I'm trying to understand why the `flatMap` is required. You don't really need the `predicateOriginal` quad push right away. If you also add it to the in-memory dataset the map step can be much like before.

tpluscode: By the way, doesn't the current implementation cause `#_original` to become a cube dimension in the cube. I don't see any changes in the `toCubeShape` etc

cli/lib/output-mapper.ts
```diff
@@ -104,23 +120,51 @@ export async function mapDimensions(this: Pick<Context, 'variables'>) {
...
+}
+
+export function injectNonObservationQuads(this: Pick<Context, 'variables'>) {
+  const nonObservationQuads = this.variables.get('nonObservationQuads')
```
tpluscode, line 157: I'm not exactly fond of the name. Maybe make it more specific to the purpose of keeping original values?

tpluscode (changes requested):

packages/core/namespace.ts
martinmaillard: It's a type that we assign on the predicate pointing to the original value, so that's why I named it that.

cli/lib/output-mapper.ts
martinmaillard: It didn't seem to cause any issue. But I'll apply your suggestion.

tpluscode: Implementation details aside, I think we need a test to verify that there are indeed no extra dimensions created in the cube constraint shape.

martinmaillard: By the way, storing all the original values in the `variable` will increase the memory footprint of the pipeline. Hopefully not too much...

tpluscode: My stream-fu is not strong enough but I wonder if we could also have it as a stream, but separate from the rest of the steps, and merged in the final output

tpluscode (approved):

PR #1095, opened 2022-02-16
Upgrade deps

martinmaillard:


martinmaillard:
Crap `Caused by: TypeError: csvw.pointer.dataset.match(...).map is not a function`. Another breaking change. I'll check that next week.

cli/lib/commands/runner.ts
```diff
@@ -66,7 +68,7 @@ export function create<TOptions extends RunOptions>({ pipelineSources, prepare,
       params: command.authParam,
     }
 
-    const apiClient = Alcaeus.create()
+    const apiClient = Alcaeus.create({ datasetFactory: $rdf.dataset }) as HydraClient<DatasetExt>
```
martinmaillard, line 71: Lying to the compiler for now because I didn't find a better way to do it.

cli/lib/commands/timeoutJobs.ts
```diff
@@ -46,7 +49,7 @@ export async function timeoutJobs({
     `
     .execute(client.query)
 
-  const apiClient = Alcaeus.create()
+  const apiClient = Alcaeus.create({ datasetFactory: $rdf.dataset }) as HydraClient<DatasetExt>
```
martinmaillard, line 52: Lying to the compiler for now because I didn't find a better way to do it.

apis/shared-dimensions/lib/handlers/shared-dimension.ts
```diff
@@ -40,7 +41,11 @@ export const put = protectedResource(shaclValidate, asyncMiddleware(async (req,
   return res.dataset(dimension.dataset)
 }))
 
-export const getExport = protectedResource(cors({ exposedHeaders: 'content-disposition' }), asyncMiddleware(async (req, res) => {
+export const getExport = protectedResource(cors({ exposedHeaders: 'content-disposition' }), asyncMiddleware(async (req, res, next) => {
+  if (!req.dataset) {
+    return next(new error.BadRequest())
+  }
```
tpluscode, line 47: Would be nice to actually create a suggestion as we discussed in rdfjs/express-handler

martinmaillard: What do you mean?

tpluscode (approved):
Looks good overall.

I would run `npx yarn-deduplicate` for good measure if you haven't

apis/shared-dimensions/lib/handlers/shared-dimension.ts
tpluscode: You proposed that it would be better if the `req.dataset` was always there but throw if there is nothing to parse (no body or not RDF)

martinmaillard:
I ran `npx yarn-deduplicate` between each commit ;)

apis/shared-dimensions/lib/handlers/shared-dimension.ts
martinmaillard: https://github.com/rdfjs-base/express-handler/issues/14

Issue #1097, opened 2022-02-21
PROD : Dimension Type cannot be deleted in Cube designer


ortnever:
in CC Prod, Cube UBD0066 Heavy Metal :  Metadata "key dimension" for dimension "Year" cannot be deleted. In the CSV Mapping no dimension type is selected for "Year" and it's correct. I transformed again but it didn't solve my problem.
Thank you to solve the bug as soon as possible.

tpluscode:
Thank you for reporting @ortnever. We found the issue and a fix is on its way.

That said, the "returning" dimension type is not actually persisted but only a default value, wrongly set in the metadata editor. If you remove it, the dimension will be saved exactly as expected, without the type. 

Sorry about the inconvenience.

ortnever:
I removed it with the minus and saved the metadata. 
The dimension is always with bold caracters in the dimension table, but you mean it's  saved correctly in the database? 

![image](https://user-images.githubusercontent.com/79834090/154961224-908d054a-7e36-4176-a388-de03c4796110.png)

![image](https://user-images.githubusercontent.com/79834090/154961737-2d19a508-3e33-484f-a9bb-61b0956b3800.png)

tpluscode:
That's correct. Also, I just merged the fix. Should be up shortly

ortnever:
Thanks Tomasz. It's ok now.

PR #1098, opened 2022-02-21
fix: missing join on project's mapping resource

tpluscode:
The missing pattern joining the table to its mapping resource would produce an unwanted cross-product of columns from multiple projects.

In the case of #1097, there exists a `jahr` column in another project which is a key dimension, thus it was wrongly determined as such when setting the default value in the form


l00mi (approved):
Please hotfix

PR #1103, opened 2022-02-23
Hierarchies

tpluscode:


ui/src/router.ts
```diff
@@ -214,6 +217,23 @@ const routes: Array<RouteConfig> = [
...
+        name: 'Hierarchies',
+        component: Hierarchies,
+        children: [
+          {
+            path: 'new',
+            name: 'HierarchyCreate',
+            component: HierarchyCreate,
+          },
+          {
+            path: '/hierarchy/:id',
```
martinmaillard, line 231: To follow the other URIs in the app, I would use `path: ':id'` so that the URL reflects the nesting of the UI.

ui/src/store/modules/sharedDimensions.ts
```diff
@@ -8,11 +8,13 @@ import { Collection, RdfResource } from 'alcaeus'
 export interface SharedDimensionsState {
   entrypoint: null | RdfResource
   collection: null | Collection,
+  hierarchies: null | Collection,
```
martinmaillard, line 11: We could create a separate store for hierarchies. But I guess that's also ok.

tpluscode: Thought about that but ultimately we're querying all from same Lindas so maybe this is more similar

tpluscode: Ah wait, this is not about the database :)

But conversely, maybe this reflects that dimensions and hierarchies indeed live closely as resources 🤔

martinmaillard: I'm fine either way. We may extract a separate store if hierarchies become a larger concern in the app.

martinmaillard (approved):

PR #1104, opened 2022-02-23
Fail publish job when store upload fails

martinmaillard:


cli/lib/publish/upload.ts
```diff
@@ -25,6 +25,7 @@ export function uploadCube(variables: VariableMap) {
         '-T',
         cubeFile,
         '-H', 'content-type:application/n-quads',
+        '--fail',
```
tpluscode, line 28: Nice, did not see that coming. Might also add that same flag to curl in `loadCube.ts`?

tpluscode (approved):

cli/lib/publish/upload.ts
martinmaillard: I think `loadCube` would already fail when loading the quads. But I can add it. It may make it easier to debug.

PR #1107, opened 2022-02-23
Upgrade ESLint to 8

martinmaillard:
I didn't upgrade @tpluscode/eslint-config to 0.3 because it seems to have broken module resolution.


tpluscode:
> I didn't upgrade @tpluscode/eslint-config to 0.3 because it seems to have broken module resolution.

It was missing the peer dependency `eslint-import-resolver-typescript`

tpluscode (approved):
Please dobule check @martinmaillard

PR #1113, opened 2022-02-24
Separate cli tests

tpluscode:


tpluscode:
Oh, the problem with publish pipeline might have been #1111. Still, I think it's good to have more fine-grained CI jobs

.github/workflows/ci.yaml
```diff
@@ -8,22 +8,10 @@ jobs:
     timeout-minutes: 45
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2.4.0
-        with:
-          node-version: 14
-          cache: 'yarn'
-      - run: yarn install --ci
+      - uses: ./.github/workflows/setup-env
```
martinmaillard, line 11: Nice, I never realised you could do that :)

tpluscode: Neither did I

.github/workflows/setup-env/action.yml
```diff
@@ -0,0 +1,21 @@
+name: "Env setup"
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@v2.4.0
+      with:
+        node-version: 14
```
ludovicm67, line 8: Is it possible to have it working with Node 16, since it's the latest LTS?

.github/workflows/ci.yaml
ludovicm67: It's a lot cleaner like this 👍

.github/workflows/setup-env/action.yml
tpluscode: I don't see why not :+1:

martinmaillard (approved):

PR #1120, opened 2022-03-01
Replace Buefy with Oruga

martinmaillard:
Replace our UI components library Buefy with [Oruga](https://oruga.io/), a Vue3-compatible alternative. I chose this one because:
1. The author of Buefy is one of the authors of Oruga (the structure of both libraries look very similar)
2. It seems to be the "officially supported" upgrade path to Vue 3
3. It has a Bulma theme

The documentation is slightly less polished and the lib is missing some components, but the transition was ok and I think it has what we need for the future.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -38,7 +38,6 @@
     "@uppy/vue": "^0.4.5",
     "@zazuko/rdf-vocabularies": "^2021.3.17",
     "alcaeus": "^2.1",
-    "buefy": "^0.9.5",
     "buffer": "^6.0.3",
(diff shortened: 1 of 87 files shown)


tpluscode (changes requested):
Cool, almost indistinguishable. I found two minor issues:

Radio button labels are off-centre:

image

File picker shows the native beside styled button:

image

martinmaillard:
Thanks! Will take a look

tpluscode (approved):
👌

Issue #1100, opened 2022-02-22
Switch to Vue3.0


l00mi:

martinmaillard:
Some notes about libraries before upgrading:

### Buefy will not support Vue 3

Proposed solution: switch to Oruga https://github.com/zazuko/cube-creator/pull/1120

### vue-class-component, vue-property-decorator, vuex-class don't support Vue 3 yet
 
If I remember correctly, we chose to use these libraries because it improved Typescript support with Vue 2. Vue 3 has much better Typescript support out of the box.

After 2 years working on this project, I can say that using a non-official way of declaring components was mostly frustrating. Using classes felt more natural at the beginning, but it turns out that using a different API than 100% of documentation out there means you always have to translate in your head. It's not great.

I would propose to get rid of these libraries and to switch back to the official way to define components. It will cause a bit of work now, but I'm sure it will simplify maintenance in the long run.

### vue-color, vue-markdown and vuex-oidc

Vue 3 support is uncertain for these libraries. We'll need to check and potentially find alternatives.

(I read somewhere in an issue comment that vuex-oidc works with Vue3)

### vue-custom-element

I don't think it supports Vue 3. But there is now an [offial package for custom elements](https://github.com/vuejs/vue-web-component-wrapper). We'll need to migrate.

### vuex, vue-router and vue-select have a Vue 3 version

We'll need to upgrade/migrate.

PR #1123, opened 2022-03-02
Replace all "class" components with function components

martinmaillard:
As described in https://github.com/zazuko/cube-creator/issues/1100, getting rid of `vue-property-decorator`, `vue-class-component` and `vuex-class` will simplify the migration to Vue 3, and maintenance of cube-creator in the long run.

This PR refactors all components to use the official "options API" (because it's the closest to the class API we were using).


martinmaillard:
Tests are failing because `https://register.ld.admin.ch/opendataswiss/category/agriculture` doesn't have class `http://schema.org/DefinedTerm` in the e2e `dataset/update` test. Not sure why this is happening now 🤷

fuseki/shared-dimensions.trig
```diff
@@ -13,8 +13,11 @@ base <https://ld.admin.ch/cube/dimension/>
 
 <shared-dimensions> a void:Dataset .
 
-graph <https://register.ld.admin.ch/opendataswiss/categories> {
-  <https://register.ld.admin.ch/opendataswiss/category/agriculture> a schema:DefinedTerm .
+graph <https://lindas.admin.ch/sfa/opendataswiss> {
+  <https://register.ld.admin.ch/opendataswiss/category/agriculture> a schema:DefinedTerm ;
+    schema:inDefinedTermSet <https://register.ld.admin.ch/opendataswiss/category> ;
+    schema:name "Agriculture"@en ;
+  .
```
martinmaillard, line 20: This is required so that the category appears in the dataset form.

ui/package.json
```diff
@@ -36,6 +36,7 @@
     "@uppy/aws-s3-multipart": "^2.2.0",
     "@uppy/core": "^2.1.4",
     "@uppy/vue": "^0.4.5",
+    "@vue/composition-api": "^1.4.9",
```
martinmaillard, line 39: This is a backport from Vue 3 that will make the migration easier.

martinmaillard:
@tpluscode I don't expect you to go through the whole code diff. I just hope you'll be ok with the general idea of switching back to "official" Vue components instead of the classes we were using.

fuseki/shared-dimensions.trig
tpluscode: And it wasn't until now?

tpluscode (approved):
Slightly sad to see class components but if that's the way forward :)

LGTM, the app works as it used to.

ui/src/components/CsvSourceMapping.vue
```diff
@@ -18,9 +18,11 @@
...
             </o-button>
             <o-dropdown position="bottom-left" class="has-text-weight-normal" aria-role="list">
-              <button class="button is-text is-small" slot="trigger">
-                <o-icon icon="ellipsis-h" />
-              </button>
+              <template #trigger>
+                <button class="button is-text is-small">
+                  <o-icon icon="ellipsis-h" />
+                </button>
+              </template>
```
tpluscode, line 25: I'm curious about this

fuseki/shared-dimensions.trig
martinmaillard: No because the query only returns categories with the proper type and with a name.

It was working in the past because we were retrieving these categories from `test.lindas.admin.ch`, but not anymore since you merged the Hierarchies PR ;)

ui/src/components/CsvSourceMapping.vue
martinmaillard: The `slot` attribute has been deprecated https://v2.vuejs.org/v2/api/#slot-deprecated

PR #1126, opened 2022-03-10
fix: dcterms:identifier validation error

tpluscode:


martinmaillard:
I think `dcterms:identifier` is still needed by the CKAN harvester endpoint. Is it populated on publication using the one from the project?

tpluscode:
Yes, hence validation errors. If they got out of sync, the resulting cube would have 2 values.

I figure there is no point in duplicating them anyway

martinmaillard (approved):

l00mi:
Do I get this correct that is making a clone in the UI vs making a clone in publishing.

My question then, how could the UI clone get out of sync then? Should it not always be changed immediately?

tpluscode:
I don't follow. What clone?

PR #1132, opened 2022-03-16
fix: correct type to find hierarchies

tpluscode:
A simple mistake which caused hierarchies from inside cubes to show up in the main listing. Instead of `meta:Hierarchy` we should use `md:Hierarchy` which is a more specific type used by the API itself

Similar to how we have `md:SharedDimension` as an API specific (informal) subclass of `schema:DefinedTermSet`

### DIFF
--- a/.changeset/tall-pots-drop.md
+++ b/.changeset/tall-pots-drop.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/shared-dimensions-api": patch
+---
+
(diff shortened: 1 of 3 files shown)


l00mi (commented):
shall we not directly open a hierarchies namespace?

tpluscode:
I don't understand?

l00mi (approved):

PR #1133, opened 2022-03-16
fix: align button style

tpluscode:
I think I missed this when applying the buefy->oruga changes


martinmaillard (approved):
I think I fixed it in #1130, but we can merge this before

l00mi (approved):
🎉

PR #1137, opened 2022-03-18
Hierarchy details page

tpluscode:
This adds mode details to hierarchy detail page

For now I used a 3rd party tree component but maybe we'd want to render it on our own, possibly making styling easier?


ui/src/store/modules/sharedDimensions.ts
```diff
@@ -70,6 +80,10 @@ const mutations: MutationTree<SharedDimensionsState> = {
   storeHierarchies (state, collection) {
     state.hierarchies = collection || null
   },
+
+  storeHierarchy (state, hierarchy) {
+    state.hierarchy = hierarchy || null
+  },
```
tpluscode, line 86: Maybe this now warrants a separate store model for hierarchies?

martinmaillard: I would say yes

martinmaillard:
Since you are using an external library to render it, I was slightly surprised by the bare-bone-ness of the tree 😄 Which features does it provide that a simple component wouldn't?

I'll let you be the judge of how this page should look like. From my understanding, it's not a critical page yet.

tpluscode:
@martinmaillard indeed. I was almost going to get rid of the component since it looks like crap but decided we might do it as a next step. I would be very happy if you could propose a oruga-feel structure for a tree which we might build ourselves.

tpluscode:
You are right @martinmaillard. My mistake that I represent the hierarchy definition and not the actual hierarchy as the tree. 

I thought to dereference the root and first level automatically and then have the deeper level loaded on demand. And all level probably in batches

martinmaillard:
I pushed a tree that presents a sample of the hierarchy data.

I kept the hierarchy definition for now but I'm not sure it's necessary to keep it on the page like that.

tpluscode:
> I kept the hierarchy definition for now but I'm not sure it's necessary to keep it on the page like that.

I think I would keep it but only as an outline

```
Dimension: Country
Root: Switzerland
Levels:
  - Canton
    - District
```

PR #1141, opened 2022-03-22
feat(ui): add simple text filter for dimension mapping values

martinmaillard:
The filtering algorithm is very naïve at the moment: lowercase + startsWith. We can implement something more complex if needed.


l00mi:
Nice idea to group the filters.. it was anyway to prominent the: mapped / not mapped. Obviously it scales not for very big datasets, being client only. So something to keep in mind to fix later. https://github.com/zazuko/cube-creator/issues/1142

PR #1147, opened 2022-03-24
clean-up and rearrange of some ui fields.

l00mi:
@tpluscode can you please finish this PR with creating the necessary tabs apis/core/bootstrap/shapes/column-mapping.ts
* main: Cube Metadata
* otab: Opendata.swiss
* atab: About

I tried to translate it over from the shared dimensions, but as is a different style of defining the shapes I failed.


l00mi (commented):
@tpluscode please merge

PR #1150, opened 2022-03-30
feat: hierarchy in dimension metadata

tpluscode:
This adds hierarchy UI to dimension metadata form. 

I chose to duplicate the hierarchy shape because dereferencing it and making a copy was more effort than seemed necessary in the UI, especially given that the shape does actually differ from what is needed in the UI to create hierarchies:

1. Adds form to choose the original hierarchy
2. All other fields until selection is made and shape copied (see `sh1:hideWithoutObjects` property annotation)
3. All copied fields are now read-only
4. Root dimension, root and property are displayed differently

https://user-images.githubusercontent.com/588128/160812049-d47c8407-d03b-41cf-9d34-3a4ef07acf91.mp4

Of other notable changes, I added support for `dash:readOnly` to text field, as was necessary


l00mi:
Looking good, lets have a look on a live version at 14:00.

martinmaillard (approved):

PR #1153, opened 2022-04-04
refactor: only poll jobs when necessary

martinmaillard:
With the new structure of the app, it is a waste to poll all jobs every
3 seconds on any page of the project.

With this change, jobs will only be polled on jobs listing pages. And
the detail page of a job will only poll the job itself, not the full
collection.

### DIFF
--- a/ui/src/views/Job.vue
+++ b/ui/src/views/Job.vue
@@ -62,43 +62,48 @@
 </template>
 
 <script lang="ts">
-import { defineComponent } from 'vue'
-import { Job } from '@cube-creator/model'
 import type { CreativeWork } from '@rdfine/schema'
 import { schema } from '@tpluscode/rdf-ns-builders'
+import { computed, defineComponent, shallowRef } from 'vue'
+import { useStore } from 'vuex'
+import { useRoute } from 'vue-router'
+
+import { api } from '@/api'
 import BMessage from '@/components/BMessage.vue'
 import ExternalTerm from '@/components/ExternalTerm.vue'
-import JobStatus from '../components/JobStatus.vue'
-import LoadingBlock from '../components/LoadingBlock.vue'
-import MarkdownRender from '../components/MarkdownRender.vue'
-import ValidationReportDisplay from '../components/ValidationReportDisplay.vue'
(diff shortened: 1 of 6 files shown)


ui/src/views/Job.vue
```diff
@@ -62,43 +62,45 @@
...
+    const jobId = route.params.jobId
+    const job = computed(() => findJob(jobId))
+    const link = computed(() => job.value?.link?.id.value)
+    const createdDate = computed(() => job.value?.created.toLocaleString() ?? '')
 
-    job (): Job | undefined {
-      const jobId = this.$route.params.jobId
-      return this.findJob(jobId)
-    },
+    usePolling(() => store.dispatch('project/fetchJob', jobId))
```
tpluscode, line 94: What magic causes `onMounted/onUnmounted` to be called in the right moments for the right component?

ui/src/store/modules/project.ts
```diff
@@ -373,6 +380,13 @@ const mutations: MutationTree<ProjectState> = {
     state.jobCollection = collection ? serializeJobCollection(collection) : null
   },
 
+  storeJob (state, job) {
+    if (!state.jobCollection) throw new Error('JobCollection not loaded')
+
+    state.jobCollection.member = state.jobCollection.member
+      .map((storedJob) => storedJob.id.equals(job.id) ? Object.freeze(job) : storedJob)
```
tpluscode, line 387: `state.jobCollection.member` gets replaced when storing job? I expected a new field in the state

ui/src/views/Job.vue
martinmaillard: I think that should help https://vuejs.org/guide/extras/composition-api-faq.html#what-is-composition-api

In summary, Vue's "new" composition API allows for better reuse between components. It looks a bit like React hooks if you've ever used them (although it doesn't work the same).

ui/src/store/modules/project.ts
martinmaillard: Now that you mention it, I don't even need to store it in the global state. Will change this.

tpluscode (approved):

PR #1181, opened 2022-04-14
fix: form data binding

tpluscode:
I'm shooting blind a little but I managed to improve the form, slightly defective since vue3 update

Tl;dr; the graph passed to the pointer would be repeatedly set to the form element, causing it to "reset". This is why the tab would reverted to the initial as well as what happens in #1179, that switching between SHACL form and RDF editor would not preserve the changes.

What I did is to remove a little bit of the magical reactivity and instead use watchers to update the form only when necessary

### DIFF
--- a/ui/src/components/HydraOperationForm.vue
+++ b/ui/src/components/HydraOperationForm.vue
@@ -1,19 +1,22 @@
 <template>
-  <form @submit.prevent="$emit('submit', clone)">
-    <cc-form :resource="clone" :shapes="shapePointer" no-editor-switches />
+  <form @submit.prevent="$emit('submit', value)">
+    <div v-if="!shape || !resource">
+      <loading-block />
+    </div>
+    <div v-else>
+      <cc-form :resource="value" :shapes="shapePointer" no-editor-switches />
 
-    <loading-block v-if="!shape" />
+      <hydra-operation-error :error="error" :shape="shape" class="mt-4" />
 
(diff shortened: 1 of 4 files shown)


ui/src/components/HydraOperationForm.vue
```diff
@@ -68,12 +71,40 @@ export default defineComponent({
...
   computed: {
-    clone (): GraphPointer | null {
-      const { resource } = this
+    _submitLabel (): string {
+      return this.submitLabel ?? this.operation.title ?? 'Save'
+    },
+  },
 
+  methods: {
+    clone (resource: GraphPointer | undefined | null) {
```
cristianvasquez, line 105: Question: What's the purpose of cloning the Clownface?

tpluscode: God question. This is a consequence of what I think was bad design choice for Alcaeus. In the current version it keeps all dereferenced resources in a single dataset. This was meant to offer easy correlation of the resource representation and `ApiDocumentation` resource. Something I intend to change in a next major release.

The clone extracts only the actual resource triples to its own separate dataset

cristianvasquez (approved):
I cannot comment meaningfully since I don't know much of the context, or how it's used 

I see Clownface instances passing around, and I wonder if Vue.js is able to know when they change to trigger rendering. If it cannot, then I would make use of `:key`

PR #1185, opened 2022-04-19
Hierarchies uris canonical

tpluscode:
This backs out of earlier attempt to mitigate the URL rewriting in shared dimension API responses

When working with hierarchies we should expect canonical identifiers of shared dimensions. This further modifies the `GET` handler of a hierarchy so that anything matching the `/dimension/*` pattern (except `/dimension/hierarchy*`) not to be rewritten in the response.

Additional, the hierarchy form would load shared dimension/terms from SPARQL directly to skip any of rewriting

### DIFF
--- a/apis/shared-dimensions/lib/domain/shared-dimension.ts
+++ b/apis/shared-dimensions/lib/domain/shared-dimension.ts
@@ -29,6 +29,10 @@ export async function create({ resource, store }: CreateSharedDimension): Promis
     throw new DomainError('Missing dimension identifier')
   }
 
+  if (identifier === 'hierarchy') {
+    throw new DomainError('`hierarchy` is not an allowed shared dimension identifier')
+  }
+
   const termSetId = newId(env.MANAGED_DIMENSIONS_BASE, `dimension/${identifier}`)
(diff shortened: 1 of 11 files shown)


apis/shared-dimensions/lib/domain/shared-dimension.ts
```diff
@@ -29,6 +29,10 @@ export async function create({ resource, store }: CreateSharedDimension): Promis
     throw new DomainError('Missing dimension identifier')
   }
 
+  if (identifier === 'hierarchy') {
+    throw new DomainError('`hierarchy` is not an allowed shared dimension identifier')
+  }
```
tpluscode, line 34: Since the decision to rewrite is based on the URIs, I added an additional restriction so that one cannot create a shared dimension called `/hierarchy`. That would interfere with the [`rewriteSharedDimensionTerms` function](https://github.com/zazuko/cube-creator/pull/1185/files#diff-7e20c4d710b7de02ee4a6072fbc7e09077f2b2d9020235d2857d14fb829de9e9R10)

tpluscode:
Closing in favour of #1200

PR #1186, opened 2022-04-19
fix: smaller CONSTRUCT

tpluscode:
Reduces the size of `CONSTRUCT` by combing patterns of common subject. To prevent an excess of duplicate triples, which can bring down the database, the `WHERE` patterns are separated into a series of UNION to prevent a massive cartesian

### DIFF
--- a/.changeset/thin-frogs-explode.md
+++ b/.changeset/thin-frogs-explode.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/shared-dimensions-api": patch
+---
+
+`Payload too large` when exporting shared dimensions (fixes #1184)
--- a/apis/shared-dimensions/lib/domain/shared-dimension.ts
+++ b/apis/shared-dimensions/lib/domain/shared-dimension.ts
@@ -163,11 +163,9 @@ export async function getExportedDimension({ resource, store, client = streamCli
     .filter(shape => !shape.in(sh.node).terms.length) // exclude nested shapes
   const patterns = getPatternsFromShape(shapes)
 
-  const quads = await CONSTRUCT`${patterns}`
+  const quads = await CONSTRUCT.WHERE`${patterns}`
     .FROM(store.graph)
-    .WHERE`
-      ${patterns}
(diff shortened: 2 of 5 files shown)


cristianvasquez (commented):
(Only if this approach turns out to  be slow or high in memory consumption)

I haven't tested it, but I imagine that using `Union` over several subjects will force the optimizer to compute the trees for each scope and only then join them together. If that is the case, I would perhaps go for a basic query with all the properties that come from the shape, optional.

Example: 
```sparql
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX sh: <http://www.w3.org/ns/shacl#>
PREFIX dcterms: <http://purl.org/dc/terms/>
PREFIX hydra: <http://www.w3.org/ns/hydra/core#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>

construct {  
	?s 	rdf:type ?type ;
		schema:name ?SchemaNameO ;
		schema:validFrom ?validFromO ;
		skos:broader ?skosBroaderO ;
		schema:inDefinedTermSet ?inDefinedTermSetO ;
		schema:identifier ?identifierO .
}
where {
	?s 	rdf:type ?type .
  OPTIONAL {		
	?s schema:name ?SchemaNameO .
  }		
  OPTIONAL {		
	?s schema:validFrom ?validFromO .
  }
  OPTIONAL {
  	?s schema:validFrom ?validFromO .
  }		
  OPTIONAL {
  	?s skos:broader ?skosBroaderO .
  }
  OPTIONAL {
	?s schema:inDefinedTermSet ?inDefinedTermSetO .
  }
  OPTIONAL {
	?s schema:identifier ?identifierO .
  }
  FILTER(?s IN (<https://ld.admin.ch/cube/dimension/fabcofogtest3/gf04>, <https://ld.admin.ch/cube/dimension/fabcofogtest3/gftot>)) .
}
```

PR #1191, opened 2022-04-21
fix: publishing cube with sh:or

tpluscode:
The query built to retrieve cube metadata was previously meant to exclude concepts which appear in `sh:in` lists

The filter used inadvertently also excluded triples deep inside `sh:or`, which resulted in dangling blank nodes without the expected `sh:datatype`

### DIFF
--- a/cli/lib/metadata.ts
+++ b/cli/lib/metadata.ts
@@ -46,8 +46,8 @@ export async function loadDataset(jobUri: string, Hydra: HydraClient) {
 interface Params {
   jobUri: string
   endpoint: string
-  user: string
-  password: string
+  user?: string
+  password?: string
 }
 
 interface QueryParams {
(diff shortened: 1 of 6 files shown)


cli/lib/metadata.ts
```diff
@@ -78,9 +78,11 @@ function sourceCubeAndShape({ project, revision, cubeIdentifier }: QueryParams)
...
 
         # exclude properties of sh:in values (concept dimensions)
-        FILTER (NOT EXISTS {
-          [] ${rdf.first} ?s .
-        })
+        MINUS {
+          ?cube ${cube.observationConstraint}/${sh.property} ?prop .
+          ?prop ${sh.in}/${rdf.rest}*/${rdf.first} ?s .
+          ?s ?conceptP ?conceptO .
+        }
```
tpluscode, line 85: This is the most important change

l00mi (approved):

PR #1214, opened 2022-05-17
fix: no enormous CONSTRUCT

tpluscode:
We are still having issues with `413 Payload Too Large` when exporting dimensions with many terms

This is because currently the export happens in two queries. First one find all properties of shared terms from their persisted shapes. The second then does a massive CONSTRUCT with those exact patterns. It quickly grows out of proportion and it simply unnecessary. Here's a sample diff of how this PR changes the export to be executed in only one query

```diff
CONSTRUCT {
  ?s ?p ?o 
}
FROM <https://lindas.admin.ch/cube/dimension>
WHERE {   
  {     
    SELECT ?rootShape  WHERE {   
      ?term schema:inDefinedTermSet* <https://ld.admin.ch/cube/dimension/baqua> .   
      ?rootShape sh:targetNode ?term .    
      MINUS {    
        # Exclude shapes which are children of property shapes   
        ?propertyShape sh:node ?rootShape .  
      }    
    }   
  }
- ?rootShape (!sh:targetNode)* ?s .
+ ?rootShape (!sh:targetNode)* ?shape .

+ ?shape sh:property ?shProp .
+ ?shProp sh:path ?p .
+ ?shape sh:targetNode ?s .
  ?s ?p ?o .
}

-CONSTRUCT {
- <https://ld.admin.ch/cube/dimension/baqua/ag>
-       a md:SharedDimensionTerm , schema:DefinedTerm , hydra:Resource ;
-       schema:hasPart <https://ld.admin.ch/dimension/bgdi/inlandwaters/bathingwater/CH19006> .
-
- # and all else explicit
-}
-FROM <https://lindas.admin.ch/cube/dimension>
-WHERE {
- <https://ld.admin.ch/cube/dimension/baqua/ag>
-       a md:SharedDimensionTerm , schema:DefinedTerm , hydra:Resource ;
-       schema:hasPart <https://ld.admin.ch/dimension/bgdi/inlandwaters/bathingwater/CH19006> .
-
- # and all else explicit
-}
```

Rather than building all the resource patterns in the second `CONSTRUCT`, they are simply rebuilt from the `sh:property/sh:path` of the shapes inline in one query

### DIFF
--- a/apis/shared-dimensions/lib/domain/shared-dimension.ts
+++ b/apis/shared-dimensions/lib/domain/shared-dimension.ts
@@ -7,13 +7,13 @@ import { dcterms, hydra, rdf, schema, sh } from '@tpluscode/rdf-ns-builders'
 import { md, meta } from '@cube-creator/core/namespace'
 import { DomainError } from '@cube-creator/api-errors'
 import httpError from 'http-errors'
-import { CONSTRUCT, sparql, SparqlTemplateResult } from '@tpluscode/sparql-builder'
+import { CONSTRUCT, sparql } from '@tpluscode/sparql-builder'
 import { StreamClient } from 'sparql-http-client/StreamClient'
 import TermMap from '@rdfjs/term-map'
 import { streamClient } from '../sparql'
 import { SharedDimensionsStore } from '../store'
 import env from '../env'
-import { getPatternsFromShape, resourceShapePatterns } from '../resource'
+import { resourceShapePatterns } from '../resource'
 import { newId, replace } from './resource'
 
 export { importDimension } from './shared-dimension/import'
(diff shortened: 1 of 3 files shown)


apis/shared-dimensions/lib/resource.ts
```diff
@@ -94,7 +96,10 @@ export async function resourceQuery(id: NamedNode, graph: NamedNode, client: Par
 }
 
 export function deleteShapesQuery(id: NamedNode, graph: NamedNode) {
-  return WITH(graph, DELETE`?s ?p ?o`.WHERE`${resourceShapePatterns(id, true)}`)
+  return WITH(graph, DELETE`?shape ?p ?o .`.WHERE`
+    ${resourceShapePatterns(id, true)}
+    ?shape ?p ?o .
```
tpluscode, line 101: Because the `resourceShapePatterns` function now does not include the `?s ?p ?o` pattern, that has to be added where necessary

apis/shared-dimensions/lib/domain/shared-dimension.ts
```diff
@@ -154,28 +149,18 @@ export async function getExportedDimension({ resource, store, client = streamCli
...
-    }
+    .WHERE`
+      ${resourceShapePatterns(dimensionAndTerms, true)}
 
-    return sparql`${union}\nUNION { ${pattern} }`
-  }, '' as SparqlTemplateResult | string)
+      ?shape ${sh.property} ?shProp .
+      ?shProp ${sh.path} ?p .
+      ?shape ${sh.targetNode} ?s .
+      ?s ?p ?o .
```
l00mi, line 160: Lets call ?s here what it is?

apis/shared-dimensions/lib/resource.ts
l00mi: Sounds good, and is clear as the VAR Name is called ?shape. And it is restricted on the specific graph.

(I feel at some point we also need SHACL shapes for delete operations, this here seems a bit dangerous a times.)

l00mi (approved):
See comments, but overall I am curious to see how this improves performance.

apis/shared-dimensions/lib/resource.ts
tpluscode: Not sure if I understood the last remark. Could you elaborate?

PR #1223, opened 2022-05-19
Project delete

tpluscode:
I found two cases when deleting resources would leave some dangling leftovers

1. When removing a column mapping, the dimension mapping would not be removed
2. When deleting the entire project, a lot was not removed

In the latter case I opted for simply dropping all graphs which share the project's URL stem (all resources are named as `https://{Project URL}/{child resource path}`

### DIFF
--- a/apis/core/test/domain/cube-projects/delete.test.ts
+++ b/apis/core/test/domain/cube-projects/delete.test.ts
@@ -1,0 +1,48 @@
+import { before, beforeEach, describe, it } from 'mocha'
+import sinon from 'sinon'
+import { expect } from 'chai'
+import { SELECT } from '@tpluscode/sparql-builder'
+import $rdf from 'rdf-ext'
+import { ccClients } from '@cube-creator/testing/lib'
+import { insertTestProject } from '@cube-creator/testing/lib/seedData'
+import { deleteProject } from '../../../lib/domain/cube-projects/delete'
+import ResourceStore from '../../../lib/ResourceStore'
+import '../../../lib/domain'
+import * as storage from '../../../lib/storage'
+
+describe('@cube-creator/core-api/lib/domain/cube-projects/delete @SPARQL', function () {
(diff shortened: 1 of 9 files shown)


apis/core/test/domain/cube-projects/delete.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    })
+    await store.save()
+
+    // then
+    const anyGraph = SELECT.DISTINCT`?g`.WHERE`graph ?g {
+      ?s ?p ?o
+    }
+
+    filter (strstarts(str(?g), "${project.value}"))`.execute(ccClients.parsingClient.query)
+    await expect(anyGraph).to.eventually.deep.equal([])
```
cristianvasquez, line 46: Could we have some data supposed to remain? 

- Given : data type A, data type B
- When: Delete data type A
- Then: Remain data type B

To check if this doesn't delete all.

cristianvasquez:
I support the simplicity of the solution.

But If we rely on URL patterns for the graphs, we could have the logic of minting graph names centralized so that one can look up the  Stem of each entity and they are not spread around. In the particular case of Projects,  https://{Project URL}/{child resource path}

it could have methods,  newGraphUri(type Project) & stem(type Project)

cristianvasquez (changes requested):
To add a test:

    Given: data type A, data type B
    When: Delete data type A
    Then: Data type B remains

To check if this method doesn't delete all after some weird URL pattern change.

tpluscode:
> But If we rely on URL patterns for the graphs, we could have the logic of minting graph names centralized [...]

Indeed, identifiers should be created [here](https://github.com/zazuko/cube-creator/blob/master/apis/core/lib/domain/identifiers.ts)

cristianvasquez:
I was not aware of that one, It was just in case :)

cristianvasquez (approved):

PR #1226, opened 2022-05-23
feat: warn when dimension mapping changes

tpluscode:


apis/core/lib/domain/dimension-mapping/update.ts
```diff
@@ -29,7 +34,10 @@ export async function update({
   dimensionMappings.changeSharedDimensions(sharedDimensions)
 
   dimensionMappings.onlyValidTerms = newMappings.onlyValidTerms
-  dimensionMappings.replaceEntries(newMappings.hadDictionaryMember)
+  const newEntries = dimensionMappings.replaceEntries(newMappings.hadDictionaryMember)
 
-  return dimensionMappings.pointer
+  return {
+    dimensionMapping: dimensionMappings.pointer,
+    hasChanges: newEntries.size > 0,
```
cristianvasquez, line 41: One could maintain the number of entries to give better feedback to the user, 

'There are 3 new entries.'

packages/model/test/BaseResource.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { blankNode } from '@cube-creator/testing/clownface'
+import { fromPointer } from '@rdfine/schema/lib/Thing'
+import { expect } from 'chai'
+import { schema } from '@tpluscode/rdf-ns-builders/strict'
+import { literal } from '@rdf-esm/data-model'
+import { BaseResourceMixin } from '../BaseResource'
+
+describe('@cube-creator/model/BaseResource', () => {
+  describe('addError', () => {
+    it('adds error when new', () => {
```
cristianvasquez, line 12: Just for curiosity, why are errors added as RDF?

l00mi: We only have a triple store to persist. How would you add them?

cristianvasquez: That was more of a question to understand the app. 

I also was trying to understand if we need to store things like errors. (Not particularly for this pull request)

cristianvasquez: I have the feeling that persisting errors can become (in the future) something difficult to manage. The drawback I see is that in our code, we will have to check:

- What are the current errors? 
- Are the errors still valid?
- Can we delete the errors? 

I don't have a clear idea yet about the model, so I'm not sure if the following applies:

 I imagine a draft area where one builds up entities (cube, dimension mapping, etc.) to be created in RDF later.

The app validates everything in the draft area with, for example, SHACL; messages are calculated and displayed at runtime (never persisted). 

If the user decides to create a 'valid' element, it is promoted from draft to 'real' in RDF.

l00mi (approved):

cristianvasquez (approved):

PR #1257, opened 2022-07-04
fix: timeout unstarted jobs

tpluscode:


cli/lib/commands/timeoutJobs.ts
```diff
@@ -37,9 +38,10 @@ export async function timeoutJobs({
     .WHERE`
       graph ?job {
         ?job a ${cc.Job} .
-        ?job ${schema.actionStatus} ${schema.ActiveActionStatus} .
+        ?job ${schema.actionStatus} ?status .
         ?job ${dcterms.modified} ?modified .
 
+        filter ( ?status ${IN(schema.ActiveActionStatus, schema.PotentialActionStatus)} )
```
l00mi, line 44: why to introduce another "IN", when we have seen some unwanted behaviour? Could we not also work with minus here?

l00mi (commented):
Not using IN, would also remove this one more dependency.

cli/lib/commands/timeoutJobs.ts
tpluscode: What unwanted behaviour have we seen?

tpluscode: I see. The difference is that in that case the intention was to eliminate the predicate. Hence, `MINUS` is the way to go.

Here we do indeed want to eliminate `?job`, which is why `IN` should be fine.

cristianvasquez (approved):

PR #1263, opened 2022-07-06
feat: add validation step of project import

tpluscode:


apis/core/lib/domain/cube-projects/import.ttl
```diff
@@ -0,0 +1,18 @@
...
+  sh:targetClass cc:CubeProject ;
+  sh:property
+    [
+      sh:path ( <https://cube-creator.zazuko.com/vocab#dataset> dcterms:identifier ) ;
+      sh:maxCount 0 ;
+      dash:hidden true ;
+      sh:message "Project metadata must not have a dcterms:identifier" ;
+    ]
+]
+.
```
tpluscode, line 18: Most other validation loads shapes from the store albeit expecting a more resource-oriented flow. For project import, I added a dedicated handler middleware to validate the import.

I'm not sure, maybe it should actually also load shapes from the store? Right now seems that it might be start of some duplication

apis/core/lib/handlers/cube-projects/import.ts
```diff
@@ -27,6 +31,22 @@ export const postImportedProject = protectedResource(
...
+    next()
+  }),
+  shaclValidate.override({
+    async parseResource(req, res) {
+      const { project, importedDataset } = res.locals
+      return clownface({ dataset: importedDataset }).node(project.id)
+    },
+    async loadShapes() {
+      return $rdf.dataset().import(fromFile(path.resolve(__dirname, '../../domain/cube-projects/import.ttl')))
+    },
```
tpluscode, line 45: That said, the current implementation of shacl middleware is a little strange in how it assumes targets based on the request.

Thus, a dedicated import shape, loaded explicitly, may be easier to manage

apis/core/lib/domain/cube-projects/import.ttl
cristianvasquez: I'm not sure either. 

What are the pros/cons of having the shape in the store?

I would inspect what the scope of this specific shape is. 

If that shape is used exclusively in the cube-creator application (part of its model) and is not used to publish public data, I would definitively keep the shape (in the form of a file or just code) in the code-creator codebase and not in the store. I think that diminishes the overall complexity.

If, on the other hand, the project entities are to be used by other applications, then the model (shape) should be shared between the apps, I think.

apis/core/lib/handlers/cube-projects/import.ts
cristianvasquez: I would pre-import all the application-related shapes, and reuse them when adequate.

apis/core/lib/domain/cube-projects/import.ttl
cristianvasquez: The application itself can publish it's shapes if required.

tpluscode: All right, thanks for challenging me on this.

The problem with some shapes is that the do not have a `sh:target*` property. Also, they are loaded on a per-request basis, based on `hydra:expects` of the current operation and not based on the data model. Slightly limiting, but not something I would like to touch now...

To reduce the duplication thought, it is definitely possible to re-use what shapes we have. I would not blindly load all shapes but for now only import the one for cube metadata resource. Might add more in the future as long as they have target properties.

packages/shacl-middleware/index.ts
```diff
@@ -86,15 +101,21 @@ export const shaclMiddleware = ({ getTargetNode, loadResource, loadResourcesType
...
-    .out(sh.path)
-  const linkedInstancesIds = resource.out(classProperties).terms.filter(r => r.termType !== 'BlankNode')
-  const linkedInstancesQuads = await loadResourcesTypes(linkedInstancesIds)
-
-  const dataset = $rdf.dataset([...resource.dataset, ...linkedInstancesQuads])
+  let dataset: DatasetCore
+  if (disableShClass) {
+    clownface({ dataset: shapes }).has(sh.class).addOut(sh.deactivated, true)
+
+    dataset = resource.dataset
```
tpluscode, line 108: Only one hawkish change here. The validation of `sh:class` constraint is problematic in a request context because it's a closed world slice.

In typical cases, there is a root resource and shape with such properties. You will find below how the types for relevant resources is loaded to satisfy the `sh:class` constraints.

Here, however, it would need to be more sophisticated because it's not right away known what is the target for any given property.

Thus, I have added an option to deactivate those properties in the import validation

tpluscode:
Good to go @cristianvasquez ?

cristianvasquez:
> Good to go @cristianvasquez ?

Yes!

Issue #1218, opened 2022-05-18
Doubling of shape contraints with publish on INT


l00mi:
With an import project:

https://int.cube-creator.lindas.admin.ch/app/cube-projects/cube-project!!ubd0066import-de-prod-1lmakvo9eek/publication
https://s.zazuko.com/fADRv3

l00mi:
Do we have any hint why this could happen?

Rdataflow:
yet another example of trippled key dimensions...
job https://gitlab.ldbar.ch/pipelines/cube-creator/-/jobs/30832
from proj https://int.cube-creator.lindas.admin.ch/app/cube-projects/cube-project!!nfi4-total-basal-area-of-main-tree-species-in-production-regions-no-hierarchy-rdwdmlrzn3o/edit
describe cube:Constraint https://s.zazuko.com/hGgLy
@tpluscode how to fix this?

Rdataflow:
currently also @kronmar is impacted by this. 
@tpluscode can you have a look at this ...

tpluscode:
I am again unable to reproduce when running the data pipeline locally. Looking at the job listing I think it looks like the issue is same version running publish simultaneously. 

It is not possible to proactively prevent parallel jobs because of a waiting period before GitLab picks up a jobs and the pipeline actually starts. What I can try is to add a runtime check to forcibly stop a job if a newer publication of same cube version starts

PR #1266, opened 2022-07-18
fix: job timeout ignored jobs without `dcterms:modified`

tpluscode:
While investigating #1218 I found that there are still numerous jobs "blinking"

That was because the query checked only for `dcterms:modified` which is set one job starts and the status gets updated. A new job, which never got the first update would not match.

### DIFF
--- a/cli/lib/commands/timeoutJobs.ts
+++ b/cli/lib/commands/timeoutJobs.ts
@@ -39,7 +39,7 @@ export async function timeoutJobs({
       graph ?job {
         ?job a ${cc.Job} .
         ?job ${schema.actionStatus} ?status .
-        ?job ${dcterms.modified} ?modified .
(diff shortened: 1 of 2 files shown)


cli/lib/commands/timeoutJobs.ts
```diff
@@ -39,7 +39,7 @@ export async function timeoutJobs({
       graph ?job {
         ?job a ${cc.Job} .
         ?job ${schema.actionStatus} ?status .
-        ?job ${dcterms.modified} ?modified .
+        ?job ${dcterms.modified}|${dcterms.created} ?modified .
```
tpluscode, line 42: The fix is simple, to inspect either `modified` OR `created` and cancel those jobs

cristianvasquez (approved):
looks ok to me!

PR #1268, opened 2022-07-18
feat: separate status and icon for canceled jobs

### LINKED ISSUE #1218: Doubling of shape contraints with publish on INT

tpluscode:
To properly fix #1218, I figure I would need a separate "canceled" status for jobs neither failed, nor succeeded.

Might as well assign that status to timeout jobs. Especially that I noticed that some "blinking" jobs were actually successful but somehow the update did not register...

### DIFF
--- a/.changeset/hip-tools-work.md
+++ b/.changeset/hip-tools-work.md
@@ -1,0 +1,6 @@
+---
+"@cube-creator/cli": patch
+"@cube-creator/ui": patch
+---
+
+Add dedicated status for timed-out jobs and show them differently in the UI
--- a/.github/workflows/test-pipelines.yaml
+++ b/.github/workflows/test-pipelines.yaml
@@ -3,47 +3,21 @@ name: Test pipelines
 on: push
 
 jobs:
-  transform:
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-    steps:
-      - uses: actions/checkout@v2
-      - uses: ./.github/workflows/setup-env
-      - run: yarn test:cli:transform
-        env:
-          AUTH_RUNNER_CLIENT_SECRET: ${{ secrets.AUTH_RUNNER_CLIENT_SECRET }}
-      - name: logs on fail
-        if: ${{ failure() }}
-        run: lando logs
-      - name: Codecov
-        uses: codecov/codecov-action@v1.0.5
-        with:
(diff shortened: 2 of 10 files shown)


l00mi (commented):
Not sure if I understand it correctly, but the jobs will only be timedout when a transform, publish or import job is triggereD?

tpluscode:
Not yet. This only adds a new "canceled" status to distinguish them from failed jobs.

Regarding the second part of your question, the fix will only be for publish jobs because we post to one graph. Transform and import always rewrite the entire working copy in cube creator. Not harm done if jobs run in parallel

cristianvasquez:
I'm not sure I understand how a new state helps to prevent https://github.com/zazuko/cube-creator/issues/1218, which apparently happens when multiple parallel jobs work on the same resources. 

Shouldn't we interact with the jobs API directly, having locking mechanisms to handle concurrency? 

https://docs.gitlab.com/ee/ci/resource_groups/index.html#pipeline-level-concurrency-control-with-cross-projectparent-child-pipelines

cristianvasquez:
What if it fails to persist the Cancel state?

tpluscode:
Well, yes, it does not help prevent. This is only an indicator for the UI as to what happened to a job

Thanks for the link, quite interesting but I don't think this helps us.

> If other deployment pipelines are running, GitLab waits until those pipelines finish before running another one.

This will not prevent the result of duplicate cubes. It will delay the second publish job but ultimately the cube will be published multiple times. What I we need is to break a pipeline so that it does not write to Lindas

> What if it fails to persist the Cancel state?

Nothing. The "timeout jobs" job runs periodically and it will try again at the next attempt

cristianvasquez:
But wouldn't be ok to show the data of Gitlab pipelines' API directly in the UI? it might provide controls to cancel all kinds of jobs related to publishing in Lindas. Does it make sense?

tpluscode:
Well, not directly in the UI because of authorisation. 

But yea, I like the idea. We send a trigger to GitLab and immediately and get a job URL back. It should be possible to cancel previous jobs at GitLab if someone triggers to publish the same version again. Thanks!

PR #1270, opened 2022-07-25
fix: API performance of common requests

tpluscode:
I finally figured out one problem with API performance

There was unchecked query which should only run a basic pattern with `FROM` clauses.

```sparql
SELECT ?parent ?type
FROM <link-1> 
FROM <link-2>
WHERE {
  graph ?parent {
    ?parent a ?type
  }
}
```

However, that should only be executed when there are links to look up. The problem was that in such cases, the query would have been completely unbound (no `FROM`) resulting in huge results.

Additionally, I also improved th query before that, which actually finds these linked resources

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     ]
   },
   "resolutions": {
-    "@types/eslint": "^8"
+    "@types/eslint": "^8",
+    "hydra-box": "hypermedia-app/hydra-box#loader-req"
   }
 }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -67,6 +67,7 @@
     ]
   },
   "resolutions": {
-    "@types/eslint": "^8"
+    "@types/eslint": "^8",
+    "hydra-box": "hypermedia-app/hydra-box#loader-req"
```
tpluscode, line 71: Requires https://github.com/zazuko/hydra-box/pull/102

cristianvasquez (approved):

PR #1273, opened 2022-08-18
Forms fixes

tpluscode:
Maybe I went a little overboard here but changing some of the form view components into custom elements make them that much more easier to manage and data-bind


ui/src/components/JobForm.vue
```diff
@@ -3,14 +3,13 @@
...
-      :error="error"
-      :is-submitting="isSubmitting"
-      @submit="onSubmit"
-      :show-cancel="false"
+    <cc-hydra-operation-form
+      :operation.prop="operation"
+      :resource.prop="resource"
+      :shape.prop="shape"
+      :error.prop="error"
+      :submitting.prop="isSubmitting"
```
cristianvasquez, line 11: The name is-submitting is more evident than `:resource.prop`

Why the change?

tpluscode: Because vue insisted on keeping the kebab syntax as `is-submitted` and not `isSubmitted` as the property is called. the binding did not work that way

cristianvasquez: state returning whatever the form needs to communicate.

tpluscode: Ok, I see the comment update now and maybe I did not understand the question...

ui/src/customElements/FormSubmitCancel.ts
```diff
@@ -0,0 +1,4 @@
+import { createCustomElement } from '@/forms/custom-element'
+import FormSubmitCancel from '@/components/FormSubmitCancel.vue'
+
+createCustomElement('cc-form-submit-cancel')(FormSubmitCancel)
```
cristianvasquez, line 4: Just to tell you,  one can declare them globally, like https://github.com/zazuko/cube-viewer/blob/489063e6d89f4c7ab2c02345184916ad73577d1a/vue.config.js#L32

But I prefer being explicit on it, like you did :)

tpluscode: That's not the same. The line you linked is so that due does not treat custom elements as vue components.

Here it's registering the element with the browser itself

ui/src/components/JobForm.vue
cristianvasquez: I took that idea from: https://vueuse.org/core/useAsyncState/

cristianvasquez (approved):

ui/src/customElements/FormSubmitCancel.ts
cristianvasquez: Ok, I need to read more about that. ;)

tpluscode: It wraps the native [`customElements.define()`](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define)

l00mi (commented):
👍

PR #1274, opened 2022-08-22
ci: remove boilerplate to set up lando

tpluscode:
https://github.com/tpluscode/action-setup-lando

### DIFF
--- a/.github/workflows/setup-env/action.yml
+++ b/.github/workflows/setup-env/action.yml
@@ -16,7 +16,7 @@ runs:
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v1
     - name: Start site
-      uses: tpluscode/action-setup-lando@v0.1.2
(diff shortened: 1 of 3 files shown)


ludovicm67 (approved):
The only small issue I can see is the GitHub Action logs, where we can see a famous `[object Object]` output instead of something relevant, like here: https://github.com/zazuko/cube-creator/runs/7946622683#step:3:732

To fix this, you will need to make sure that a string is passed to `core.info` here: https://github.com/tpluscode/action-setup-lando/blob/67a5b7efc6cf5cdcb81de5d93de0f75cf9d30c5a/index.js#L20-L24

But in the current situation, it is working as expected from what I can see and looks good to me ; that's why I'm approving this PR :)

tpluscode:
Oh curious. Good catch.
It was working in 0.1.1 but I must have broken it :)

PR #1287, opened 2022-09-01
fix: performance of cube-projects resource

tpluscode:
TL;DR; a query was generated to load projects' mapping resources in a single request. That request used many `FROM` claueses

```sparql
CONSTRUCT { ?s ?p ?o }
FROM <A>
FROM <B>
FROM <C>
FROM <D>
# FROM ...
WHERE { ?s ?p ?o }
```

This proved to be slow as the number of graphs grow. On the other hand this is fast

```sparql
CONSTRUCT { ?s ?p ?o }
WHERE {
  { GRAPH <A> { ?s ?p ?o } }
  UNION
  { GRAPH <B> { ?s ?p ?o } }
  UNION
  { GRAPH <C> { ?s ?p ?o } }
  UNION
  { GRAPH <D> { ?s ?p ?o } }
  # UNION ...
}

### DIFF
--- a/.changeset/empty-shoes-smile.md
+++ b/.changeset/empty-shoes-smile.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/core-api": patch
+---
+
+Improve performance of queries, particularly the loading of projects list (fixes #1281)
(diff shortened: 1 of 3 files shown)


l00mi (approved):

l00mi:
Are there any Side effects to be expected with this PR?

tpluscode:
Don't think so. The important change is [here](https://github.com/hypermedia-app/creta/pull/462/commits/7effd299eb985aebb855cb0c0011e9f958c78c98).

The queries should be equivalent whether using `FROM` or `UNION GRAPH`

PR #1296, opened 2022-09-21
fix: unable to edit shared dimension term

tpluscode:


ortnever:
Hello @tpluscode , 
Did you find a solution for this bug? Is that possible to deploy it on int? I need to do some corrections.
Thank you.

l00mi (approved):

PR #1313, opened 2022-10-12
fix: no blank cube warnings

tpluscode:
I noticed one more small issue, that instead of removing an error completely when it no longer applies to a given cube, empty yellow rows remain in the cube designer.

image

That is because the SPARQL removed the error details but not the error node itself. 

```turtle
<dataset>
  schema:error <dataset#error-id> .

# This was removed but the above was not
<dataset#error-id>
  schema:description "Displayed on UI" .
```

This PR makes sure that the errors are properly cleaned

### DIFF
--- a/cli/test/lib/commands/transform.test.ts
+++ b/cli/test/lib/commands/transform.test.ts
@@ -24,8 +24,9 @@ describe('@cube-creator/cli/lib/commands/transform', function () {
   })
 
   describe('successful job', () => {
-    const job = `${env.API_CORE_BASE}cube-project/ubd/csv-mapping/jobs/test-job`
-    const expectedGraph = $rdf.namedNode(`${env.API_CORE_BASE}cube-project/ubd/cube-data`)
+    const projectNs = namespace(`${env.API_CORE_BASE}cube-project/ubd/`)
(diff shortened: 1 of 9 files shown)


cli/test/lib/commands/transform.test.ts
```diff
@@ -46,6 +47,20 @@ describe('@cube-creator/cli/lib/commands/transform', function () {
...
 
+    it('removes previous errors from dataset', async () => {
+      const datasetGraph = projectNs.dataset
+
+      const hasPreviousErrors = ASK`
+        <dataset> schema:error <dataset#non-unique-observations> .
+        <dataset> schema:error <dataset#missing-values-error> ; ${schema.description} "Test message" .
+      `
+        .FROM(datasetGraph)
+        .prologue`BASE ${projectNs()}`
```
cristianvasquez, line 58: Just curious, what is `prologue`?

cristianvasquez (approved):

cli/test/lib/commands/transform.test.ts
tpluscode: That's what SPARQL calls in the spec the part with prefixes and `BASE`

PR #1322, opened 2022-11-03
feat: shared dimension abberviations

tpluscode:


tpluscode:
@l00mi quick opinion, which would be better

image

or inline

image

l00mi:
Gut feeling goes for the first option, especially as you bring a good point. If the Titles will be used in future for Front-End we should also show them in all languages.

But potentially we should introduce also the language drop-down here.

tpluscode:
Right, I missed the fact that the name is also multilingual but we only display one. Yes, this should be more like the cube designer, where you can switch the language...

Would you keep the abbreviations displayed in all languages now?

l00mi:
No, if we have a "language switch" i propose to show one language only.

tpluscode:
Then question is whether we add that dropdown now

l00mi:
@tpluscode one more thing, if you have this already in focus.

Please hide the 'deprecated' shared dimensions, and add a checkmark on top to show them. Simply:

> [ ] deprecated (which by default is not checked)

tpluscode:
https://user-images.githubusercontent.com/588128/201070690-dd43114f-da7d-4afe-943d-86419bd7b8b7.mov

cristianvasquez (approved):
No comments on this one!

PR #1327, opened 2022-11-10
New cube meta

tpluscode:


.changeset/four-panthers-flash.md
```diff
@@ -0,0 +1,10 @@
+---
+"@cube-creator/core-api": patch
+---
+
+Added new fields to cube metadata:
+ - `Planned update https://schema.ld.admin.ch/datasetNextDateModified>`
```
tpluscode, line 6: I'm not particularly happy with putting the `https://schema.ld.admin.ch/` namespace directly into code but I don't have a good idea for a better option

apis/core/lib/domain/dataset/update.ts
```diff
@@ -55,5 +58,10 @@ export async function update({
     })
   })
 
+  const project = await store.get(await findProject({ dataset }), { allowMissing: true })
+  project
+    ?.deleteOut(lindasSchema.datasetNextDateModified)
+    .addOut(lindasSchema.datasetNextDateModified, datasetResource.out(lindasSchema.datasetNextDateModified))
+
```
tpluscode, line 65: Here is important that when a cube metadata is saved, the "next planned date" is copied to the project itself so that it will be possible to sort and display it in the project list

I would prefer event-driven or  a more flexible way to retrieve project collection but not an option in current codebase

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -135,6 +135,24 @@ ${shapeId} {
...
+      ${sh.order} 15 ;
+      ${sh.group} ${mainGroup} ;
+    ] ;
+    ${sh.property} [
+      ${sh.name} "Subtitle" ;
+      ${sh.path} ${schema.disambiguatingDescription} ;
+      ${sh.datatype} ${rdf.langString} ;
+      ${sh.languageIn} ( ${supportedLanguages} ) ;
+      ${sh.uniqueLang} true ;
+      ${sh.order} 15 ;
```
cristianvasquez, line 153: it seems `sh:order` value for "Subtitle" is the same as "Abbreviation"

apis/core/lib/domain/dataset/update.ts
cristianvasquez: For sure, that would be a big change in the codebase.

For more complex workflows, one could use an out-of-the-box job scheduler (with the added dependencies)

apis/core/lib/domain/dataset/update.ts
```diff
@@ -1,19 +1,22 @@
...
+import * as projectQuery from '../cube-projects/queries'
 
 interface AddMetaDataCommand {
   dataset: GraphPointer<NamedNode>
   resource: GraphPointer
   store: ResourceStore
+  findProject?(arg: { dataset: GraphPointer }): Promise<NamedNode | undefined>
 }
 
 export async function update({
```
cristianvasquez, line 15: This is minor, but sometimes in the code I was wondering, 'update what?', what about 'updateProject' ?

cristianvasquez (approved):
Looks ok to me!

The change to updateProject, is totally optional

apis/core/lib/domain/dataset/update.ts
tpluscode: I see. The "full picture" is like

```js
import { update } from './dataset.js'
```

But yea, as soon as you import multiple such methods one needs to become creative.

PR #1330, opened 2022-11-15
feat: search projects

tpluscode:
https://user-images.githubusercontent.com/588128/202178978-85e1135f-eb42-4f14-b126-96f11fe0b51c.mp4

### DIFF
--- a/apis/core/bootstrap/cube-projects.ts
+++ b/apis/core/bootstrap/cube-projects.ts
@@ -1,6 +1,7 @@
 import { turtle } from '@tpluscode/rdf-string'
-import { hydra, rdf } from '@tpluscode/rdf-ns-builders'
+import { hydra, rdf, rdfs, schema } from '@tpluscode/rdf-ns-builders'
 import { cc } from '@cube-creator/core/namespace'
+import { Draft, Published } from '@cube-creator/model/Cube'
 
 export const cubeProjects = turtle`
 <cube-projects> {
@@ -10,5 +11,17 @@ export const cubeProjects = turtle`
       ${hydra.property} ${rdf.type} ;
       ${hydra.object} ${cc.CubeProject}
     ] .
+}
+
+<cube-projects/status> {
+  <cube-projects/status> a ${hydra.Collection}, ${schema.DefinedTermSet} ;
+    ${hydra.member} ${Draft}, ${Published} ;
+  .
+
+  ${Published} a ${schema.DefinedTerm} ;
+    ${rdfs.label} "Published" .
+
+  ${Draft} a ${schema.DefinedTerm} ;
+    ${rdfs.label} "Draft" .
 }
 `
--- a/apis/core/bootstrap/shapes/cube-project.ts
+++ b/apis/core/bootstrap/shapes/cube-project.ts
@@ -189,7 +189,7 @@ ${shape('cube-project/update')} {
 ${shape('cube-project/search')} {
   ${shape('cube-project/search')} a ${sh.NodeShape}, ${hydra.Resource} ;
     ${sh.property} [
-      ${sh.name} "Content" ;
+      ${sh.name} "Keyword Search" ;
       ${sh.order} 10 ;
       ${sh.path} ${hydra.freetextQuery} ;
       ${sh.minCount} 1 ;
--- a/apis/core/bootstrap/shapes/cube-project.ts
+++ b/apis/core/bootstrap/shapes/cube-project.ts
@@ -185,4 +185,34 @@ ${shape('cube-project/update')} {
     ${projectProperties} ;
   .
 }
+
+${shape('cube-project/search')} {
+  ${shape('cube-project/search')} a ${sh.NodeShape}, ${hydra.Resource} ;
+    ${sh.property} [
+      ${sh.name} "Content" ;
+      ${sh.order} 10 ;
+      ${sh.path} ${hydra.freetextQuery} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+    ],
+    [
+      ${sh.name} "Author" ;
+      ${sh.path} ${dcterms.creator} ;
+      ${sh.order} 20 ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${dash.editor} ${dash.InstancesSelectEditor} ;
+      ${hydra.collection} </users> ;
+    ],
+    [
+      ${sh.name} "Status" ;
+      ${sh.path} ${schema.creativeWorkStatus} ;
+      ${sh.order} 30 ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${dash.editor} ${dash.InstancesSelectEditor} ;
+      ${hydra.collection} </cube-projects/status> ;
+    ] ;
+  .
+}
 `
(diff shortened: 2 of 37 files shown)


apis/core/bootstrap/cube-projects.ts
```diff
@@ -11,4 +12,16 @@ export const cubeProjects = turtle`
...
+  <cube-projects/status> a ${hydra.Collection}, ${schema.DefinedTermSet} ;
+    ${hydra.member} ${Draft}, ${Published} ;
+  .
+
+  ${Published} a ${schema.DefinedTerm} ;
+    ${rdfs.label} "Published" .
+
+  ${Draft} a ${schema.DefinedTerm} ;
+    ${rdfs.label} "Draft" .
+}
```
tpluscode, line 26: A static collection of statuses to populate the search dropdown

apis/core/bootstrap/shapes/cube-project.ts
```diff
@@ -185,4 +185,34 @@ ${shape('cube-project/update')} {
...
+    [
+      ${sh.name} "Status" ;
+      ${sh.path} ${schema.creativeWorkStatus} ;
+      ${sh.order} 30 ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;
+      ${dash.editor} ${dash.InstancesSelectEditor} ;
+      ${hydra.collection} </cube-projects/status> ;
+    ] ;
+  .
```
tpluscode, line 216: Shape which defines the fields for search form

apis/core/bootstrap/users.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const users = turtle`
+<users> {
+  <users> a ${hydra.Collection} ;
+    ${hydra.manages} [
+      ${hydra.property} ${rdf.type} ;
+      ${hydra.object} ${schema.Person}
+    ] .
+}
+`
```
tpluscode, line 12: This reflects a bigger change where users actually become independent resources in their own named graphs. Here, the collection is added to retrieve them

apis/core/lib/domain/cube-projects/import.ts
```diff
@@ -103,7 +101,7 @@ export async function importProject({
   const projectNode = await store.createMember(projectsCollection.term, id.cubeProject(label))
 
   const project = createMinimalProject(projectNode, {
-    creator: { id: user, name: userName },
+    creator: user,
```
tpluscode, line 104: Since we now have graphs for users, only the URI will be stored with the project now, unlike the lazy duplication we've had so far

tpluscode:
@l00mi a bigger change to be aware of is that users become their own resources. Until now the user names were simply stored with projects themselves

```turtle
graph <project> {
  <project> dcterms:creator <user> .
  <user> schema:name "John Doe" .
}
```

Now it will be like

```turtle
graph <project> {
  <project> dcterms:creator <user> .
}

graph <user> {
  <user> schema:name "John Doe" ; a schema:Person.
}
```

Projects already referred to users with URIs so I don't find this a breaking change per se.
User resources will be automatically created as users sign in. For existing users, I would run an update like this on every environment:

```sparql
prefix dcterms: <http://purl.org/dc/terms/>
prefix schema: <http://schema.org/>
prefix cc: <https://cube-creator.zazuko.com/vocab#>
 
DELETE {
  GRAPH ?project {
    ?user schema:name ?name
  }
}
INSERT {
  GRAPH ?user {
    ?user schema:name ?name ; a schema:Person .
  }
} WHERE {
  GRAPH ?project {
    ?project a cc:CubeProject ; dcterms:creator ?user .
    ?user schema:name ?name .
  }
}
```

apis/core/bootstrap/users.ts
cristianvasquez: I love this change!

apis/core/lib/auth.ts
```diff
@@ -83,7 +86,38 @@ export default async () => {
...
+
+  return (req, res, next) => {
+    if (req.user?.id && !users.has(req.user.id)) {
+      const { id, name, sub } = req.user
+      users.add(id)
+
+      req.once('end', () => {
+        DELETE`
+          GRAPH ${id} {
+            ${id} ${schema.name} ?name .
```
cristianvasquez, line 105: If users live in their own named graph, wouldn't it be cleaner to delete the graph as a whole?

patches/hydra-box+0.6.6.patch
```diff
@@ -0,0 +1,23 @@
...
+       }
+ 
+-      templateParams.addOut(property, createTermFromVariable({ template: iriTemplateNode, value }))
++      templateParams.addOut(property, createTermFromVariable({
++        template: iriTemplateNode,
++        value: isQueryParam ? value.replace(/\+/g, ' ') : value
++      }))
+     })
+ 
+     req.dataset = () => {
```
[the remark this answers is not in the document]
cristianvasquez: Could you provide me a link to learn more about these kind of patches? I haven't seen this before

ui/src/components/FormSubmitCancel.vue
```diff
@@ -17,6 +17,11 @@
           Cancel
         </o-button>
       </div>
+      <div class="control" v-if="showClear">
```
cristianvasquez, line 20: UI-wise, I think it would look better with an X instead of a button 'clear',  (optional)

image

ui/src/views/CubeProjects.vue
```diff
@@ -73,18 +99,36 @@ export default defineComponent({
...
+      if (template && e.detail?.value) {
+        const url = template.expand(e.detail.value)
+        const search = new URLSearchParams(url.substring(url.indexOf('?')))
+        const entries = [...search.entries()].filter(([, value]) => !!value && value !== '""')
+
+        this.$router.push({
+          query: Object.fromEntries(entries),
+        })
+      }
+    }
```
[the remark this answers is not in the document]
cristianvasquez: yeah, it's difficult to understand. Perhaps you could factorize it in a function with a name  ('pushSearchRoute'?) to explain the intent?

cristianvasquez (approved):
Looks ok to me!

Only (optional) comments

ui/src/components/FormSubmitCancel.vue
tpluscode: Yes, I expected you'd say that. Unfortunately, [oruga select](https://oruga.io/components/select.html) does not appear to allow that

apis/core/lib/auth.ts
tpluscode: Well, this is but not deleting the whole graph. Only `schema:name` :)

patches/hydra-box+0.6.6.patch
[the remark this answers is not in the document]
tpluscode: This is done using [patch-package](https://npm.im/patch-package)

When you modify a library in `node_modules/foo-bar`, you the run `npx patch-package foo-bar`. That will create the patch file.

Running `patch-package` in postinstall script applies the package when installed

ui/src/views/CubeProjects.vue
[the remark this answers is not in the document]
cristianvasquez: well, they don't charge for  number of chars in function names ;)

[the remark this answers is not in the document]
tpluscode: Else Java people would be richest :D

PR #1339, opened 2022-11-29
fix: clearing values when switching sh:xone form alternatives

tpluscode:
To improve the editor switch similarly to what we [do in view builder](https://github.com/zazuko/ssz-view-builder/blob/main/apps/www/src/forms/dynamicXone.js#L46)

[Cube Creator.webm](https://user-images.githubusercontent.com/588128/204589448-784f0fab-aa27-45a7-905b-a60a82f9d0a9.webm)

Since that was introduced in a breaking release of shaperone, some more changes were necessary to fix the usage of select/combobox form controls. A lot could actually be simplified now, especially that `@hydrofoil/shaperone-hydra` now comes with a baseline implementation for the `dash:AutoCompleteEditor`


cristianvasquez (approved):
Look good to me!

also the code looks cleaner

cristianvasquez:
However, UI-wise, I'm unsure if that is the user expectation that the form 'forgets' previous inputs.

cristianvasquez:
I know why. The XoR looks a bit like a tabbed options panel.
image

tpluscode:
> However, UI-wise, I'm unsure if that is the user expectation that the form 'forgets' previous inputs.

Interesting observation. I hope it's a viable compromise though. It would require deeper changes to actually clear the unwanted values of unselected xone properties. Granted, inadvertently switching the selector, forces to redo selection when you go back but at least it fixes the current problem that it is impossible to unselect from a dropdown and the validation finds a violation.

PR #1344, opened 2022-12-15
feat: add agroscope as publisher

dabo505:
Voilà. 
As the link https://register.ld.admin.ch/opendataswiss/org/agroscope showed something, I used that one instead of the staatskalender...


tpluscode (approved):
LGTM

l00mi (approved):
I changed the observedBy, @tpluscode please merge and deploy.

PR #1353, opened 2023-01-04
Upgrade GitHub Actions

ludovicm67:
This upgrades many of the used GitHub Actions and configure some permissions used for publishing packages, so that dependabot PRs can work.


tpluscode (approved):
looks good

PR #1369, opened 2023-02-01
Prevent duplicate target properties

tpluscode:
...when creating a table from multiple CSV columns


cristianvasquez (approved):
Looks good to me!

PR #1377, opened 2023-02-14
build(deps): update opentelemetry

tpluscode:


cli/lib/otel/index.ts
```diff
@@ -1,15 +1,19 @@
...
+const jaegerExporter = new JaegerExporter()
+const prometheusExporter = new PrometheusExporter()
+
 const sdk = new NodeSDK({
   // Automatic detection is disabled, see comment below
   autoDetectResources: false,
-  traceExporter: new CollectorTraceExporter(),
-  metricExporter: new CollectorMetricExporter(),
+  traceExporter: jaegerExporter,
+  metricReader: prometheusExporter,
```
ludovicm67, line 16: The exporter is automatically loaded depending on the `OTLP_TRACES_EXPORTER` environment variable if the `traceExporter` here is omitted.

Unfortunately, it's not the case for the metrics exporter, and in our infra, we want to push metrics via OTLP, so

```suggestion
  metricReader: otlpMetricExporter,
```

cli/lib/otel/index.ts
```diff
@@ -1,15 +1,19 @@
...
+import { PrometheusExporter } from '@opentelemetry/exporter-prometheus'
 import { HttpInstrumentation } from '@opentelemetry/instrumentation-http'
 import { WinstonInstrumentation } from '@opentelemetry/instrumentation-winston'
-import { Resource, envDetector, processDetector } from '@opentelemetry/resources'
+import { Resource } from '@opentelemetry/resources'
 import { NodeSDK } from '@opentelemetry/sdk-node'
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
 
+const jaegerExporter = new JaegerExporter()
+const prometheusExporter = new PrometheusExporter()
```
ludovicm67, line 10: Use the OTLP exporter instead. Leaving all settings will make it use the various env variables to set the endpoint & co

```suggestion
const otlpMetricExporter = new OTLPMetricExporter();
```

cli/lib/otel/index.ts
```diff
@@ -1,15 +1,19 @@
-import { CollectorTraceExporter, CollectorMetricExporter } from '@opentelemetry/exporter-collector'
+import { JaegerExporter } from '@opentelemetry/exporter-jaeger'
+import { PrometheusExporter } from '@opentelemetry/exporter-prometheus'
```
ludovicm67, line 2: Export via OTLP over HTTP instead (you will need to install `@opentelemetry/exporter-metrics-otlp-http`)

```suggestion
import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-http'
```

tpluscode:
Jaeger traces are back when I applied your changes but no metrics are being collected still

I'm pretty sure that I am configuring something wrong. Compared to the example you showed me @ludovicm67, I am using `NodeSDK` class. However when I added the metric reader argument, it would fail with `Error: MetricReader can not be bound to a MeterProvider again.`

On the other hand, similarly to you code, [@opentelemetry/exporter-metrics-otlp-http](https://www.npmjs.com/package/@opentelemetry/exporter-metrics-otlp-http) shows usage where the meter provider is used directly and it should "just work". 

And yet, cube-creator's metrics do show up in prometheus. Any ideas?

cli/lib/otel/metrics.ts
```diff
@@ -1,7 +1,19 @@
-import { metrics } from '@opentelemetry/api-metrics'
+import { MeterProvider, PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics'
+import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-http'
+
+const otlpMetricExporter = new OTLPMetricExporter()
+
+const metrics = new MeterProvider()
```
ludovicm67, line 6:
```suggestion
import { metrics } from '@opentelemetry/api'
```

cli/lib/otel/metrics.ts
```diff
@@ -1,7 +1,19 @@
...
 
-export const bufferObserver = meter.createValueObserver('buffer')
+export const bufferObserver = meter.createHistogram('buffer')
+
+export const metricReader = new PeriodicExportingMetricReader({
+  exporter: otlpMetricExporter,
+  exportIntervalMillis: 1000,
+})
+
+metrics.addMetricReader(metricReader)
```
ludovicm67, line 19:
```suggestion

```

This should be configured with the SDK

cli/lib/otel/index.ts
```diff
@@ -1,15 +1,12 @@
...
 import { HttpInstrumentation } from '@opentelemetry/instrumentation-http'
 import { WinstonInstrumentation } from '@opentelemetry/instrumentation-winston'
-import { Resource, envDetector, processDetector } from '@opentelemetry/resources'
+import { Resource } from '@opentelemetry/resources'
 import { NodeSDK } from '@opentelemetry/sdk-node'
 import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
 
 const sdk = new NodeSDK({
   // Automatic detection is disabled, see comment below
   autoDetectResources: false,
```
ludovicm67, line 9:
```suggestion
import { PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics'
import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-http'

export const metricReader = new PeriodicExportingMetricReader({
  exporter: new OTLPMetricExporter(),
  exportIntervalMillis: 1000,
})

const sdk = new NodeSDK({
  // Automatic detection is disabled, see comment below
  autoDetectResources: false,
  metricReader,
```

ludovicm67 (commented):
Some logic for the metrics should be moved to the SDK part.
This will register them globally.

ludovicm67 (approved):

PR #1379, opened 2023-02-16
fix: auto-mapping large dimensions

tpluscode:


apis/core/lib/domain/queries/dimension-mappings.ts
```diff
@@ -138,10 +154,6 @@ export async function importMappingsFromSharedDimension({ dimensionMapping, dime
...
 
       ${validThroughFilter}
-    }
-
-    FILTER (str(?value) = str(?identifier))
-  `
-
-  await insert.execute(client.query)
+    `
+    .execute(publicClient.query)
```
cristianvasquez, line 158: Oh, where did away go?

cristianvasquez (approved):
Look ok to m!

PR #1384, opened 2023-03-01
Suggestions for OTEL upgrade

ludovicm67:
Some suggestions for the `otel-upgrade` branch:
- import order in `cli/server.ts` is important
- the `finally` in `cli/server.ts` is not required, since it's a process that is always running ; and it was directly calling the shutdown, which is not something we want
- removed `cli/lib/otel/metrics.ts` and moved the metrics in the relevant files ; this was also to be sure that things gets initialized in the right order


cli/server.ts
```diff
@@ -30,7 +30,7 @@ async function main() {
 
   app.use(cors())
 
-  app.use(bodyParser.urlencoded())
+  app.use(bodyParser.urlencoded({ extended: true }))
```
tpluscode, line 33: what does this do?

ludovicm67: This was the default value, but with the current version if we don't specify it, it throws a warning message in the console.

ludovicm67: And even better:

```suggestion
  app.use(express.urlencoded({ extended: true }))
```

> This middleware is available in Express v4.16.0 onwards.

Source: https://expressjs.com/fr/api.html#express.urlencoded

tpluscode: To be honest, I would like to skip the big lockfile upgrade. See the failed image builds caused by mocha and some other changed packages. Only stick to the minimum

ludovicm67: This change will not have any impact in the app, since in `package.json` it is using `"express": "^4.17.1"`, so it's already supported ;)

body-parser is deprecated for `urlencoded` and `json` middlewares, since they are now part of `express` directly since 4.16.0

PR #1388, opened 2023-03-06
fix: relax patterns for shared dimension identifiers

tpluscode:


giacomociti:
lokks good to me, thanks

PR #1392, opened 2023-03-09
fix: only one cube table

tpluscode:


apis/core/lib/domain/queries/table.ts
```diff
@@ -88,3 +88,16 @@ export async function getTableForColumnMapping(columnMapping: NamedNode, client
...
+    .WHERE`
+      GRAPH ?table {
+        ?table a ${cc.ObservationTable} .
+        ?table ${cc.csvMapping} ${csvMapping.id} .
+      }
+    `
+    .execute(parsingClient.query)
+
+  return bindings.shift()?.table
+}
```
tpluscode, line 103: Because there is no way to access the other tables directly when create a new one, the domain code calls a SPARQL query to ensure a cube table does not already exist

apis/core/lib/domain/table/create.ts
```diff
@@ -26,13 +27,15 @@ interface CreateTableCommand {
   resource: GraphPointer
   store: ResourceStore
   dimensionMetadataQueries?: Pick<typeof DimensionMetadataQueries, 'getDimensionMetaDataCollection'>
+  tableQueries?: Pick<typeof TableQueries, 'getCubeTable'>
```
tpluscode, line 30: As in other functions in `lib/domain/`, this SPARQL query is injected as an optional function call

apis/core/lib/domain/queries/table.ts
cristianvasquez: How are cube-tables identified? Is the `csvMapping.id`  discriminative attribute?

tpluscode: I don't understand. `csvMapping.id` is that resource's URI. The data is that all tables are linked to their CSV Mapping resource. And only one of those tables can have the type `cc:ObservationTable` (notice old terminology, as now refer to them as cube tables)

cristianvasquez: yes, I'm confused 

so, one can have:
- Many tables, each pointing to a csvMapping
- One of those tables is an Observation table

Is this always?

So this method wants to check for ObservationTables?

It could be named 

`currentObservationTable``

apis/core/test/domain/table/create.test.ts
```diff
@@ -139,6 +143,27 @@ describe('domain/table/create', () => {
     })
   })
 
+  it('does not allow multiple Cube Tables', async () => {
```
cristianvasquez, line 146: Is this one for any Cube Table or the ones that have isObservationTable?

cristianvasquez (commented):
I'm confused about the model, from the code alone is difficult to tell.

I think changing the function names or adding more comments could help,

PR #1399, opened 2023-03-21
Add Node 18 support

ludovicm67:
This PR is doing the following:
- use Node 18 in the CI
- use Node 18 in the Docker images
- use Node 18 in `.lando.yml` file (needed to create a `/var/www/.npm-global/lib` directory manually so that it is able to run `npx nodemon`)
- upgrade `@vue/eslint-config-standard` from v6 to v8 in the `ui` part to allow the installation of dependencies for Node > 16
- configure a setter in `packages/core/env.ts` so that tests are not failing


.lando.yml
```diff
@@ -5,10 +5,10 @@ env_file:
 services:
   core:
     type: node:custom
-    command: npx nodemon -w packages -w apis --verbose --ext ts,ttl,js,trig --ignore **/*.test.ts --ignore **/*.spec-graphs.ts --exec node --inspect=0.0.0.0:45671 -r ts-node/register --inspect apis/core/index.ts
+    command: mkdir -p /var/www/.npm-global/lib && npx nodemon -w packages -w apis --verbose --ext ts,ttl,js,trig --ignore **/*.test.ts --ignore **/*.spec-graphs.ts --exec node --inspect=0.0.0.0:45671 -r ts-node/register --inspect apis/core/index.ts
```
tpluscode, line 8: What's in the `/var/www/.npm-global/lib` dir?

ludovicm67: Calling `npx nodemon` will install it first, and to install it, this path is needed.
It seems that in the node 18 container this directory is missing and that it is created on the fly, that's why it needed to create it manually.

tpluscode (approved):

PR #1407, opened 2023-03-27
refactor: graph store post new dimension mappings

tpluscode:
This changes how new dimension mappings are saved to the database when "auto-mapping"

The current implementation performed an `INSERT` with a complex pattern and a long list `VALUES` clause. Potentially thousands of items long.

Since the full set of these `VALUES` is loaded to memory anyway, they are actually cross-referenced in memory and new items are streamed directly as quads using the graph protocol. This appears much more database-friendly. While Stardog does cope, local fuseki actually failed to insert when mapping a large dimension of municipalities (3k terms)

### DIFF
--- a/apis/core/lib/domain/queries/dimension-mappings.ts
+++ b/apis/core/lib/domain/queries/dimension-mappings.ts
@@ -1,5 +1,6 @@
 import { Literal, NamedNode, Term } from 'rdf-js'
-import { INSERT, SELECT } from '@tpluscode/sparql-builder'
+import { Readable } from 'stream'
+import { SELECT } from '@tpluscode/sparql-builder'
 import { sparql, SparqlTemplateResult } from '@tpluscode/rdf-string'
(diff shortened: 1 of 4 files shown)


apis/core/lib/domain/queries/dimension-mappings.ts
```diff
@@ -90,40 +96,44 @@ interface ImportMappingsFromSharedDimension {
...
+      })
+
+    for (const quad of pointer.dataset) {
+      this.push(quad)
+    }
+    return next()
+  })
 
-  await insert.execute(client.query)
+  await client.store.post(Readable.from(unmappedValues).pipe(newMappings))
```
tpluscode, line 136: This is the most important change. The unmapped values are streamed and transformed in the `through2` stream to new dictionary entries.

This removes the need to serialize the result of `getSharedDimensionTerms` as a gigantic list of `VALUES`

PR #1410, opened 2023-03-28
feat: add ofcom profile

dabo505:
We need to add this new profile to Cube Creator

### DIFF
--- a/.changeset/few-shirts-double.md
+++ b/.changeset/few-shirts-double.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/core-api": patch
+---
+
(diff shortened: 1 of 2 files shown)


tpluscode (approved):
Looks good but note that the domain `communication.ld.admin.ch` does not have a valid SSL certificate.

dabo505:
> Looks good but note that the domain `communication.ld.admin.ch` does not have a valid SSL certificate.

Ah good to know, will forward that to Pierre.

tpluscode:
Would you like proceed with this profile already or add another one to this same PR? @dabo505

PR #1414, opened 2023-03-30
Improve select boxes

tpluscode:
I added a missing functionality in the form which lets us sort by `schema:name` as we have in shared dimensions and users. The default was to sort by `rdfs:label` which was hardcoded prior

That said, the unit are already sorted correctly but seeing how log the list was, I added text filter to the underlying query.


apis/core/bootstrap/shapes/cube-project.ts
```diff
@@ -210,6 +210,7 @@ ${shape('cube-project/search')} {
       ${sh.minCount} 1 ;
       ${sh.maxCount} 1 ;
       ${dash.editor} ${dash.InstancesSelectEditor} ;
+      ${sh1.orderBy} ( ${schema.name} ) ;
```
giacomociti, line 213: just asking: I guess this change relies on the UI already supporting 'orderBy'

apis/core/bootstrap/lib/query.ts
```diff
@@ -6,3 +6,12 @@ import env from '@cube-creator/core/env'
 export function lindasQuery(query: SparqlTemplateResult): NamedNode {
   return $rdf.namedNode(`${env.PUBLIC_QUERY_ENDPOINT}?query=${encodeURIComponent(query.toString())}`)
 }
+
+export function lindasQueryTemplate(query: SparqlTemplateResult, ...variables: string[]) {
+  const url = lindasQuery(query)
+  return variables.reduce(replacePlaceholders, url.value)
+}
+
+function replacePlaceholders(prev: string, variable: string) {
```
giacomociti, line 15: is there any SPARQL parser library so that we may do the replacement at the query AST level? Or the regex approach is good enough anyway?

apis/core/bootstrap/shapes/cube-project.ts
tpluscode: Yes, added to shaperone, hence it was updated in `package.json`

apis/core/bootstrap/lib/query.ts
tpluscode: Yes, there is [sparqljs](https://npm.im/sparqljs) but here the case is not exactly parsing SPARQL

The query goes to the query string, urlencoded. The replacement is to insert template placeholder like `{VAR}` into it. The braces need not be escaped

PR #1418, opened 2023-04-05
fix: mapping dimension with typed literals

tpluscode:
Typed literals would not be correctly replaced with shared dimension terms because the comparison was done strictly, according to RDF rules

This PR relaxes that and compares the values as their plain literal representations


apis/core/lib/domain/dimension-mapping/DimensionMapping.ts
```diff
@@ -11,7 +11,7 @@ interface DictionaryEx {
   about: NamedNode
   sharedDimensions: NamedNode[]
   onlyValidTerms: boolean
-  replaceEntries(entries: KeyEntityPair[]): Map<Term, Term>
+  replaceEntries(entries: KeyEntityPair[]): boolean
```
giacomociti, line 14: consider replacing `boolean`  with a more explicit type like `{ entriesChanged:  boolean }` so that client code doesn't have to guess the meaning.

giacomociti (approved):

PR #1419, opened 2023-04-06
feat: inline shacl violations

tpluscode:
Not perfect but at least some field will start showing inline errors

Problematic cases include nested forms and dynamic, conditional fields. For those it's not that easy to figure out that a given error should be rendered where

https://user-images.githubusercontent.com/588128/230398895-92dc54fc-e72b-4a2d-b2bd-a41c4ca66a61.mp4

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12050,21 +12050,13 @@ rdf-utils-fs@^2, rdf-utils-fs@^2.1.0:
     "@rdfjs/formats-common" "^2.2.0"
     readable-stream "^3.6.0"
 
-rdf-validate-datatype@^0.1.1:
+rdf-validate-datatype@^0.1.1, rdf-validate-datatype@^0.1.3, rdf-validate-datatype@^0.1.4:
   version "0.1.5"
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -1569,6 +1569,13 @@
     clownface "^1.2.0"
     clownface-shacl-path "^1.3.2"
 
+"@hydrofoil/shaperone-rdf-validate-shacl@^1.0.3":
+  version "1.0.3"
+  resolved "https://registry.yarnpkg.com/@hydrofoil/shaperone-rdf-validate-shacl/-/shaperone-rdf-validate-shacl-1.0.3.tgz#8141b0e2dec7b3e14fe798162864e983cb949537"
+  integrity sha512-hYNGdJagQMHwF98nFj7qVBK6SKrMGPm/WtuHmwayTnRZbC9n1n6FmybHDVfGC9+NE9U5vbq8eicpMCqHFtKU5w==
+  dependencies:
+    rdf-validate-shacl zazuko/rdf-validate-shacl#bnodes-clash
```
tpluscode, line 1577: omg I really published with a dependency like that 🙈

PR #1421, opened 2023-04-11
feat: searchable users collection

tpluscode:


giacomociti (approved):
it's not easy for me to spot issues or provide suggestions, but at least I had the chance to dig a bit into hydra concepts like IRI template mappings

PR #1430, opened 2023-04-24
build(lando): node:custom

tpluscode:
Using `node:custom` and specifying the image version in `overrides` allows the local set up to work on macOS Catalina which is not supported by later versions of lando (last being 3.6.5)


ludovicm67:
I will quickly check with Ventura if everything is still OK.
I will approve this PR if it's the case.

ludovicm67 (approved):

PR #1433, opened 2023-05-25
fix: performance of getting terms

tpluscode:


apis/shared-dimensions/lib/domain/shared-dimensions.ts
```diff
@@ -36,42 +43,52 @@ interface GetSharedTerms {
...
-      !bound(?validThrough) || ?validThrough >= ${toRdf(validThrough)}
-    )`
+    const placeholder = $rdf.literal('VALID-THROUGH')
+    filterShape.out(sh.expression).deleteOut(sh.deactivated)
+    filterShape.dataset.match(null, null, placeholder).forEach((quad) => {
+      filterShape.dataset.delete(quad).add($rdf.quad(quad.subject, quad.predicate, toRdf(validThrough), quad.graph))
+    })
   }
 
-  return DESCRIBE`
```
tpluscode, line 66: I changed a DESCRIBE query to a CONSTRUCT generated from SHACL shape using https://shape-to-query.hypermedia.app/docs/

apis/shared-dimensions/lib/shapeToQuery.ts
```diff
@@ -0,0 +1,93 @@
...
+const setup = onetime(async () => {
+  await defineConstraintComponents()
+})
+
+async function defineConstraintComponents() {
+  const { ConstraintComponent } = await _importDynamic('@hydrofoil/shape-to-query/model/constraint/ConstraintComponent.js')
+  const { constraintComponents } = await _importDynamic('@hydrofoil/shape-to-query/model/constraint/index.js')
+  const { PatternConstraintComponent } = await _importDynamic('@hydrofoil/shape-to-query/model/constraint/pattern.js')
+
+  constraintComponents.set(md.FreeTextSearchConstraintComponent, class TextSearch extends ConstraintComponent {
```
tpluscode, line 38: Here is an implementation of SHACL node expression which generates full-text search patterns

In Stardog/Fuseki, vendor-specific constructs are used. Otherwise it falls back to plain REGEX

apis/shared-dimensions/lib/shapes/terms-query-shape.ttl
```diff
@@ -0,0 +1,102 @@
+@prefix sparql: <http://datashapes.org/sparql#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix hydra: <http://www.w3.org/ns/hydra/core#> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix schema: <http://schema.org/> .
+@prefix meta: <https://cube.link/meta/> .
+@prefix s2q: <https://hypermedia.app/shape-to-query#> .
+
+[
```
tpluscode, line 10: This is the shape which is the source for the query. The limit, offset, sh:filterShape and sh:expression are dynamically preprocessed to adjust for API query arguments

PR #1485, opened 2024-03-05
Support for hidden cubes

mchlrch:
* Adds flag to the cubes metadata
* Sets the publishGraph in the PublishJob to a `.../hidden graph

### DIFF
--- a/apis/core/bootstrap/shapes/dataset.ts
+++ b/apis/core/bootstrap/shapes/dataset.ts
@@ -1,4 +1,4 @@
-import { shape, editor, lindasSchema } from '@cube-creator/core/namespace'
+import { shape, editor, lindasSchema, cc } from '@cube-creator/core/namespace'
 import { supportedLanguages } from '@cube-creator/core/languages'
 import { hydra, rdfs, sh, dcat, dcterms, xsd, rdf, vcard, schema, _void, dash, skos } from '@tpluscode/rdf-ns-builders'
 import { sparql, turtle } from '@tpluscode/rdf-string'
@@ -209,6 +209,19 @@ ${shapeId} {
       ${sh.description} "Only published datasets will be listed in the external tools. A draft will be nevertheless be public." ;
       ${sh.group} ${mainGroup} ;
     ] ;
+
+    ${sh.property} [
+      ${sh.name} "Hide this cube?" ;
+      ${sh.path} ${cc.isHiddenCube} ;
+      ${sh.datatype} ${xsd.boolean} ;
+      ${sh.defaultValue} ${false} ;
+      ${sh.minCount} 1 ;
+      ${sh.maxCount} 1 ;      
+      ${sh.order} 31 ;
+      ${sh.description} "Hidden cubes are only visible in *visualize* for logged-in users" ;
+      ${sh.group} ${mainGroup} ;
+    ] ;
+
     ${sh.property} [
         ${sh.name} "Publish to" ;
         ${sh.path} ${schema.workExample} ;
--- a/apis/core/lib/domain/job/create.ts
+++ b/apis/core/lib/domain/job/create.ts
@@ -76,12 +76,14 @@ export async function createPublishJob({
 
   const metadata = await store.getResource(project.dataset)
 
+  const targetGraph = metadata.isHiddenCube ? $rdf.namedNode(organization.publishGraph.value + '/hidden') : organization.publishGraph
+
   const jobPointer = await store.createMember(jobCollection.term, id.job(jobCollection))
   const job = Job.createPublish(jobPointer, {
     project: projectPointer,
     name: 'Publish job',
     revision: project.nextRevision,
-    publishGraph: organization.publishGraph,
+    publishGraph: targetGraph,
     status: metadata?.pointer.out(schema.creativeWorkStatus).term,
     publishedTo: metadata?.pointer.out(schema.workExample).term,
   })
--- a/packages/model/Dataset.ts
+++ b/packages/model/Dataset.ts
@@ -16,6 +16,7 @@ export interface Dataset extends RdfResource {
   title: Literal[]
   created: Date
   published: Date
+  isHiddenCube: boolean
 }
 
 export const Error = {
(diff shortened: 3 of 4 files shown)


packages/model/Dataset.ts
```diff
@@ -43,6 +44,9 @@ export function DatasetMixin<Base extends Constructor>(Resource: Base): Mixin {
 
     @property.literal({ path: schema.datePublished, type: Date, datatype: xsd.date })
     published!: Date
+
+    @property.literal({ path: cc.isHiddenCube, datatype: xsd.boolean })
+    isHiddenCube!: boolean
```
tpluscode, line 49: As discussed, I would move and duplicate this field in order to memoize the current cube setting for the event of rerunning a job

1. Add to [`Project`](https://github.com/zazuko/cube-creator/blob/dd4cba3934efbb20e2d4e91a909b9ab3845dfd21/packages/model/Project.ts#L18)
2. Add to [`PublishJob`](https://github.com/zazuko/cube-creator/blob/dd4cba3934efbb20e2d4e91a909b9ab3845dfd21/packages/model/Job.ts#L35)

In order to connect those pieces you'd need to modify the project shape (create and update) and then when then copy to the job when it is created. Look for `createPublishJob` which the domain logic entrypoint.

Finally, you'd need to add or modify some step in `publish.ttl`. There ar already some steps which generate additional metadata so it should not be a big deal.

apis/core/lib/domain/job/create.ts
```diff
@@ -76,12 +76,14 @@ export async function createPublishJob({
 
   const metadata = await store.getResource(project.dataset)
 
+  const targetGraph = metadata.isHiddenCube ? $rdf.namedNode(organization.publishGraph.value + '/hidden') : organization.publishGraph
```
tpluscode, line 79: Just thinking out loud, whether we want this implicit or maybe add to the organisation profile. Final effect may be the same but maybe it's going to be a little more visible for VSHN et.al. which graphs are intended to be hidden.

(I realised while typing that the current ACL idea does not really need that information but who knows...)

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -209,6 +209,19 @@ ${shapeId} {
       ${sh.description} "Only published datasets will be listed in the external tools. A draft will be nevertheless be public." ;
       ${sh.group} ${mainGroup} ;
     ] ;
+
+    ${sh.property} [
+      ${sh.name} "Hide this cube?" ;
+      ${sh.path} ${cc.isHiddenCube} ;
+      ${sh.datatype} ${xsd.boolean} ;
+      ${sh.defaultValue} ${false} ;
```
tpluscode, line 217: I think we discussed with @ktk that a new project should be hidden by default to prevent accidents where someone publishes data publicly.

That would apply to the shape to create projects

tpluscode (changes requested):

mchlrch:
Where should the flag show up in UI?

image

image

image

image

image

apis/core/lib/domain/job/create.ts
mchlrch: I moved the flag from Cube Metadata to Project.

Now I have the issue, that it fails on `<Create Project>`. The validation reports seems to be empty - what does that mean?

image

packages/model/Dataset.ts
mchlrch: I moved the flag from Cube Metadata to Project.

Now I have the issue, that it fails on `<Create Project>`. The validation reports seems to be empty - what does that mean?

image

cli/lib/publish/hiddenCube.ts
```diff
@@ -0,0 +1,25 @@
...
+  const dataset = $rdf.dataset()
+  clownface({ dataset })
+    .node(cubeId)
+    .addOut(cc.isHiddenCube, !!isHiddenCube)
+
+  return through2.obj(
+    (chunk, enc, cb) => cb(null, chunk),
+    function (done) {
+      dataset.forEach(this.push.bind(this))
+      done()
```
tpluscode, line 23: `throught2` works of course but for readability, I think [generators](https://data-centric.zazuko.com/docs/workflows/how-to/implement-steps#async-generators) are the preferred option.

tpluscode: Also interesting that you used clownface for a single triple :)

apis/core/lib/domain/job/create.ts
```diff
@@ -76,12 +76,14 @@ export async function createPublishJob({
 
   const metadata = await store.getResource(project.dataset)
 
+  const targetGraph = project.isHiddenCube ? $rdf.namedNode(organization.publishGraph.value + '/hidden') : organization.publishGraph
```
tpluscode, line 79: Still thinking this could be made explicitly part of the organisation resources in DB. Could make communication with VSHN easier.

If not, might at least wrap as a getter in the [`OrganizationMixin` class](https://github.com/zazuko/cube-creator/blob/master/packages/model/Organization.ts#L28-L74).

apis/core/lib/domain/cube-projects/Project.ts
```diff
@@ -16,6 +16,7 @@ interface ApiProject {
   initializeJobCollection(store: ResourceStore): void
   incrementPublishedRevision(): void
   updateMaintainer(organization: Term | undefined): { before: Term; after: Term }
+  updateIsHiddenCube(hiddenCube: Term | undefined): void
```
tpluscode, line 19: Would rather see this simplified. 

```suggestion
  updateIsHiddenCube(hiddenCube: boolean): void
```

But why a setter in the first place. You can already just set `isHiddenCube` directly

apis/core/lib/domain/cube-projects/update.ts
```diff
@@ -23,6 +23,7 @@ export async function updateProject({
   const project = await store.getResource<ImportProject | CsvProject>(resource.term)
 
   project.rename(resource.out(rdfs.label).value)
+  project.updateIsHiddenCube(resource.out(cc.isHiddenCube).term)
```
tpluscode, line 26: And then set the flag

```suggestion
  project.isHiddenCube = $rdf.literal('true', xsd.boolean).equals(resource.out(cc.isHiddenCube).term))
```

tpluscode (changes requested):
Please update the requests in `e2e-tests`
Please add a test scenario in `cli/test/lib/commands/publish.test.ts` which checks that publishing a hidden cube writes to the expected graph

giacomociti (approved):
I'm not able to provide nuanced advice, but overall the changes make sense to me

PR #1508, opened 2024-04-26
improve validation: ensure consistent observedBy in ObservationConstraint

Rdataflow:
- pass contributor as argument
- make `cube:observedBy` to be consistent in the ObservationConstraint
- add test
- drop injection

### DIFF
--- a/cli/lib/metadata.ts
+++ b/cli/lib/metadata.ts
@@ -296,3 +296,17 @@ export async function injectObservedBy(this: Context, jobUri: string) {
     callback()
   })
 }
+
+export async function getObserver(this: Context, jobUri: string) {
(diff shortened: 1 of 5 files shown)


Rdataflow:
@tpluscode would you mind starting the CI test ?

Rdataflow:
https://github.com/zazuko/barnard59/pull/324

Rdataflow:
@tpluscode it seems the current workflow injects/overrides the observer but only in the observations...

How would it be possible to set the observer globally (observations, shape)? 
would it i.e. make more sense to pass the `observer` argument in the transform step already (instead of injection) WDYT?

feel free to build upon theese crafted commits... it's at your disposition 👍

PR #1514, opened 2024-05-14
build(deps): update barnard59 packages

tpluscode:
First step getting the new cube shape steps from `barnard59-cube`

I kept `barnard59-graph-store` at v5 because v6 now only sends the entire stream to a single graph


cli/pipelines/publish.ttl
```diff
@@ -129,24 +142,24 @@
...
       a code:EcmaScriptModule  ;
-      code:link <node:stream#Readable.from> ;     
+      code:link <node:stream#Readable.from> ;
     ] ;
      code:arguments (
       """[
-        'https://cube.link/v0.1.0/shape/standalone-constraint-constraint',
-        // 'https://cube.link/v0.1.0/shape/datacatalog-constraint',        
-      ]"""^^code:EcmaScript  
+        'https://cube.link/latest/shape/standalone-constraint-constraint',
```
giacomociti, line 151: v0.1.0 is obsolete, but we really want the latest instead of a fixed, recent version?

tpluscode: Good point. This should be configurable too, at some point

PR #1515, opened 2024-05-16
build(deps): replace local `toCubeShape` with `barnard59-cube`

tpluscode:
* [x] removed the `toCubeShape` copy
* [x] patched `rdf-parser-csvw`
* [ ] adjust tests


tpluscode:
@giacomociti I removed one last failing test for the case of min/max + IRI dimension. Thought it did not make sense to replicate that test from b59 as it ws getting a little awkward

giacomociti (approved):

Issue #1373, opened 2023-02-09
empty identifier template


giacomociti:
When creating a new table from CSV, leaving the identifier template empty should get an auto-generated identifier.

But if I leave it empty and press 'Create Table', a template is shown anyway (it lacks the table name and includes all the columns' names). Then, if I edit the table, in the edit panel the identifier is not there and the page becomes irresponsive.

PR #1516, opened 2024-05-20
fix: editing table identifier template

tpluscode:


.changeset/dull-coins-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cube-creator/ui": patch
+---
+
+CSV mapping: it was impossible to save an existing table with auto-generated identifier. Fixes #1737
```
giacomociti, line 5: I don't see any issue #1737

giacomociti (approved):

.changeset/dull-coins-thank.md
tpluscode: Of course, I meant #1373

PR #1519, opened 2024-05-27
Error messages

giacomociti:
Use a SHACL shape to prevent mixing language and datatype (`xsd:string` is treated as an exception and silently replaced with `rdf:langString` to avoid bothering the user with this subtle distinction).

Provide detailed but simple error messages for invalid values of the supported data types (with links to formal specs).

### DIFF
--- a/apis/core/test/domain/column-mapping/create.test.ts
+++ b/apis/core/test/domain/column-mapping/create.test.ts
@@ -98,7 +98,7 @@ describe('domain/column-mapping/create', () => {
       .node($rdf.namedNode(''))
       .addOut(cc.sourceColumn, $rdf.namedNode('my-column'))
       .addOut(cc.targetProperty, $rdf.namedNode('test'))
-      .addOut(cc.datatype, xsd.integer)
+      .addOut(cc.datatype, xsd.integer) // becomes rdf.langString because of language
       .addOut(cc.language, $rdf.literal('fr'))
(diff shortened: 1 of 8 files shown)


cli/lib/validation.ts
```diff
@@ -17,3 +19,45 @@ function quadToString(quad: Quad): string {
     graph: quad.graph as NamedNode,
   })
 }
+
+// consider replacing should with may
```
giacomociti, line 23: for example, the message for `xsd:dateTime` is "**should** be formatted as YYYY-MM-DDThh:mm:ss", but the spec allows for many more variations; describing all of them in a simple yet precise way is challenging, that's why **may** instead of **should** may be more appropriate

apis/core/test/domain/column-mapping/create.test.ts
```diff
@@ -121,7 +121,7 @@ describe('domain/column-mapping/create', () => {
         minCount: 1,
       }, {
         path: cc.datatype,
-        hasValue: xsd.integer,
```
tpluscode, line 124: That I actually made such a nonsensical unit test 🙈

cli/lib/validation.ts
tpluscode: I think **should** is appropriate :)

tpluscode (changes requested):
Please add changesets to API and CLI

tpluscode (approved):

PR #1523, opened 2024-06-07
fix: transform invalid dates

tpluscode:


giacomociti (approved):

giacomociti:
with the latest version of the csvw parser, invalid date values (like `'foo' `or `'500000'`) are correctly reported. Still, a value like `'2000'` is silently parsed successfully and transformed to `'2000-01-01'^^xsd:date`

tpluscode:
> I guess this is the desired behavior

Have you compared with the previous version?

PR #1533, opened 2024-09-23
opendata.swiss resources

tpluscode:


apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -46,6 +51,40 @@ const euThemesQuery = sparql`construct {
...
+  ?term ${rdfs.label} ?name .
+} WHERE {
+  BIND ( bnode('collection') as ?c )
+
+  graph <https://lindas.admin.ch/ontologies> {
+    ?term a ${euvoc.FileType} ;
+     ${skos.prefLabel} ?name ;
+      ${skos.inScheme} <http://publications.europa.eu/resource/authority/file-type> .
+
+    FILTER(regex(?name, "_q_", "i"))
```
giacomociti, line 66: why the filter? in TEST,it filters out all values

tpluscode: That's a template. Does `_q_` instead of `{q}` to prevent URL encoding an gets replaced later.

From the UI, the value of `q` is provided from the auto-complete editor

tpluscode: Yes. The queries go against `PUBLIC_QUERY_ENDPOINT`.

You can add `PUBLIC_QUERY_ENDPOINT=https://test.lindas.admin.ch/query` to `.env` in project root and then `lando rebuild`

apis/core/bootstrap/shapes/dataset.ts
```diff
@@ -384,6 +424,82 @@ ${shapeId} {
...
+    ] ;
+  .
+
+  ${opendataSwissResources} a ${sh.NodeShape} ;
+    ${sh.property} [
+      ${sh.name} "Access URL" ;
+      ${sh.path} ${dcat.accessURL} ;
+      ${sh.minCount} 1 ;
+      ${sh.nodeKind} ${sh.IRI} ;
+      ${sh.order} 10 ;
```
giacomociti, line 445: consider adding a specific validation message like: `${sh.message} "Need some URL to access the data." ;`

The current generic message: 
> Access URL:  Less than 1 values

may not be clear enough when a value is specified but it's not recognized as URL

giacomociti (approved):

PR #1534, opened 2024-09-24
Fix observer

giacomociti:
builds upon #1508 . The `cube:observedBy` property is set with the correct value during the _transform_  job, avoiding the need to replace the values during the _publish_ job. We gain in simplicity: the correct value, derived from the publishing profile in the project settings, is passed as ENV variable, it's used for all observations and the corresponding cube constraint is automatically derived. The only downsize I see is the need to rerun the _transform_  job if one changes the publishing profile.

### DIFF
--- a/.changeset/loud-news-drop.md
+++ b/.changeset/loud-news-drop.md
@@ -1,0 +1,7 @@
+---
+"@cube-creator/model": patch
+"@cube-creator/core-api": patch
+"@cube-creator/cli": patch
+---
+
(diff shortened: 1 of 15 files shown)


tpluscode:
> The only downsize I see is the need to rerun the transform job if one changes the publishing profile.

That never really happened, so :+1:

tpluscode (approved):

Rdataflow:
@giacomociti well done 🎉

PR #1535, opened 2024-09-24
Improve cube validation

giacomociti:
on _publish_ job, choose appropriate SHACL shapes if need to publish on ODS and/or visualize


cli/shapes-all.ttl
```diff
@@ -0,0 +1,16 @@
+prefix code: <https://code.described.at/>
+
+[
+    code:imports <./shapes> ;
+    code:extension "ttl" ;
+] .
+
+[
+    code:imports <https://cube.link/v0.2.2/shape/profile-opendataswiss-lindas> ;
```
tpluscode, line 9: With the version number hidden in multiple turtle sources, it may be cumbersome to update, much less override per-project or at runtime.

Can't really think of a good solution off the top of my head but I think we need something here

tpluscode (changes requested):

cli/shapes-all.ttl
giacomociti: one step forward could be using a version placeholder in the files, to be replaced on the fly with a value specified once (with an ENV variable?). Is it feasible to add a step in [<#loadCubeShapes>](https://github.com/zazuko/cube-creator/blob/e63ee83590facac142b91d7c20d8d7a5caded997/cli/pipelines/publish.ttl#L90) between <#open> and <#imports> in charge of replacing the placeholder?

tpluscode (approved):

PR #1537, opened 2024-09-26
shape version specified only once

giacomociti:
is it a step in the right [direction](https://github.com/zazuko/cube-creator/pull/1535#discussion_r1773114171)? I'm not sure the added complexity is worth it


cli/lib/commands/publish.ts
```diff
@@ -42,6 +42,7 @@ export default runner.create<PublishRunOptions>({
     variable.set('publish-graph-store-password', publishStore?.password || process.env.PUBLISH_GRAPH_STORE_PASSWORD)
     variable.set('metadata', $rdf.dataset())
     variable.set('shapesPath', path.resolve(__dirname, `../../${profile}.ttl`))
+    variable.set('shapesVersion', 'v0.2.2')
```
tpluscode, line 45: I think I'd move the default directly to the replace step module module and keep the variable only for a potential override from project

cli/lib/replace.ts
```diff
@@ -0,0 +1,19 @@
+import through2 from 'through2'
+import type { Quad } from '@rdfjs/types'
+import $rdf from 'rdf-ext'
+
+function replace(version: string) {
```
tpluscode, line 5:
```suggestion
function replace(version: string = 'v2.2') {
```

cli/lib/replace.ts
```diff
@@ -0,0 +1,19 @@
+import through2 from 'through2'
+import type { Quad } from '@rdfjs/types'
+import $rdf from 'rdf-ext'
```
tpluscode, line 3: You may want to prefer `this.env` ;)

cli/lib/replace.ts
```diff
@@ -0,0 +1,19 @@
...
+  const replacement = `https://cube.link/${version}/`
+  return through2.obj(function (quad: Quad, _, done) {
+    if (quad.predicate.value === 'https://code.described.at/imports' && quad.object.value.startsWith(prefix)) {
+      const replaced = quad.object.value.replace(prefix, replacement)
+      const triple = $rdf.triple(quad.subject, quad.predicate, $rdf.namedNode(replaced))
+      done(null, triple)
+    } else {
+      done(null, quad)
+    }
+  })
```
tpluscode, line 16: I think this perfect candidate for a generator-style step

cli/lib/replace.ts
```diff
@@ -0,0 +1,19 @@
+import through2 from 'through2'
+import type { Quad } from '@rdfjs/types'
+import $rdf from 'rdf-ext'
+
+function replace(version: string) {
+  const prefix = 'https://cube.link/latest/'
+  const replacement = `https://cube.link/${version}/`
+  return through2.obj(function (quad: Quad, _, done) {
+    if (quad.predicate.value === 'https://code.described.at/imports' && quad.object.value.startsWith(prefix)) {
```
tpluscode, line 9:
```suggestion
    if (quad.predicate.equals(env.ns.code.imports && quad.object.value.startsWith(prefix)) {
```

Of course you'd need to get `env` from step context

cli/pipelines/publish.ttl
```diff
@@ -107,6 +108,15 @@
   code:arguments ( 'shapesPath'^^:VariableName )
 .
 
+<#replace> a :Step ;
+  code:implementedBy
+    [
+      a code:EcmaScript ;
+      code:link <file:../lib/replace#default>
```
tpluscode, line 115: I think the step deserves a more descriptive name

tpluscode (changes requested):
Yea, I think it's a good direction

tpluscode (approved):

PR #1543, opened 2024-10-01
copy all turtle files in the docker image

giacomociti:
let's discuss here also with @ludovicm67  if we want a more pervasive replacement of ADD with COPY

### DIFF
--- a/.changeset/selfish-hotels-exist.md
+++ b/.changeset/selfish-hotels-exist.md
@@ -1,0 +1,5 @@
+---
+"@cube-creator/cli": patch
+---
+
+fix: include missing shape files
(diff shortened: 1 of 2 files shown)


ludovicm67 (commented):
@giacomociti I think that all the `ADD` could be replaced with `COPY` without any issue :)

.changeset/selfish-hotels-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cube-creator/cli": patch
+---
+
+fix: include missing shape files
```
tpluscode, line 5:
```suggestion
fix: missing shape files in docker image caused Publish pipeline to fail
```

PR #1546, opened 2024-10-17
reactivate list shapes

giacomociti:
list shapes were disabled to avoid validation errors with some weird lists having additional properties.
I could not reproduce the creation of such invalid lists anymore, so I suspect the problem is either fixed by a concurring change (the `observedBy` value) or due to some data inconsistency (blank node collision) during development.

Hopefully, we can restore the list shapes


Rdataflow:
@tpluscode it would be good to have this in the upcoming release...

tpluscode (approved):

PR #1551, opened 2024-10-31
Csv datatypes

giacomociti:
Guess default column datatype from CSV sample values


packages/model/CsvColumn.ts
```diff
@@ -4,12 +4,15 @@ import { Constructor, property, ResourceIdentifier } from '@tpluscode/rdfine'
...
 import { cc } from '@cube-creator/core/namespace'
+import { NamedNode } from '@rdfjs/types'
 import { initializer } from './lib/initializer'
+import { inferDatatype } from './lib/datatypeInference'
 
 export interface CsvColumn extends RdfResource {
   name: string
   order: number
   samples: string[]
+  defaultDatatype: NamedNode
```
tpluscode, line 15: maybe a better name would be `guessedDatatype`, or `inferredDatatype` to align with the module name?

apis/core/lib/domain/table/Table.ts
```diff
@@ -92,6 +92,7 @@ export default function mixin<Base extends Constructor<Table>>(Resource: Base):
       return this.addLiteralColumnMapping({
         store,
         sourceColumn: column,
+        datatype: column.defaultDatatype,
```
tpluscode, line 95: More importantly though, I wonder if we need this as a field of `CsvColumn`. Maybe just call here?

```suggestion
        datatype: inferDatatype(column.samples),
```

packages/model/CsvColumn.ts
giacomociti: the subtle rationale for choosing the name would be that the inferred datatype is used as a default value. But this doesn't matter anymore since, as you pointed out in the next comment, there's no need to carry this value around as a property of `CsVColumn`

apis/core/lib/domain/table/Table.ts
giacomociti: good observation, done

tpluscode (approved):

PR #1555, opened 2024-11-04
style: add recommended mocha rules

tpluscode:


giacomociti (approved):
just curious why mocha recommends using `function` and not arrow functions. I guess it's to make available some test utilities (like `timeout`)?

tpluscode:
> just curious why mocha recommends using `function` and not arrow functions. I guess it's to make available some test utilities (like `timeout`)?

Correct. I was surprised too. [They say](https://mochajs.org/#arrow-functions)

> Passing [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) (aka “lambdas”) to Mocha is discouraged. Lambdas lexically bind `this` and cannot access the Mocha context. For example, the following code will fail:

PR #1556, opened 2024-11-05
feat(shared-dimension): contributors

tpluscode:


apis/shared-dimensions/hydra/index.ttl
```diff
@@ -263,7 +263,7 @@ md:SharedDimension
       code:implementedBy
         [
           a code:EcmaScript ;
-          code:link <file:handlers/resource#put> ;
+          code:link <file:handlers/shared-dimension#put> ;
```
tpluscode, line 266: I was surprised to discover that the shared dimensions and hierarchies were both handled by the "resource" handler which was in fact specific to shared dimensions. I moved that to its own module and simplified the generic one

PR #1565, opened 2024-12-30
Shared Dimensions: searching and paging

tpluscode:


ui/src/views/SharedDimensions.vue
```diff
@@ -22,13 +22,18 @@
     </div>
     <div v-if="collection">
       <div class="panel-block is-flex is-justify-content-center">
-        <o-checkbox :model-value="showAll" @update:modelValue="toggleDisplay">
-          show deprecated
-        </o-checkbox>
+        <cc-hydra-operation-form
```
giacomociti, line 25: the submit button label is 'Save' (I think it's the default from `FormSubjectCancel.vue`). Something like 'Search' would be more appropriate

